PR #40, opened 2020-11-02
refactor(all): rename e2e test files to _plus_e2e to be compatible with melos

miquelbeltran:
## Description

When running `melos run test:e2e` it runs the tests based on the package name, like `/test_driver/MELOS_PARENT_PACKAGE_NAME_e2e.dart` so for example `share_plus` expects to run `test_driver/share_plus_e2e.dart`.

In this PR I rename the files so the we can run melos to run all e2e tests.

Note that I noticed that some tests won't pass, that's another issue :)


mhadaily (approved):
LGTM

PR #43, opened 2020-11-05
Add Linux support to Package Info Plus

jpnurmi:
Requires https://github.com/flutter/flutter/pull/69920

### DIFF
--- a/packages/package_info_plus_linux/lib/src/package_info.dart
+++ b/packages/package_info_plus_linux/lib/src/package_info.dart
@@ -1,0 +1,29 @@
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:package_info_plus_platform_interface/package_info_data.dart';
+import 'package:package_info_plus_platform_interface/package_info_platform_interface.dart';
+import 'package:path/path.dart' as path;
+
+/// The Linux implementation of [PackageInfoPlatform].
+class PackageInfoLinux extends PackageInfoPlatform {
+  /// Returns a map with the following keys:
+  /// appName, packageName, version, buildNumber
+  @override
+  Future<PackageInfoData> getAll() async {
+    try {
+      final exePath = File('/proc/self/exe').resolveSymbolicLinksSync();
+      final appPath = path.dirname(exePath);
+      final assetPath = path.join(appPath, 'data', 'flutter_assets');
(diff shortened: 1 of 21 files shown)


jpnurmi:
Requires https://github.com/flutter/flutter/pull/69920

mhadaily:
Seems like the build is failing on this @jpnurmi

jpnurmi:
Yeah it's still work in progress so I didn't care about the analysis failure. Let's see if they accept the patch to Flutter, first. The proposal is using the same version.json approach as on web.

P.S. We should publish all the unpublish packages because there are a few dependencies marked as ^0.1.0 here and there, but melos cannot handle it if the package doesn't exist at all. Pub get fails in the first place so melos has no chance to fix up pubspec.lock.

jpnurmi:
Or perhaps the problem is here since the CI build passes... 😁

packages/package_info_plus_linux/lib/src/package_info.dart
```diff
@@ -0,0 +1,25 @@
...
+/// The Linux implementation of [PackageInfoPlatform].
+class PackageInfoLinux extends PackageInfoPlatform {
+  /// Returns a map with the following keys:
+  /// appName, packageName, version, buildNumber
+  @override
+  Future<PackageInfoData> getAll() async {
+    final exePath = File('/proc/self/exe').resolveSymbolicLinksSync();
+    final appPath = path.dirname(exePath);
+    final assetPath = path.join(appPath, 'data', 'flutter_assets');
+    final versionPath = path.join(assetPath, 'version.json');
```
jpnurmi, line 17: handle non-existent version.json

packages/package_info_plus_linux/pubspec.yaml
```diff
@@ -0,0 +1,19 @@
+name: package_info_plus_linux
+description: Linux implementation of the package_info_plus plugin
+version: 0.1.0
+homepage: https://plus.fluttercommunity.dev/
+repository: https://github.com/fluttercommunity/plus_plugins/tree/main/packages/
+
+environment:
+  sdk: ">=2.7.0 <3.0.0"
+  flutter: ">=1.17.0 <2.0.0"
```
jpnurmi, line 9: update both

Issue #56, opened 2020-12-08
[all] Rename Method Channel names and move Java and Obj-C files


miquelbeltran:
As reported in #49, having the non-plus version of a package together with the plus version is going to cause issues.

To fix that, the following two actions need to me made on ALL PACKAGES:

1. Change the Method Channel name: Follow this format: `dev.fluttercommunity.plus.package_name` like `dev.fluttercommunity.plus.package_info`.
2. Move the Java files in the Android part of the package, the new folder structure should be `dev/fluttercommunity/plus/packagename` like `dev/fluttercommunity/plus/packageinfo`
3. Ensure package name is updated in `AndroidManifest.xml` and `build.gradle` files.
4. Rename the ObjC files in the iOS part, for example from `FLTPackageInfoPlugin.m` to `FLTPAckageInfoPlusPlugin.m`. Internal code needs to be changed as well.
5. Run the example with the non-plus plugin in pubspec.yaml, to test any collisions. Run at least the example on both Android and iOS.
6. Remove the non-plus plugin!
7. Run one last time, also run the e2e test at least on Android and iOS.
8. Release package and also interface packages and any package that depends on it, bump versions.

## TODO

- [x] battery_plus @jpnurmi  
- [x] connectivity_plus @miquelbeltran 
- [ ] ~~network_info_plus~~
- [x] device_info_plus @jpnurmi
- [x] package_info_plus @miquelbeltran 
- [x] sensors_plus @miquelbeltran 
- [x] share_plus @jpnurmi
- [x] android_alarm_manager_plus @miquelbeltran 
- [x] android_intent_plus @miquelbeltran

miquelbeltran:
I think this is now fully done! @jpnurmi @mhadaily 

Why is netform_info_plus stricked?

jpnurmi:
> I think this is now fully done! @jpnurmi @mhadaily
> 
> Why is netform_info_plus striked?

Yep, should be all done! It was me who struck out network_info_plus because we don't have such a plugin yet.

miquelbeltran:
Ah! true, I did not realise that when I looked at the list of packages in the README. Thanks!

PR #64, opened 2020-12-11
[battery_plus] rename method channel and plugins (#56)

jpnurmi:
Tested:
- [x] Android
- [x] iOS
- [x] Linux
- [x] macOS
- [x] Web
- [x] Windows


miquelbeltran (approved):
Thank you!

Like with share_plus, all the packages needs to be released again with versions and changelogs updated.

Issue #65, opened 2020-12-11
[connectivity_plus] Web is broken because of an indirect FFI dependency


jpnurmi:
Dependency chain: connectivity_plus -> connectivity_plus_linux -> dbus -> ffi

```
Launching lib/main.dart on Chrome in debug mode...
../../../../../../Flutter/.pub-cache/hosted/pub.dartlang.org/dbus-0.1.0/lib/src/getuid.dart:1:8: Error: Not found: 'dart:ffi'
import 'dart:ffi';                                                      
       ^                                                                
../../../../../../Flutter/.pub-cache/hosted/pub.dartlang.org/dbus-0.1.0/lib/src/getuid.dart:3:20: Error: Type 'Int32' not found.
typedef _getuidC = Int32 Function();                                    
                   ^^^^^                                                
../../../../../../Flutter/.pub-cache/hosted/pub.dartlang.org/dbus-0.1.0/lib/src/getuid.dart:8:17: Error: Getter not found: 'DynamicLibrary'.
  final dylib = DynamicLibrary.open('libc.so.6');                       
                ^^^^^^^^^^^^^^                                          
Waiting for connection from debug service on Chrome...
```

Unfortunately we have to apply the same workaround as in:
- package_info_plus_windows: https://github.com/fluttercommunity/plus_plugins/commit/cf13721d4daf7b838fff6b7442f41a8a67b9fd34
- path_provider_windows: https://github.com/flutter/plugins/commit/25957033132ac17094e82bec58e429fac073a483

PR #66, opened 2020-12-12
[connectivity_plus] eliminate indirect FFI dependency (#65)

jpnurmi:
Same as
- package_info_plus_windows: https://github.com/fluttercommunity/plus_plugins/commit/cf13721d4daf7b838fff6b7442f41a8a67b9fd34
- path_provider_windows: https://github.com/flutter/plugins/commit/25957033132ac17094e82bec58e429fac073a483
- ...

Moves the real implementation of connectivity_plus_linux behind a conditional export, instead of exporting a stub on platforms that don't support `dart:ffi`. This avoids build breakage in web projects that have transitive dependencies on connectivity_plus (and thus connectivity_plus_linux due to manual endorsement).

This will no longer be necessary once https://github.com/flutter/flutter/issues/52267 is fixed since only Linux builds will ever need to have a code-level dependency on connectivity_plus_linux.


miquelbeltran (approved):
Thank you! Is it necessary to release a new version for connectivity_plus and connectivity_plus_linux?

jpnurmi:
> Thank you! Is it necessary to release a new version for connectivity_plus and connectivity_plus_linux?

I think both. connectivity_plus_linux is going to be released for the first time.

PR #67, opened 2020-12-12
[share_plus] rename method channel and plugins (#56)

jpnurmi:
Tested:
- [x] Android
- [x] iOS
- [x] Linux
- [ ] ~~macOS~~ (N/A)
- [x] Web
- [ ] Windows

### DIFF
--- a/packages/share_plus/android/src/main/java/dev/fluttercommunity/plus/share/SharePlusPlugin.java
+++ b/packages/share_plus/android/src/main/java/dev/fluttercommunity/plus/share/SharePlusPlugin.java
@@ -2,7 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-package io.flutter.plugins.share;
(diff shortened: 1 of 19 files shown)


packages/share_plus/android/src/main/java/dev/fluttercommunity/plus/share/SharePlusPlugin.java
```diff
@@ -14,15 +14,15 @@
 import io.flutter.plugin.common.PluginRegistry.Registrar;
 
 /** Plugin method host for presenting a share sheet via Intent */
-public class SharePlugin implements FlutterPlugin, ActivityAware {
+public class SharePlusPlugin implements FlutterPlugin, ActivityAware {
```
miquelbeltran, line 17: I think renaming the Java class is not necessary but thanks for doing it aways!

miquelbeltran (approved):
thanks a lot!

-> This still needs a new package release (for share_plus and all dependencies). <-

Issue #74, opened 2020-12-14
[all] unify analysis_options.yaml


miquelbeltran:
Replace each analysis_options with just:

```
include: ../analysis_options.yaml
```
and have a single analysis_options.yaml in the root.

miquelbeltran:
done

PR #84, opened 2020-12-17
Unify Analysis Options YAML into a single reference file (see #74)

miquelbeltran:
This PR unifies all projects `analysis_options.yaml` into a single one and upgrades pedantic to 1.9.2, fixing all new analysis issues raised.


miquelbeltran:
It's a massive PR, sorry, not a lot of lines changed but a lot of files.

jpnurmi:
> It's a massive PR, sorry, not a lot of lines changed but a lot of files.

Could be done in steps one package at a time to make it easier to review? :)

miquelbeltran:
Maybe yes, in this case is because updating the version of `pedantic` on melos so it affected all packages. Some packages were at `1.8.0` while some others at `1.9.0`.

jpnurmi:
Is the relative include from the repo root necessary? Would that break stand-alone packages downloaded from pub.dev?

P.S. Do you know how `flutter/plugins` does it? They seem to have one `analysis_options.yaml` in the repo root without any additional includes in the packages:
```
flutter/plugins (master)$ git grep analysis_options
analysis_options.yaml:include: package:pedantic/analysis_options.1.8.0.yaml
packages/video_player/video_player_web/CHANGELOG.md:* Add `analysis_options.yaml` to the package, so we can ignore `undefined_prefixed_name` errors. Works around https://github.com/flutter/flutter/issues/41563.
script/incremental_build.sh:# Plugins that deliberately use their own analysis_options.yaml.
```

miquelbeltran:
you are right, it is not necessary to have the file with the include except if you are specifying different rules, so I remove them all and just leave the root one

packages/battery_plus/lib/battery_plus.dart
```diff
@@ -23,9 +23,7 @@ class Battery {
   // EventChannel because it is overridden. Forcing the class to be a singleton class can prevent
   // misuse of creating a second instance from a programmer.
   factory Battery() {
-    if (_singleton == null) {
-      _singleton = Battery._();
-    }
+    _singleton ??= Battery._();
```
jpnurmi, line 26: :+1: Nice, I've been tempted to fix these

packages/package_info_plus/example/analysis_options.yaml
```diff
@@ -0,0 +1 @@
+include: ../../../analysis_options.yaml
```
jpnurmi, line 1: can be removed?

packages/package_info_plus_macos/pubspec.yaml
```diff
@@ -18,6 +18,3 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
-
-dev_dependencies:
-  pedantic: ^1.8.0
```
jpnurmi, line 23: is this correct or should it be bumped to 1.9.2?

packages/package_info_plus_web/lib/package_info_plus_web.dart
```diff
@@ -17,8 +17,8 @@ class PackageInfoPlugin extends PackageInfoPlatform {
 
   @override
   Future<PackageInfoData> getAll() async {
-    String url =
-        "${Uri.parse(window.document.baseUri).removeFragment()}/version.json";
+    var url =
```
jpnurmi, line 20: final?

packages/sensors_plus/example/lib/snake.dart
```diff
@@ -33,17 +33,17 @@ class SnakeBoardPainter extends CustomPainter {
...
       ..color = Colors.black
       ..style = PaintingStyle.fill;
     canvas.drawRect(
       Rect.fromPoints(Offset.zero, size.bottomLeft(Offset.zero)),
       blackLine,
     );
-    for (math.Point<int> p in state.body) {
-      final Offset a = Offset(cellSize * p.x, cellSize * p.y);
-      final Offset b = Offset(cellSize * (p.x + 1), cellSize * (p.y + 1));
+    for (var p in state.body) {
```
jpnurmi, line 44: final?

jpnurmi (approved):
It's a big PR but the changes are very simple. I glanced quickly through and have only some nit-pick. Fix if you feel like it. :)

packages/package_info_plus_macos/pubspec.yaml
miquelbeltran: well spotted, actually there's not even dart code on the macos projects so I think the dependency could be removed

miquelbeltran:
I am not sure why I got some analysis issues on my computer by they don't appear on CI, I created a ticket for that: https://github.com/fluttercommunity/plus_plugins/issues/85

Besides, I am merging this now :)

miquelbeltran:
Thanks for reviewing @jpnurmi !

PR #86, opened 2020-12-19
[network_info_plus] adapted from connectivity_plus

jpnurmi:
A renamed fork of `connectivity_plus` without:
- `checkConnectivity()`
- `onConnectivityChanged`


miquelbeltran (approved):
Thank you! I won't be able to run and test but I trust that it works. Also saw that you already renamed the java and ios classes as well as the method channel name

Issue #96, opened 2021-01-27
Double slash for version.json in package_info_plus_web can prevent version to be read


cfiorillo:
## System info

Issue occurs on: Web
Plugin name: package_info_plus_web
Plugin version: 0.2.1
Flutter doctor output:

```
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel beta, 1.23.0-18.1.pre, on Mac OS X 10.15.7 19H114 x86_64, locale it-IT)
 
[✓] Android toolchain - develop for Android devices (Android SDK version 30.0.2)
[✓] Xcode - develop for iOS and macOS (Xcode 12.3)
[✓] Chrome - develop for the web
[✓] Android Studio (version 4.1)
[✓] VS Code (version 1.51.0)
[✓] Connected device (2 available)

• No issues found!
```

## Steps to Reproduce

When you use PackageInfo.fromPlatform() to get version info, there is an extra (forward) slash before the file name, like **http://localhost:62124//version.json** (port number can change).
If the application is deployed on a web server that doesn't change/redirect/correct the double slash in the URL (we used S3 static content deployment in AWS, that is one of these cases) the browser cannot get the version.json file (error 404) so the version number is not get.

You can get a test application [here on gist](https://gist.github.com/cfiorillo/113dc67e0c979af0365446ac87b4a530).

1. Create a new flutter application
2. Add package_info_plus to pubspec.yaml
3. Add a Text to the main screen populated using PackageInfo.fromPlatform()
4. Run the application using Chrome profile and explore the network from the developer tools
5. You can see the request **http://localhost:62124//version.json** to get the version file (port number could differ) instead of 
the correct http://localhost:62124/version.json

## Suggested fix

https://github.com/fluttercommunity/plus_plugins/blob/92798cfbd217d45933fd9d8d3d5735d86a5ea379/packages/package_info_plus_web/lib/package_info_plus_web.dart#L21

The slash should not be added since `window.document.baseUri` always ends with a slash (see the description for [#32](https://github.com/fluttercommunity/plus_plugins/pull/32#issue-510195074))

Added a PR for this: [97](https://github.com/fluttercommunity/plus_plugins/pull/97)
## Logs

In attach the log of the request from Chrome browser; you can see the double slash in the GET

```
GET //version.json HTTP/1.1
Host: localhost:62124
Connection: keep-alive
[...]

PR #97, opened 2021-01-27
[package_info_plus_web] Fix double slash in version.json lookup

cfiorillo:
## Description

When accessing the version.json file the generated URL contains an extra forward-slash, that can prevent in some deployments access to the file

## Related Issues

https://github.com/fluttercommunity/plus_plugins/issues/96

## Checklist

- [ ] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [ ] I updated/added relevant documentation (doc comments with `///`).

## Breaking Change

### DIFF
--- a/packages/package_info_plus_web/lib/package_info_plus_web.dart
+++ b/packages/package_info_plus_web/lib/package_info_plus_web.dart
@@ -18,7 +18,7 @@ class PackageInfoPlugin extends PackageInfoPlatform {
   @override
   Future<PackageInfoData> getAll() async {
     final url =
-        '${Uri.parse(window.document.baseUri).removeFragment()}/version.json';
+        '${Uri.parse(window.document.baseUri).removeFragment()}version.json';
 
     final response = await get(url);
     if (response.statusCode == 200) {


miquelbeltran:
Thanks @cfiorillo !

Is there a way to have a test for this issue?

Did you run the e2e test for this plugin?

cfiorillo:
Hi @miquelbeltran 
I tried to run the e2e test, but:
1. there are no e2e tests on the package_info_plus_web defined
2. the e2e tests on the package_info_plus fails also if I run them on the main branch.

Regarding point 2, trying the e2e tests on package_info_plus the "expect" for web are wrong; this code (in package_info_plus_e2e.dart file)

```
   } else if (kIsWeb) {
      expect(info.appName, 'package_info_example');
      expect(info.buildNumber, null);
      expect(info.packageName, null);
      expect(info.version, null);
    } else {
```

should be replaced by 

```
    } else if (kIsWeb) {
      expect(info.appName, 'package_info_example');
      expect(info.buildNumber, '4');
      expect(info.version, '1.2.3');
    } else {
```

since the pubspec.yaml file contains:

version: 1.2.3+4

miquelbeltran:
> there are no e2e tests on the package_info_plus_web defined

That's OK, the projects are structured in a way all e2e tests are on the root project :)

> the e2e tests on the package_info_plus fails also if I run them on the main branch.

That's good to know at least, thanks for checking! I am aware that not all e2e test pass on all platforms, so I will fill a ticket for that.

jpnurmi (approved):
Thanks for the fix!

I was going to ask if we should ensure that there's always a single slash between in case the base URI wouldn't end with a slash, but Flutter's `index.html` clearly states that the base URI _must_ end with a slash, so this is ok.

```
  <!--
    If you are serving your web app in a path other than the root, change the
    href value below to reflect the base path you are serving from.

    The path provided below has to start and end with a slash "/" in order for
    it to work correctly.

    Fore more details:
    * https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base
  -->
  <base href="/">
```

cfiorillo:
> Thanks for the fix!
> 
> I was going to ask if we should ensure that there's always a single slash between in case the base URI wouldn't end with a slash, but Flutter's `index.html` clearly states that the base URI _must_ end with a slash, so this is ok.
> 
> ```
>   
>   <base href="/">
> ```

HI @jpnurmi, while preparing the PR I initially added the slash before version.json only if there was no trailing slash in window.document.baseUri, but I read the comment above in index.html and changed mind ;-) 

Thanks for the approval, cool

PR #103, opened 2021-01-30
Fix integration tests on Linux

jpnurmi:


miquelbeltran:
Very nice! way to have e2e tests on CI!

miquelbeltran (approved):

PR #115, opened 2021-02-04
[all] melos analyzer support null safety

miquelbeltran:
Small change to setup null safety in the analyzer used by melos.

Two exceptions need to be done for the example project in android_intent_plus and in package_info_plus

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -4,9 +4,11 @@
 #
 # include: ../../analysis_options.yaml
 #
-include: package:pedantic/analysis_options.1.9.0.yaml
+include: package:pedantic/analysis_options.1.10.0-nullsafety.yaml
 
 analyzer:
+  enable-experiment:
+    - non-nullable
   exclude:
     # Ignore generated files
     - '**/*.g.dart'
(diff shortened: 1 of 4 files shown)


jpnurmi (approved):
Just so that I understand, do we need to add those temporary analysis options to each example when migrating the respective package to null-safety? And then we can clean up and remove the temporary analysis options once all packages have been migrated?

miquelbeltran:
> Just so that I understand, do we need to add those temporary analysis options to each example when migrating the respective package to null-safety? 

Kind of, only two examples seem to be affected by this issue. The example in `package_info_plus` complained while `package_info_plus` has not been migrated yet (???) and the example inside `android_alarm_manager_plus` didn't complain, but the package is migrated. So it may be it is just those two.

> And then we can clean up and remove the temporary analysis options once all packages have been migrated?

We can clean them up when we migrate the example to null safety, at the moment the blocker is the `integration_test` package, which will come as part of the SDK rather than a package in pub.dev.

jpnurmi:
Oh ok, thanks! Good to know that it's about the example itself, not the other packages.

PR #117, opened 2021-02-04
[device_info_plus] add macOS support (kudos to Nash)

jpnurmi:
A `device_info_plus_macos` implementation by @Nash0x7E2 was lying in an old feature branch. I rebased it and finished up the last remaining bits and pieces.


miquelbeltran (dismissed):
Thanks! I just verified that the changelogs and pubspecs imports are correct, I cannot verify if the code is right.

jpnurmi:
Thanks! Here's a screenshot btw :smile: 

device_info_plus_macos

I'm also not fluent in Swift but the implementation looked complete so I thought why not finish the remaining cosmetic things so we can tick off another item in the table... :wink:

miquelbeltran:
awesome, thanks!

mhadaily (approved):
LGTM

PR #120, opened 2021-02-05
Fix #119

### LINKED ISSUE #119: [all] Fix new warnings detected by Flutter Analyze
- 'RaisedButton' is deprecated and shouldn't be used. Use ElevatedButton instead.
- A value of type 'void' can't be returned by the 'onError' handler because it must be assignable to 'FutureOr<...>'

miquelbeltran:
Fixes analysis issues found in #119

### DIFF
--- a/melos.yaml
+++ b/melos.yaml
@@ -89,4 +89,4 @@ dev_dependencies:
 
 environment:
   sdk: ">=2.12.0-0 <3.0.0"
-  flutter: ">=1.12.13+hotfix.5 <2.0.0"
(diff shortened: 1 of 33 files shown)


miquelbeltran:
A new issue appeared:

```
* The Flutter constraint should not have an upper bound.
  In your pubspec.yaml the constraint is currently `>=1.12.13+hotfix.5 <2.0.0`.
  
  You can replace that with just the lower bound: `>=1.12.13+hotfix.5`.
```

jpnurmi (approved):
Thanks! :+1:

PR #131, opened 2021-02-11
[device_info_plus] fix macos support

jpnurmi:
MacOsDeviceInfo export was missing in the platform interface.

### DIFF
--- a/packages/device_info_plus/CHANGELOG.md
+++ b/packages/device_info_plus/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.7.1
+
+- Fix macOS support.
+
 ## 0.7.0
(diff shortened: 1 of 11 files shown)


jpnurmi:
I must have missed some of the changes while staging changes for the earlier PR. It's a bit confusing on macOS because there are tons of ID changes etc. to all kinds of Xcode project and workspace files... Anyway, the question is, how did it pass the CI in the first place?

miquelbeltran:
> Anyway, the question is, how did it pass the CI in the first place?

I think it is not even tested :(

miquelbeltran (approved):
Thanks for noticing and fixing!

PR #134, opened 2021-02-12
[battery_plus_android] send initial battery status

jpnurmi:
Desktop & web already send the current battery status when the stream is opened. Given that there is no battery status getter in the API, this seems a reasonable approach to avoid "null" battery status on startup.


miquelbeltran (approved):
Nice improvement, thanks!

mhadaily (approved):
LGTM

Issue #50, opened 2020-12-05
[all] Migrate to Null Safety


miquelbeltran:
Migrate all the plugins to null safety.

Some considerations:
- e2e package is discontinued and replaced by integration_test.
- integration_test package does not support null safety yet.
- migrate only the plugin, and not the example or the e2e test yet.

TODO
- [x] Enable Beta on CI #106 
- [x] battery_plus @miquelbeltran 
- [x] connectivity_plus @mzdm
- [x] device_info_plus @Schwusch 
- [x] network_info_plus @jpnurmi
- [x] package_info_plus @jpnurmi 
- [x] sensors_plus @jpnurmi  
- [x] share_plus @miquelbeltran 
- [x] android_alarm_manager_plus @miquelbeltran 
- [x] android_intent_plus @miquelbeltran 
- [x] Migrate examples & opt out integration test files (https://github.com/fluttercommunity/plus_plugins/issues/50#issuecomment-782154014) @jpnurmi

Tag yourself on each package if you start working on them!

miquelbeltran:
Currently `battery_plus`, `connectivity_plus` and `network_info_plus` are blocked due the dependency on `dbus`.

Ticket: https://github.com/canonical/dbus.dart/issues/119

miquelbeltran:
the dbus team has migrated the codebase to null safety and is ready for a release, once the release is done we can continue the null safety migration work.

mhadaily:
@miquelbeltran, so we technically need to migrate device_info_plus, package_info_plus,  sensors_plus, the rest were migrated?

miquelbeltran:
The following need to be migrated and are still pending:
battery_plu
 connectivity_plus
 device_info_plus
 package_info_plus
 sensors_plus

Any help is appreciated!

mhadaily:
Sure,  you are doing battery plus right?

then I'll take package_info_plus + sensors_plus for now.

mzdm:
I would like to help and give an attempt to migrate `connectivity_plus` plugin to null safety.

mhadaily:
@mzdm it's all yours, please let us know how it goes. 

Thank you for your contribution.

jpnurmi:
@mzdm I think `connectivity_plus` is still waiting for [dbus](https://github.com/canonical/dbus.dart/pull/130), but you could temporarily use it straight from the Git repo for migration purposes.

jlubeck:
[...]

PR #136, opened 2021-02-15
[device_info_plus] migrate to null safety

Schwusch:
Work for #50 migrated device_info to null safety

### DIFF
--- a/packages/device_info_plus/lib/device_info_plus.dart
+++ b/packages/device_info_plus/lib/device_info_plus.dart
@@ -40,13 +40,10 @@ class DeviceInfoPlugin {
   // of dart plugins is implemented.
   // See https://github.com/flutter/flutter/issues/52267 for more details.
   static DeviceInfoPlatform get _platform {
-    var platform = __platform;
-
-    platform ??= !kIsWeb && Platform.isLinux && !_disablePlatformOverride
-        ? DeviceInfoLinux()
-        : DeviceInfoPlatform.instance;
-
-    return __platform = platform;
+    return __platform ??=
+        !kIsWeb && Platform.isLinux && !_disablePlatformOverride
+            ? DeviceInfoLinux()
+            : DeviceInfoPlatform.instance;
   }
 
   /// This information does not change from call to call. Cache it.
--- a/packages/device_info_plus/lib/device_info_plus.dart
+++ b/packages/device_info_plus/lib/device_info_plus.dart
@@ -34,20 +34,23 @@ class DeviceInfoPlugin {
   }
 
   static bool _disablePlatformOverride = false;
-  static DeviceInfoPlatform __platform;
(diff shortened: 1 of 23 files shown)


Schwusch:
Weird, I don't get analysis errors locally, but I run the `dev` channel locally, so it might be it.

packages/device_info_plus/lib/device_info_plus.dart
```diff
@@ -34,20 +34,23 @@ class DeviceInfoPlugin {
...
   static DeviceInfoPlatform get _platform {
-    __platform ??= !kIsWeb && Platform.isLinux && !_disablePlatformOverride
+    var platform = __platform;
+
+    platform ??= !kIsWeb && Platform.isLinux && !_disablePlatformOverride
         ? DeviceInfoLinux()
         : DeviceInfoPlatform.instance;
-    return __platform;
+
+    return __platform = platform;
```
jpnurmi, line 49: does this actually need any changes? wouldn't the old assignment work with nullable __platform or did i miss something?

Schwusch: We can do it the old way and return `DeviceInfoPlatform?` instead, but if we want return a non-nullable type, the compiler can't guarantee that the class member is not null. Hence the local variable.

jpnurmi: or return `__platform!`?

Schwusch: Ouch, yeah, we could do that as well. I am hesitant to use `!` though, since it's hard to trust the compiler if we do that often. Just a conclusion I've learned from migrating _a lot_ of Java to Kotlin and gotten burnt in runtime.

jpnurmi:
Looks like #116 didn't migrate the example. Would that help to pass the analysis?

packages/device_info_plus/lib/device_info_plus.dart
jpnurmi: There's an unconditional assignment above so what could possibly go wrong? :)

https://github.com/fluttercommunity/plus_plugins/pull/116/files#diff-1ddc6908e37618bbc4937f9caad0333cfe733aba085bfba4e6e1cf3857271990

Schwusch: If that is the preferred way, sure.

Schwusch: I found a way that is safe and small :) 
```dart
 return __platform ??= !kIsWeb && Platform.isLinux && !_disablePlatformOverride
        ? DeviceInfoLinux()
        : DeviceInfoPlatform.instance;
```

jpnurmi (approved):
LGTM :+1:

jpnurmi (approved):

Issue #113, opened 2021-02-04
Improving the documentation website


pr-Mais:
This issue will be tracking the progress in improving the documentation website [plus.fluttercommunity.dev](https://plus.fluttercommunity.dev).

### TODO list

- [x] Add getting started section with overview of Plus plugins
- [x] Battery➕ docs
- [x] Connectivity➕ docs
- [x] Device Info➕ docs
- [x] Network Info➕ docs
- [x] Package Info➕ docs
- [x] Sensor➕ docs
- [x] Share➕ docs
- [x] Android Alarm Manager➕ docs
- [x] Android Intent➕ docs

mhadaily:
All docs are up now https://plus.fluttercommunity.dev/docs/overview

PR #137, opened 2021-02-15
Battery Plus website docs

pr-Mais:
Part of #113 

Website docs for `battery_plus` plugin, with a small change in the inner API docs.


docs/battery/usage.mdx
```diff
@@ -0,0 +1,58 @@
...
+    // Each time the battery state changes, the local state of the widget is updated
+    setState(() {
+      _batteryState = state;
+    });
+  });
+}
+```
+
+:::note
+Trying to run this code on *Android emulator* may result in a null battery state with a `PlatformException`, but Android emulator provides a way of emulating the different states from its settings. By clicking the menu icon (three vertical dots) then going to battery section.
```
jpnurmi, line 47: > may result in a null battery state

This will be fixed by #134 (initial state for Android) and https://github.com/jpnurmi/plus_plugins/commit/ed049b0536713b09c902615f4ad358be45a99970 (missing "unknown" state).

mhadaily (approved):
LGTM, Nice one

docs/battery/usage.mdx
pr-Mais: So there will be a forth state "Unknown" ?

jpnurmi: Yeah, exactly. :+1: I tried both Android & iOS and couldn't reproduce the null state anymore.

jpnurmi (approved):
Thanks! 🙂

Issue #16, opened 2020-10-12
Go though all published packages and see what we need to do to up score to 110.


slightfoot:
Packages:
- [x] battery_plus 100
- [x] connectivity_plus  100
- [x] device_info_plus 110
- [x]  network_info_plus 100
- [x] package_info_plus 110
- [x] sensors_plus 110
- [x] ~share_plus 100~ ([2.1.0-nullsafety.0](https://pub.dev/packages/share_plus/versions/2.1.0-nullsafety.0/score) has 110)
- [x] android_alarm_manager_plus 80
- [x] android_intent_plus 90 (android only)

mhadaily:
We have now the highest score possible. 💯

PR #140, opened 2021-02-16
[battery_plus] address pub.dev score (#16) 

jpnurmi:
Importing `platform_interface/src/foo.dart` in the main package loses 10 pub score points:

    Pass static analysis

    20/30 points: code has no errors, warnings, lints, or formatting issues

    INFO: Don't import implementation files from another package.

### DIFF
--- a/packages/battery_plus/CHANGELOG.md
+++ b/packages/battery_plus/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.10.1
+
+- Address pub score
+
 ## 0.10.0
 
 - Added "unknown" battery state for batteryless systems.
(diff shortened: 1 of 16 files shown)


jpnurmi:
- [`path_provider`]( https://pub.dev/packages/path_provider/score) has the same problem.
- [`url_launcher`](https://pub.dev/packages/url_launcher/score) doesn't lose points even if it has similar import (https://github.com/flutter/plugins/blob/master/packages/url_launcher/url_launcher/lib/url_launcher.dart) because the platform_interface structure is flat: https://github.com/flutter/plugins/tree/master/packages/url_launcher/url_launcher_platform_interface/lib

mhadaily (approved):
LGTM

PR #141, opened 2021-02-16
[connectivity_plus] address pub score (#16)

jpnurmi:
Importing `platform_interface/src/foo.dart` in the main package loses 10 pub score points:

    Pass static analysis

    20/30 points: code has no errors, warnings, lints, or formatting issues

    INFO: Don't import implementation files from another package.


mhadaily (approved):
LGTM

PR #142, opened 2021-02-16
[website] minor tweaks to make it more mobile friendly

jpnurmi:
Make the table a bit narrower to make it fit mobile screens better. These tweaks make all columns almost fully visible in landscape mode on Pixel 5. In portrait mode, it's not possible to scroll all the way to the left to see the package name, but I have no idea how to fix that. :)

### DIFF
--- a/website/plugins.js
+++ b/website/plugins.js
@@ -99,7 +99,7 @@ module.exports = [
     },
   },
   {
-    name: 'Sensor ➕',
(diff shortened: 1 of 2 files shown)


jpnurmi:
| Before | After |
|---|---|
| Screenshot_20210216-131207 | Screenshot_20210216-131147 |

mhadaily (approved):
Nice thanks

PR #155, opened 2021-02-18
[package_info_plus] get rid of mockito & port to latest win32/ffi

jpnurmi:
Preparing for null-safety migration.

### DIFF
--- a/packages/package_info_plus_platform_interface/pubspec.yaml
+++ b/packages/package_info_plus_platform_interface/pubspec.yaml
@@ -13,7 +13,6 @@ dependencies:
 dev_dependencies:
   flutter_test:
     sdk: flutter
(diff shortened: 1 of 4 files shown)


mhadaily (approved):
Don’t we really need mockito ? Or until they support null safety

jpnurmi:
> Don’t we really need mockito ? Or until they support null safety

The old mockito is not null-safety compatible, but we can generate null-safe mocks with mockito 5.

The test was using a mock to test that a class that must be extended doesn't work if it's implemented instead. It didn't seem like a super useful test, so I thought dropping it would be reasonable and avoid the null-safety hassle with mockito.

Nevertheless, I left out the mockito change from this PR so we can get the more important win32/ffi changes in. :)

PR #162, opened 2021-02-21
[all] cleanup and ignore generated files in Linux & Windows runners

jpnurmi:
These files are (re-)generated by flutter build/run.

P.S. The formatting of these files doesn't match 100% with clang-format's Google style, so removing these files helps a bit with unrelated format changes we've been experiencing.


miquelbeltran (approved):
very nice, thanks!

PR #168, opened 2021-02-28
[network_info_plus] migrate to null-safety (#50)

jpnurmi:


mhadaily (approved):
LGTM, this was the last piece of the puzzle, thank you @jpnurmi so my next PR for documentation doesn't have anything missing now

jpnurmi:
Thanks! I almost missed this because it didn't exist when the null-safety migration task #50 was created so it was missing from the list. :smile:

Issue #173, opened 2021-03-02
make purpose of plus_plugins more clear in READMEs


nt4f04uNd:
<strike>
firsly - the plugins in this repository are forks, so it should be clear indicated what functionalities these suggest opposed to their origins
</strike>

<br/>
<br/>

<strike>
secondly - it isn't clear for me in general why wouldn't you make PRs improving the official https://github.com/flutter/plugins and just forked them. this way instead of improving and building as a community one solid thing, you are splitting the community into those who want to use and contribute to your version of plugins and those who would like to stick to officials
</strike>

<br/>
<br/>

<strike>
and now if someone suggests feature to official plugin you will end up copying their code (same applies in the opposite direction).
not to say that some PR that just improves the existing implementation might be simply missed
</strike>

nt4f04uNd:
i stumbled upon https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md#important-note which really clarifies things

the poitns a made in OP are no longer valid
the only thing that has to be added to READMEs is something similar to what said in this contributing section

miquelbeltran:
Oh I totally missed the note on the contributing.md, thanks for finding it out!

I've added a short note on the README, I'll be happy if you can take a look at the PR and give us your review

nt4f04uNd:
wouldn't it make sense to add this note to all plugins readmes?

nt4f04uNd:
and also, any particular reason why readmy converted table from https://github.com/flutter/plugins readme into a long list
this looks significantly worse imo

miquelbeltran:
IMO the flutter/plugins readme doesn't provide any useful info besides the link to pub.dev.

Agreed on updating the individual readmes. You are free to submit a PR with the changes as well!

mhadaily:
I added this note to all main README files. we don't need to add to all readme files. main plugins are enough I believe https://github.com/fluttercommunity/plus_plugins/pull/283

nt4f04uNd:
yep, i actually meant all main plugin readmes, not literally all =)

mhadaily:
Addressed already.

PR #174, opened 2021-03-04
Update README.md

miquelbeltran:
Added info on Plus Plugins vs. original flutter/plugins as reported in https://github.com/fluttercommunity/plus_plugins/issues/173


mhadaily (approved):
Thanks

PR #175, opened 2021-03-05
[all] fix linter warnings and bump SDK constraints to pass analysis

jpnurmi:


miquelbeltran (approved):
Thanks for fixing this and cleaning up all the sdk versions!

Issue #201, opened 2021-04-08
Disable jumping to "tomail:" when refusing to "Share" on a mobile device:: share_plus_web


u0690b:
## Use case

when refusing to "Share" on a mobile device then jumped to mail.
but i don't want it

## Proposal

plus_plugins/packages/share_plus_web/lib/share_plus_web.dart

```
  /// Share text
  Future<void> share(
    String text, {
    String subject,
    Rect sharePositionOrigin,
  }) async {
    try {
      await _navigator.share({'title': subject, 'text': text});
    } catch (e) {
     // i want to new condition
      if(e.toString()!="AbortError: Share canceled"){   //<<---- this line
        final uri = Uri.encodeFull('mailto:?subject=$subject&body=$text');
        return launch(uri);
     }
}
```

thiagops:
This seems to be a bug. We should not open the email composer if the user cancels the sharing action.

miquelbeltran:
Ah I see what you mean now, this line here: https://github.com/fluttercommunity/plus_plugins/blob/main/packages/share_plus_web/lib/share_plus_web.dart#L33 is always going back to `mailto` also when the user cancels.

PRs with the fix welcome :)

Is there a better way than to compare the toString() of the error?

u0690b:
> Ah I see what you mean now, this line here: https://github.com/fluttercommunity/plus_plugins/blob/main/packages/share_plus_web/lib/share_plus_web.dart#L33 is always going back to `mailto` also when the user cancels.
> 
> PRs with the fix welcome :)
> 
> Is there a better way than to compare the toString() of the error?

 /// Share text
  Future<void> share(
    String text, {
    String? subject,
    Rect? sharePositionOrigin,
  }) async {
    try {
      await _navigator.share({'title': subject, 'text': text});
    } on NoSuchMethodError catch (e) {
      //Navigator is not available or the webPage is not served on https
      final uri = Uri.encodeFull('mailto:?subject=$subject&body=$text');
      await launch(uri);
    } catch (e) {
      //Navigator share cancel
    }
  }

miquelbeltran:
definitely! feel free to open a PR

miquelbeltran:
fix released in share_plus_web 2.0.2

PR #232, opened 2021-05-10
catch handle  if the user cancels #201

u0690b:
## Description

We should not open the email composer if the user cancels the sharing action.

## Related Issues

 [issue database](https://github.com/fluttercommunity/plus_plugins/issues/201)


miquelbeltran (approved):

miquelbeltran:
Thanks! as soon as CI passes I will merge

miquelbeltran:
the change should be available in share_plus_web 2.0.2, please run a flutter pub upgrade and let me know if it worked :) thanks for the help!

Issue #255, opened 2021-05-27
deviceMemory: Expected a value of type 'int', but got one of type 'Null'


KirillArtemenko:
I have an error **TypeErrorImpl (Expected a value of type 'int', but got one of type 'Null')** when i trying to get webBrowserInfo on web version of my app:
`final WebBrowserInfo webBrowserInfo = await deviceInfoPlugin.webBrowserInfo;`

deviceMemory value is None and it produces an error: ""

This what i got on "map" variable inside WebBrowserInfo.fromMap(map) method:
![2021-05-27_14-33-23](https://user-images.githubusercontent.com/667503/119819631-404f3900-bef9-11eb-8d14-eb16481fe19d.png)

As we can see "deviceMemory" -> null and it's produce an error.

this simple fix is solve my problem, maybe you need to add this:
```
deviceMemory: map['deviceMemory'],
// replaced to
deviceMemory: map['deviceMemory'] ?? 0,
```

Flutter version: 2.2.0
Issue occurs on: Web
Plugin name: device_info_plus
Plugin version: 1.0.1
Flutter doctor output: all is ok

miquelbeltran:
Thanks for the report @KirillArtemenko ! PRs with the fix are more than welcome.

KirillArtemenko:
> Thanks for the report @KirillArtemenko ! PRs with the fix are more than welcome.
PR ready!
https://github.com/fluttercommunity/plus_plugins/pull/256

KirillArtemenko:
I created another PR, it is better coz more backward compatible and no need any changes for ios/android of plugin developers.

https://github.com/fluttercommunity/plus_plugins/pull/257

miquelbeltran:
Version 2.0.0 with the fixes are published

KirillArtemenko:
Thank you but still not available here https://pub.dev/packages/device_info_plus
Update please.

miquelbeltran:
It's up now, I had an issue with the publishing script

PR #256, opened 2021-05-27
sometimes deviceMemory is null on web

KirillArtemenko:
more info here:
https://github.com/fluttercommunity/plus_plugins/issues/255

### DIFF
--- a/packages/device_info_plus_platform_interface/lib/model/web_browser_info.dart
+++ b/packages/device_info_plus_platform_interface/lib/model/web_browser_info.dart
@@ -112,7 +112,7 @@ class WebBrowserInfo {
       appCodeName: map['appCodeName'],
       appName: map['appName'],
       appVersion: map['appVersion'],
-      deviceMemory: map['deviceMemory'],
+      deviceMemory: map['deviceMemory'] ?? 0,
       language: map['language'],
       languages: map['languages'],
       platform: map['platform'],


KirillArtemenko:
I created another PR, it is better coz more backward compatible and no need any changes for ios/android of plugin developers.

https://github.com/fluttercommunity/plus_plugins/pull/257

miquelbeltran:
Thanks for the fix @KirillArtemenko !

I am thinking that maybe it makes more sense to make all the properties in `WebBrowserInfo` nullable instead, because we really don't have any warranties that the "navigator" will provide all or any of those fields.

But I am happy to just merge this as a temporal solution. Could you also bump the build number in the pubspec.yaml and add an entry on CHANGELOG as explained in the contributing document?

KirillArtemenko:
> Thanks for the fix @KirillArtemenko !
> 
> I am thinking that maybe it makes more sense to make all the properties in `WebBrowserInfo` nullable instead, because we really don't have any warranties that the "navigator" will provide all or any of those fields.

I thought about it, but this change will entail a change to the main interface. And other developers will need to update their code due to the fact that some of the fields will become nullable. So I decided another this solution.

KirillArtemenko:
> But I am happy to just merge this as a temporal solution. Could you also bump the build number in the pubspec.yaml and add an entry on CHANGELOG as explained in the contributing document?

Please make the PR and changes yourself. Sorry for the mess. I'm creating this PR from the internet, and right now I can't create it properly. Next time I'll create a normal PR from my local git. Take my word for it )

PR #257, opened 2021-05-27
add default values for all nullable

KirillArtemenko:
Add default values for all nullable properties, without it we can get NPE at WebBrowserInfo.fromMap() at the main device_info_platform_interface plugin
more info here:
https://github.com/fluttercommunity/plus_plugins/issues/255

PR #258, opened 2021-05-27
nullable must be nullable

KirillArtemenko:
Original issue https://github.com/fluttercommunity/plus_plugins/issues/255. After some not quite correct PR's (https://github.com/fluttercommunity/plus_plugins/pull/256 and https://github.com/fluttercommunity/plus_plugins/pull/257). There was a more correct solution, nullable fields must be nullable and let the developers decide exactly how to handle null value. This solution is ideologically more correct than setting the default values when they are null. I set default value for the "languages" property as a list with a single non-zero value, because I think this is logically correct and expected.


KirillArtemenko:
Pubspec.yaml and Changelog updated.

packages/device_info_plus_platform_interface/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.0.3
```
stuartmorgan, line 1: This isn't following semver; changing a return value from non-nullable to nullable is a breaking change.

miquelbeltran: Yes, and also requires to release the main package and all the subpackages because the `device_info_plus_platform_interface` is going to be incompatible. Yay federated plugins 🤦 

There's no much you can do @KirillArtemenko we need to take it from here.

KirillArtemenko: Perhaps then you should temporarily use my previous solution https://github.com/fluttercommunity/plus_plugins/pull/257 ?

mhadaily:
can you also please merge with `main`

miquelbeltran:
Going to try to unblock this now, I may create a new branch tho

PR #254, opened 2021-05-26
fix(share_plus): Present Share Sheet on the top ViewController 

searchy2:
## Description

Update iOS share target to present on the top ViewController. This fixes "Unable to present" errors when the app is already presenting such as in an add to app scenario.

## Related Issues

https://github.com/fluttercommunity/plus_plugins/issues/251

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`).
This will ensure a smooth and quick review process. Updating the `pubspec.yaml` and changelogs is not required.

- [X] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [X] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
No tests need to be updated. 
- [X] All existing and new tests are passing.
- [X] I updated/added relevant documentation (doc comments with `///`).
No doc changes needed.
- [X] The analyzer (`flutter analyze`) does not report any problems on my PR.
`warning - The path F:\Development\FlutterPlusPlugins\packages\share_plus is not posix - example\pubspec.yaml:9:11 - path_not_posix`
- [X] I read and followed the [Flutter Style Guide].
- [X] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [X] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/fluttercommunity/plus_plugins/blob/main/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning

PR #307, opened 2021-06-15
fix(share_plus): Present Share Sheet on the top ViewController (continuation)

### LINKED ISSUE #251: Share plugin does not work on iOS when it is invalid for a view controller to be presented on iOS's root view controller
Copied from the original share plugin at https://github.com/flutter/flutter/issues/67103
There is a PR to resolve the issue that was rejected because they stopped taking PRs on the original repo.

https://github.com/flutter/plugins/pull/3105

The solution looks good. 
What is needed to get the fixes merged into this library?

miquelbeltran:
Continue work from #254 

Infinite thanks to @searchy2 for the work done in #254

Fix #251


mhadaily (approved):
LGTM, thank you @miquelbeltran and @searchy2

searchy2:
You're welcome! Thanks for the merge :)

PR #365, opened 2021-07-13
doc: Callback needs to be static

thomassth:
I can't make the code from readme work, but the example app works.

By comparing the two, somehow `static` is missing in the example in readme and the alarm won't set without it.


mhadaily (approved):
LGTM, thanks for the PR

PR #371, opened 2021-07-16
update all deps for package info

mhadaily:
Just updating packages versions for package info plus to be up-to-date


jpnurmi (approved):
LGTM :+1:

Issue #279, opened 2021-06-06
Hello, When I create a web application, if it is connected with ethernet it returns ConnectivityResult.none


astudilloalex:
## System info

Issue occurs on: Web
Plugin name: connectivity_plus
Plugin version: 1.0.2
Flutter doctor output: 

```
[√] Flutter (Channel stable, 2.2.0, on Microsoft Windows [Version 10.0.19042.985], locale es-EC)
[!] Android toolchain - develop for Android devices (Android SDK version 30.0.2)
    X Android license status unknown.
      Run `flutter doctor --android-licenses` to accept the SDK licenses.
      See https://flutter.dev/docs/get-started/install/windows#android-setup for more details.
[√] Chrome - develop for the web
[√] Android Studio (version 4.1.0)
[√] Connected device (2 available)
```

## Steps to Reproduce

1.  Only use the example on the official website: https://pub.dev/packages/connectivity_plus/example
2. Return ConnectivityResult.none when I use ethernet cable, with wifi work correctly

mhadaily:
@astudilloalex we have released a new version, can you give it a try?

astudilloalex:
> @astudilloalex we have released a new version, can you give it a try?

```
class NetworkSensitive extends StatelessWidget {
  final Widget child;

  NetworkSensitive({required this.child});

  @override
  Widget build(BuildContext context) {
    final ConnectivityResult result = Provider.of<ConnectivityResult>(context);
    if (result == ConnectivityResult.none) {
      return _ConnectionError();
    }
    return child;
  }
}
```
Hello, I have tried the new version, but the error continues since it keeps returning ConnectivityResult.none, when I use the ethernet cable to connect, even sometimes with the Wifi this error appears but then it is solved.

This happens only when I use the Web; Hello, I have tried the new version, but the error continues since it keeps returning ConnectivityResult.none, when I use the ethernet cable to connect, even sometimes with the Wifi this error appears but then it is solved.

This occurs only when I use the Web, it works perfectly on the mobile.

mhadaily:
@astudilloalex  we have a new pR https://github.com/fluttercommunity/plus_plugins/pull/399 and release which is fixing this issue, please test and if that fixes it please close this issue. thanks

jpnurmi:
#399 added ethernet support for Android, Linux, and Web, and https://github.com/fluttercommunity/plus_plugins/pull/417 added support for macOS.

PR #399, opened 2021-08-08
Add ethernet as connectivity result

jpnurmi:
So far, implemented for Android, Linux, and Web.


mhadaily:
@jpnurmi This looks good to me, but what do you want to do for Web and iOS, and Windows? should we just return null or an error?

jpnurmi:
Android, Linux, and Web have been implemented. I also have a working prototype for macOS, but it will follow in a separate PR. On iOS, we can probably use the same.

So basically the only remaining platform is Windows, and it seems to be a real pain. Perhaps we can get some Windows expert to help with that.

mhadaily (approved):
LGTM

mhadaily:
We have another PR related to connectivity, Maybe we should merge this after that one. although they are not related.

mhadaily:
@jpnurmi feel free to resolve conflict and merge and release.

Elias8:
I think this closes #279

PR #417, opened 2021-08-20
[connectivity_plus_macos] add ethernet support

jpnurmi:
Add ethernet as connectivity result - continuation of https://github.com/fluttercommunity/plus_plugins/pull/399 for macOS.

NOTE: The functionality is provided by [`NWPathMonitor`](https://developer.apple.com/documentation/network/nwpathmonitor), which is available since macOS 10.14. Therefore the minimum macOS deployment target has been changed from 10.11 to 10.14.


mhadaily (approved):
LGTM

Issue #421, opened 2021-08-21
[android_alarm_manager_plus] Reschedule or clear alarms when app is launched after force stop


gazialankus:
## Use case

When we force stop the app, its alarms are cleared and they won't fire. However, these alarms that did not fire still show up in `AndroidAlarmManager.getScheduledAlarms()`. They stay in that list indefinitely because they never actually fire. 

## Proposal

When the user starts the app after a force stop, we can check the scheduled alarms, remove the ones that are in the past and reschedule the ones that are in the future. This way, a force stop followed by app launch would not prevent previously set alarms from firing.

miquelbeltran:
mmhh true, I haven't thought about this edge case. I think we should revert https://github.com/fluttercommunity/plus_plugins/pull/384 probably then because this can lead to misunderstandings.

Chinmay-KB:
Yup, this is not something I had thought about.  This will need more deliberations.

gazialankus:
I guess past alarms should be rescheduled rather than deleted. Alarms that are set for dates in the past are fired as soon as possible. It could be that the alarm never actually fired and would have fired soon if the force stop had not happened. 

So maybe rescheduling all alarms at every start of the app would cover everything. Rescheduling the same alarm with the same id has no effect in Android if it was already there. 

- App stopped normally. App starts and reschedules everything with the same ids, no change.
- App is force stopped. App starts and reschedules everything, the cleared alarms are revived, as if the app was never force-stopped.

miquelbeltran:
> I guess past alarms should be rescheduled rather than deleted.

That's not how the Android OS framework works unfortunately. We shouldn't reinvent how alarms work on the OS level. The feature @Chinmay-KB was a nice addon to keep track of Alarms but personally I think we are trying to overcome the OS limitations which shouldn't be the goal of this plugin. Rather, that should be managed by the app developer on the app level.

gazialankus:
Fair enough. But I do believe this example could be improved: 

I just set up an alarm with id 11, force closed, set up one more alarm with the same id (11) and force closed. When I open the app again, I have two entries in the list and I can't ever get rid of them. 

1. It would be nice if I saw only one entry for alarms I set with the same id. I would see only the second one in the above example. So, alarms set with the same id should overwrite each other in this list. 
2. Optionally, it would be nice if I can clear this list. If I know I was force stopped this would be useful. Not sure if I can know that, though. So I'm not sure if this is a must.

[...]

PR #384, opened 2021-07-23
feat: get scheduled alarms functionality added.

Chinmay-KB:
## Description

The plugin now stores all the upcoming alarms, be it one shot or periodic and returns a json string containing info about all the
scheduled alarms.  Oneshot alarms are removed from this list when the alarm goes off. Peroidic alarms are removed only when it is explicitly cancelled.

## Related Issues

#342 

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`).
This will ensure a smooth and quick review process. Updating the `pubspec.yaml` and changelogs is not required.

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ ] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [x] All existing and new tests are passing.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [x] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/fluttercommunity/plus_plugins/blob/main/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning

PR #429, opened 2021-08-24
Revert #384

miquelbeltran:
## Description

In the PR #384 we added a feature to list the scheduled alarms. However, this feature doesn't follow the OS convention on how Alarms work and causes misunderstandings as discussed in https://github.com/fluttercommunity/plus_plugins/issues/421.

For that, I took the decision of reverting that feature and releasing a new version without it.

At the end, is on the app developer responsibility to manage the alarms and NOT on the plugin side.

## Related Issues

- #384
- #421


mhadaily (approved):
LGTM, thanks for taking care of this.

PR #595, opened 2021-11-03
[battery_plus_linux] get rid of the _real vs. _stub hack

jpnurmi:
This hack is no longer required as dbus.dart has a conditional dart:ffi
import (since 0.3.2) to avoid breaking web as an indirect dependency.


miquelbeltran:
LGTM, feel free to merge this and the other PRs related!

miquelbeltran (approved):

PR #665, opened 2021-12-22
feat(android_intent_plus): ignoring apps in intent

NikulinEB:
## Description

This update allows to specify apps (by their package names) that should not be used to resolve an intent and should not be displayed in the chooser.
Next, I will describe the situation in which I needed these changes. Perhaps it will be useful to someone else, which is why I am creating this PR.
It became necessary for me when opening the links of the Steam store. The problem is that the Steam application appears in the chooser, but if you select it, then the first time a gray screen appears for a moment (probably an attempt to open a link), and next times nothing happens at all (tap on Steam, the chooser disappears and that's it). Therefore, I found it necessary to remove Steam from the list of applications in the chooser, so as not to upset users. And if the user wants to open the link in the Steam app, and not in the browser, then I do it with explicit intent.

## Related Issues

No related issues.

## Checklist

## Breaking Change


piotrsed (approved):

neiljaywarner:
@piotrsed  this will help us with facebook and whatsapp which don't work

vbuberen:
Finally got to this PR. Thanks for contribution, but after some testing I see that it doesn't work for `launchChooser` and breaks `launch`. I will rework this PR.

vbuberen:
Here is how I tested:
1. Re-used one of functions in example app to show all apps that can handle ACTION_SEND and saw multiple apps in my chooser:
Screenshot 2022-10-02 at 14 02 36

2. Added new `ignoredPackages` with package name for Gmail and no chooser appears as well as not log entries about something going wrong:
Screenshot 2022-10-02 at 14 02 25

vbuberen:
It seems that the suggested implementation causes issues and can't be merged.

1. `launch()` is broken in case user specifies `ignoredPackages` parameter because new added code picks first element of the list which can handle the intent, ignoring the fact that it can be something not suitable (in my case I saw some system bluetooth package receiving that SEND intent a screenshot above).

2. `launchChooser()` is broken as well due to the same logic and causes the phone to show `No apps can handle this intent` even for those, which can be handled for sure.

I researched for options on ignoring specific packages and there is no universal solution, especially for intents called via `launch()` in this plugin. So closing this PR and will try to play around the topic later. In case there is a success I will create a separate PR.

PR #703, opened 2022-01-16
[device_info_plus] fix LinuxDeviceInfo.name docs

jpnurmi:


vbuberen (commented):
Looks good to me. However, I am not sure if we need to do a `+1` bump for such a small change, considering that it doesn't change any of APIs.

jpnurmi:
Thanks. A version bump is needed to be able to publish on pub and get online docs generated.

vbuberen (approved):

PR #720, opened 2022-02-01
Update Readme

aydemiromer:
Undefined name fixed. Added var for the subscription.


Elias8:
Hi @aydemiromer 👋,

Thank you for the PR. However, the sample code was intentionally incomplete, and it was written assuming the `subscription` variable is declared somewhere globally. So adding the `var` keyword will make the variable local to the `initState` method and create confusion for the readers.

Thanks again!

PR #745, opened 2022-02-14
[share_plus] fixed message sharing for file selection

parmarravi:
Fixed issue related to message when sharing the files


vbuberen (commented):
Could you describe the issue with more details or provide a link to the existing issue if you tried to address it?

For it me it seems wrong that the same text is used both in `title` and `text` extras as they clearly should be different.

vbuberen:
Closing due to no activity.

PR #765, opened 2022-02-26
[share_plus] add ability to get feedback on user action + 100% coverage

### LINKED ISSUE #186: [share_plus] know when the sharing is finished
I ask if there is a chance to know when the sharing is finished and the sharing window closes.
If this is not possible you could add this callback
Thanks for your valuable work

### LINKED ISSUE #386: share_plus - Give back result of user action
I will propose below a feature for the plugin https://pub.dev/packages/share_plus

Request is:
Give back some enum of user action after share screen is shown.
Currently we have this signature, which returns void
```
  static Future<void> share(
    String text, {
    String? subject,
    Rect? sharePositionOrigin,
  })
```

In many use cases as I will describe in the other section, will be very useful to know what user did in the share screen.

## Use case

1. Tracking. We need to know if user shared the text/file/etc =, or just closed the dialog(view) without doing nothing. This enabled a lot of opportunities, like A/B testing, trying to improve ui (explanation text) for what user is about to share and so on.
2. Follow up logic. Based on whether user shared or not, app might want to follow up with different code paths. Example: thank you for sharing, you received 10 super app points.

## Proposal

The functions will return instead of Future<void> Future<ShareResult>

```
  static Future<Result> share(
    String text, {
    String? subject,
    Rect? sharePositionOrigin,
  })

//where Result can be something like

enum ShareResult {
  shared,
  closed,
  // maybe something else here no idea currently i see big need for binary did share or not  
}
```

### LINKED ISSUE #748: Await for feedback if message shared successfully
Hi everyone,
It would be useful if it was possible to await for result of share action to check, if user successfully shared a message or cancelled whole process.

Coronon:
## Description

As per #386 a lot of users would like to know how the user interacted with the share-sheet, so this PR adds exactly this capability. It does this by providing the two extra methods `shareWithResult` and `shareFilesWithResult` on `Share`, which both return a `ShareResult` which can then be used to determine if the user completed an action or dismissed the dialogue. It also provides the user with information on the specific share action (app/build-in) the user might have taken.

By adding this as two new methods, no other APIs have to change. It would be possible to add this feature as an argument to the "old" `share` and `shareFiles` -> `bool withResult = false`, but that would require updating all other *share_plus* platforms implementation, which could probably break some code somewhere and does not seem that good of an idea if not even all platforms support this feature at the moment.

While adding tests for my two new methods I also rewrote parts of the old tests to make them easier and boost coverage to 100% for _share_plus_platform_interface_.

**This requires Android API version 21:**
This simply can not be implemented without writing our completely own sharing-sheet in API version 16(2012) and it should be reasonable to require version 21+(2014).

**Update! requirement for API version 23:**
API Version 31 made it mandatory to either specify [`FLAG_IMMUTABLE`](https://developer.android.com/reference/android/app/PendingIntent#FLAG_IMMUTABLE)(API v.23) or [`FLAG_MUTABLE`](https://developer.android.com/reference/android/app/PendingIntent#FLAG_MUTABLE)(API v.31). This means that in order to support API version 31 - the most current and desirable - we need to set our lower bound to 23+(2015) which should still be reasonable.

This has been my first time writing Kotlin, Objective C and Swift, so any feedback is greatly appreciated ;)

## Related Issues

Resolves #386, resolves #748, resolves #186.

## Checklist

## Breaking Change

### DIFF
--- a/packages/share_plus/share_plus/CHANGELOG.md
+++ b/packages/share_plus/share_plus/CHANGELOG.md
@@ -1,7 +1,8 @@
 ## 4.0.0
 
-- iOS, Android: Add `shareWithResult` methods to get feedback on user action
+- iOS, Android, MacOS: Add `shareWithResult` methods to get feedback on user action
 - Android: Increased minSdkVersion to 23
+- MacOS: Native sharing implementation
 
 ## 3.1.0
 
--- a/packages/share_plus/share_plus/CHANGELOG.md
+++ b/packages/share_plus/share_plus/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 2.0.2
+## 3.1.1
 
 - Add *WithResult methods to get feedback on user action
 
--- a/packages/share_plus/share_plus/CHANGELOG.md
+++ b/packages/share_plus/share_plus/CHANGELOG.md
@@ -1,13 +1,12 @@
-## 3.2.0
+## 4.0.0
 
-- Add *WithResult methods to get feedback on user action
+- iOS, Android: Add `shareWithResult` methods to get feedback on user action
+- Android: Increased minSdkVersion to 23
 
 ## 3.1.0
 
 - Android: Migrate to Kotlin
 - Android: Update dependencies, build config updates
-- Example: Fix project title
-- Example: Set min Flutter version to 1.20.0
 
 ## 3.0.5
 
--- a/packages/share_plus/share_plus/CHANGELOG.md
+++ b/packages/share_plus/share_plus/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.0.2
+
+- Add *WithResult methods to get feedback on user action
+
 ## 3.1.0
 
 - Android: Migrate to Kotlin
--- a/packages/share_plus/share_plus/CHANGELOG.md
+++ b/packages/share_plus/share_plus/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 3.1.1
+## 3.2.0
 
 - Add *WithResult methods to get feedback on user action
 
--- a/packages/share_plus/share_plus/lib/share_plus.dart
+++ b/packages/share_plus/share_plus/lib/share_plus.dart
@@ -6,6 +6,8 @@ import 'dart:async';
 import 'dart:ui';
 
 import 'package:share_plus_platform_interface/share_plus_platform_interface.dart';
+export 'package:share_plus_platform_interface/share_plus_platform_interface.dart'
+    show ShareResult, ShareResultStatus;
 
 /// Plugin for summoning a platform share sheet.
 class Share {
@@ -82,10 +84,15 @@ class Share {
 
   /// Behaves exactly like [share] while providing feedback on how the user
   /// interacted with the share-sheet. Until the returned future is completed,
-  /// any other call to any share method that returns a result will result in
+  /// any other call to any share method that returns a result might result in
   /// a [PlatformException].
   ///
-  /// Currently only implemented on IOS & Android.
+  /// Because IOS and Android provide different feedback on share-sheet interaction,
+  /// a result on IOS will be more specific than on Android. While IOS can detect if
+  /// the user actually completed his selected action or aborted it midway, Android
+  /// only records if the user selected an action or outright dismissed the share-sheet.
+  ///
+  /// **Currently only implemented on IOS & Android.**
   static Future<ShareResult> shareWithResult(
     String text, {
     String? subject,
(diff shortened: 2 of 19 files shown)


Coronon:
I will add tests the coming week (probably until Tuesday German time), just a bit short on time

Coronon:
I added tests now and rewrote parts of the old ones on the way - we are at 100% coverage for _share_plus_platform_interface_ now.

Sorry for the force push, I accidentally pushed my whole working state with all locally generated coverage files and pubspec edits. The now pushed version should pass :)

packages/share_plus/share_plus/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 3.1.1
+
+- Add *WithResult methods to get feedback on user action
+
```
miquelbeltran, line 4: small thing, I'd prefer if it was `3.2.0` instead. It shouldn't be a breaking change so no need for `4.0.0`

miquelbeltran:
I will need some time I look into this. 

Meanwhile, I would like that more people in the community test it and give their feedback here before I merge, as mentioned in the thread: https://github.com/fluttercommunity/plus_plugins/issues/386

packages/share_plus/share_plus/lib/share_plus.dart
```diff
@@ -79,4 +81,57 @@ class Share {
...
+  /// interacted with the share-sheet. Until the returned future is completed,
+  /// any other call to any share method that returns a result might result in
+  /// a [PlatformException].
+  ///
+  /// Because IOS and Android provide different feedback on share-sheet interaction,
+  /// a result on IOS will be more specific than on Android. While IOS can detect if
+  /// the user actually completed his selected action or aborted it midway, Android
+  /// only records if the user selected an action or outright dismissed the share-sheet.
+  ///
+  /// **Currently only implemented on IOS & Android.**
```
miquelbeltran, line 95: What would happen if a dev calls this method from desktop or web?

miquelbeltran:
Since this functionality cannot be supported on Web and Desktop, I am more inclined that this should be implemented in a separated plugin and not as part of share_plus. We had similar discussions on other PRs, and I personally think having feature parity is essential.

packages/share_plus/share_plus/lib/share_plus.dart
Coronon: It would raise an UnimplementedError

Coronon:
> Since this functionality cannot be supported on Web and Desktop, I am more inclined that this should be implemented in a separated plugin and not as part of share_plus. We had similar discussions on other PRs, and I personally think having feature parity is essential.

While I do see your point, the community seems to really want this feature and it truly seems like something that should have been included from the start.

We also don't have full support on all platforms already. Take Windows, Linux and Web for example, which all do not support **shareFiles**. Therefore I think that separating this into another method (not as an argument) which already highlights the fact that it is not supported on all platforms in its doc comment is the way to go here.

packages/share_plus/share_plus/CHANGELOG.md
Coronon: Will do as soon as I am home 👍🏻

miquelbeltran:
It's true that they don't support shareFiles on those platforms, that's because this was a legacy project from the original share plugin from Google, which was expanded to try to support other platforms. Share functionality is indeed a special one, because neither desktop nor web platforms have a native way to support those.

I am changing my mind on this. Actually, I think it doesn't make much sense that `share_plus` supports desktop or web at all, because it is misleading in the pub.dev listing, because in reality you cannot share stuff on those platforms. I am not a big fan of the `mailto:` solution used for those platforms. Maybe this is something we maintainers should reconsider.

Regarding this PR, let's continue with it. I'd like to have some feedback from other devs who can give it a go, before merging.

miquelbeltran:
Let's not do that in this PR. I will bring the idea to the rest of the maintainer team and see what they think.

Coronon:
Jup, wasn't planing to do that this PR anyway - I'll just add macOS support for good measure and then let it be 👍🏻

Coronon:
### Update! Requirement for API version 23
See PR description for more details

miquelbeltran:
FYI, I am testing the feature, I have added a "Share With Result" button to the example/lib/main.dart

miquelbeltran:
Setting minSdk to 23 sounds reasonable, but we need to do a major release instead.

miquelbeltran:
I have updated the minSdkVersion in the example, as well as tested on Android and iOS and seems to work :)

If you don't have any more changes pending, I will merge and release. Let me know @Coronon and thanks for the work!

Coronon:
Looks pretty much good to merge, I only have one little remark: This PR adds support for IOS, Android _and Mac osX_ :)

Coronon:
Should we not mention macOS in the `share_plus` changelog as well?

miquelbeltran:
> Should we not mention macOS in the `share_plus` changelog as well?

Ah true, I did not test that, I will have a go at it and update the changelog

vbuberen:
@Coronon @miquelbeltran
I am late to this discussion because of events in Ukraine, but there was no need to bump minSDK to 23 and we could still be fine with 21. I will create a PR with lowering the minSDK to have more devices supported.

Coronon:
> @Coronon @miquelbeltran
> I am late to this discussion because of events in Ukraine, but there was no need to bump minSDK to 23 and we could still be fine with 22. I will create a PR with lowering the minSDK to have more devices supported.

Yup, just like I said (because we use the API version check now).

PR #797, opened 2022-03-19
[share_plus] Fix types, lower Android minSDK

### LINKED ISSUE #758: [share_plus]  Type mismatch: inferred type is String? but String was expected
## System info

Platform the Issue occurs on:  build error 
Plugin name: share plus
Plugin version: 3.1.0
flutter sdk : 2.11.0 - master channel

<details>
<summary>Logs</summary>

```
/share_plus-3.1.0/android/src/main/kotlin/dev/fluttercommunity/plus/share/MethodCallHandler.kt: (34, 24): Type mismatch: inferred type is String? but String was expected
```

</details>

<details>
<summary>Flutter doctor output</summary>

```
[✓] Flutter (Channel master, 2.11.0-0.0.pre.633, on macOS 12.2.1 21D62
    darwin-arm, locale ko-KR)
[✓] Android toolchain - develop for Android devices (Android SDK version 31.0.0)
[✓] Xcode - develop for iOS and macOS (Xcode 13.2.1)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2020.3)
[✓] IntelliJ IDEA Community Edition (version 2021.3.2)
[✓] VS Code (version 1.64.2)
[✓] Connected device (3 available)
[✓] HTTP Host Availability
```

</details>

### LINKED ISSUE #789: share_plus Android minSdk bump drops support for a lot of devices
## System info

Platform the Issue occurs on: Android
Plugin name: share_plus
Plugin version: 4.0.1

## Steps to Reproduce

The latest `share_pus` plugin update bumped minSdk Android API from 16 to 23 (from Android 4.1 to Android 6). With Flutter itself supporting API 16 that is a large drop of supported Android devices.

My app is primarily used on Android tablets, and there is a lot of them running older Android versions. For example, older generation Amazon Fire 7 tablets (Fire OS 5 is API level 22) and some of 3..4 year old Samsung's Tab As are also running Android 5 or older.

vbuberen:
## Description

In this PR I fixed issues with type in `result.error()`, which some users reported in #758. Saw that the code wasn't formatted according to our `editorconfig` file, which was added some time ago, so reformatted the code.

I didn't like that `share_plus` now has minSDK 23 as it cuts quite a lot of users/devices. I thought we could make it much lower, but it seems that 22 is the minimum possible, because `Intent.createChooser()` with `sender` parameter appeared only in SDK 22. So I made required changes to set minSDK to 22 and some additional small fixes.

Finally, I saw that in #788 there were changes to min Flutter version, but not all platforms were updated. Fixed it as well, so `share_plus` requires Flutter 1.20 on all platforms now.

## Related Issues

Closes #758
Partially closes #789 

## Checklist

## Breaking Change

### DIFF
--- a/packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/MethodCallHandler.kt
+++ b/packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/MethodCallHandler.kt
@@ -33,7 +33,7 @@ internal class MethodCallHandler(private val share: Share, private val manager: 
           )
           result.success(null)
         } catch (e: IOException) {
-          result.error(e.message, null, null)
+          result.error("Share failed", e.message, null)
         }
       }
       "shareWithResult" -> {
@@ -61,7 +61,7 @@ internal class MethodCallHandler(private val share: Share, private val manager: 
             true,
           )
         } catch (e: IOException) {
-          result.error(e.message, null, null)
+          result.error("Share failed", e.message, null)
         }
       }
       else -> result.notImplemented()
--- a/packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/MethodCallHandler.kt
+++ b/packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/MethodCallHandler.kt
@@ -5,71 +5,74 @@ import io.flutter.plugin.common.MethodChannel
 import java.io.IOException
 
 /** Handles the method calls for the plugin.  */
-internal class MethodCallHandler(private val share: Share, private val manager: ShareSuccessManager) : MethodChannel.MethodCallHandler {
+internal class MethodCallHandler(
+    private val share: Share,
+    private val manager: ShareSuccessManager
+) : MethodChannel.MethodCallHandler {
 
-  override fun onMethodCall(call: MethodCall, result: MethodChannel.Result) {
-    when (call.method) {
-      "share" -> {
-        expectMapArguments(call)
-        // Android does not support showing the share sheet at a particular point on screen.
-        share.share(
-          call.argument<Any>("text") as String,
-          call.argument<Any>("subject") as String?,
-          false,
-        )
-        result.success(null)
-      }
-      "shareFiles" -> {
-        expectMapArguments(call)
+    override fun onMethodCall(call: MethodCall, result: MethodChannel.Result) {
+        when (call.method) {
+            "share" -> {
+                expectMapArguments(call)
+                // Android does not support showing the share sheet at a particular point on screen.
+                share.share(
+                    call.argument<Any>("text") as String,
+                    call.argument<Any>("subject") as String?,
+                    false,
+                )
+                result.success(null)
+            }
+            "shareFiles" -> {
+                expectMapArguments(call)
 
-        // Android does not support showing the share sheet at a particular point on screen.
-        try {
-          share.shareFiles(
-            call.argument<List<String>>("paths")!!,
-            call.argument<List<String>?>("mimeTypes"),
-            call.argument<String?>("text"),
-            call.argument<String?>("subject"),
-            false,
-          )
-          result.success(null)
-        } catch (e: IOException) {
-          result.error("Share failed", e.message, null)
+                // Android does not support showing the share sheet at a particular point on screen.
+                try {
+                    share.shareFiles(
+                        call.argument<List<String>>("paths")!!,
+                        call.argument<List<String>?>("mimeTypes"),
+                        call.argument<String?>("text"),
+                        call.argument<String?>("subject"),
+                        false,
+                    )
+                    result.success(null)
+                } catch (e: IOException) {
+                    result.error("Share failed", e.message, null)
+                }
+            }
+            "shareWithResult" -> {
+                expectMapArguments(call)
+                if (!manager.setCallback(result)) return
+
+                // Android does not support showing the share sheet at a particular point on screen.
+                share.share(
+                    call.argument<Any>("text") as String,
+                    call.argument<Any>("subject") as String?,
+                    true,
+                )
+            }
+            "shareFilesWithResult" -> {
+                expectMapArguments(call)
+                if (!manager.setCallback(result)) return
+
+                // Android does not support showing the share sheet at a particular point on screen.
+                try {
+                    share.shareFiles(
+                        call.argument<List<String>>("paths")!!,
+                        call.argument<List<String>?>("mimeTypes"),
+                        call.argument<String?>("text"),
+                        call.argument<String?>("subject"),
+                        true,
+                    )
+                } catch (e: IOException) {
+                    result.error("Share failed", e.message, null)
+                }
+            }
+            else -> result.notImplemented()
         }
-      }
-      "shareWithResult" -> {
-        expectMapArguments(call)
-        if(!manager.setCallback(result)) return
-
-        // Android does not support showing the share sheet at a particular point on screen.
-        share.share(
-          call.argument<Any>("text") as String,
-          call.argument<Any>("subject") as String?,
-          true,
-        )
-      }
-      "shareFilesWithResult" -> {
-        expectMapArguments(call)
-        if(!manager.setCallback(result)) return
-
-        // Android does not support showing the share sheet at a particular point on screen.
-        try {
-          share.shareFiles(
-            call.argument<List<String>>("paths")!!,
-            call.argument<List<String>?>("mimeTypes"),
-            call.argument<String?>("text"),
-            call.argument<String?>("subject"),
-            true,
-          )
-        } catch (e: IOException) {
-          result.error("Share failed", e.message, null)
-        }
-      }
-      else -> result.notImplemented()
     }
-  }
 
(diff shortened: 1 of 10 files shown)


packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/MethodCallHandler.kt
```diff
@@ -5,71 +5,74 @@ import io.flutter.plugin.common.MethodChannel
...
+                    share.shareFiles(
+                        call.argument<List<String>>("paths")!!,
+                        call.argument<List<String>?>("mimeTypes"),
+                        call.argument<String?>("text"),
+                        call.argument<String?>("subject"),
+                        false,
+                    )
+                    result.success(null)
+                } catch (e: IOException) {
+                    result.error("Share failed", e.message, null)
```
Mosc, line 39: I was considering submitting a similar fix, but keep in mind that this may be a breaking change for some users that rely on the returned error code or error message. At the very least it should coincide with more than a simple PATCH version bump.

packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/ShareSuccessManager.kt
```diff
@@ -36,13 +33,17 @@ internal class ShareSuccessManager(private val context: Context): BroadcastRecei
      * the componentname of the chosen option or an empty string on dismissal.
      */
     fun setCallback(callback: MethodChannel.Result): Boolean {
-        if (calledBack.compareAndSet(true, false)) {
+        return if (calledBack.compareAndSet(true, false)) {
```
Coronon, line 36: While this does do the job, don't you agree that this makes the code less readable? We don't really get anything from this change, and it's not consistent (see lines 72-77).

Mosc:
Not sure how these cases are generally handled, but I would prefer the `minSdkVersion` to go all the way back to 16 and simply return an appropriate `result.error(...)` for `shareWithResult` and `shareFilesWithResult` on SDK versions lower than 22. That would allow users to just continue to use `share` and `shareFiles` as they did before v4.0.0.

Coronon:
> Not sure how these cases are generally handled, but I would prefer the `minSdkVersion` to go all the way back to 16 and simply return an appropriate `result.error(...)` for `shareWithResult` and `shareFilesWithResult` on SDK versions lower than 22. That would allow users to just continue to use `share` and `shareFiles` as they did before v4.0.0.

As said - I am pretty new to Android development. If that compiles and works as expected, it is probably the best option we have. The only risk I see there is people unintentionally using `shareWithResult` with a `minSdkVersion` lower than 16 and not noticing because their development environment is API level 22+. This could lead to users effected by unforeseen errors - can we somehow statically check this at compile time?

packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/ShareSuccessManager.kt
vbuberen: No, I don't think so. It is a normal Kotlin syntax, which is used for a long time in Android development. If you would use Android Studio you would even see a lint suggestion to change this way. 
Lines you suggested to consider are a different case, which can't be turned into construction like here.

packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/MethodCallHandler.kt
vbuberen: I doubt that somebody relied on this as otherwise this person would open an issue that `e.message` was set on a place of `errorCode` parameter.

packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/ShareSuccessManager.kt
Coronon: Well, you could just do this on lines 72-77:
```kotlin
return if (requestCode == ACTIVITY_CODE) {
    returnResult("")
    true
} else {
    false
}
```

packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/MethodCallHandler.kt
Mosc: Right, I would also doubt it, but no reason not to err on the side of caution in terms of versioning.

Mosc:
> > Not sure how these cases are generally handled, but I would prefer the `minSdkVersion` to go all the way back to 16 and simply return an appropriate `result.error(...)` for `shareWithResult` and `shareFilesWithResult` on SDK versions lower than 22. That would allow users to just continue to use `share` and `shareFiles` as they did before v4.0.0.
> 
> As said - I am pretty new to Android development. If that compiles and works as expected, it is probably the best option we have. The only risk I see there is people unintentionally using `shareWithResult` with a `minSdkVersion` lower than 16 and not noticing because their development environment is API level 22+. This could lead to users effected by unforeseen errors - can we somehow statically check this at compile time? And API version 22 is from 2015, so is this really necessary?

I understand your concerns. A runtime error is certainly not ideal, although to be fair, `shareWithResult` and `shareFilesWithResult` _already_ fail on runtime on web, Windows and Linux despite the package claiming those platforms are supported! We rely on the user reading the documentation for those specific methods.

How about falling back to the regular non-result based share operations and just returning `ShareResultStatus.unavailable` for Android SDKs under 22?

packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/ShareSuccessManager.kt
vbuberen: > Well, you could just do this on lines 72-77 or am I wrong?

You are right, I overlooked that case.

> guess I just don't like how it looks 🤷🏻‍♂️

Just a case of tastes, I guess, because for me as for the person, who was working with Kotlin for a long time it seems fine. Will update code for consistency, though.

vbuberen:
> How about falling back to the regular non-result based share operations and just returning `ShareResultStatus.unavailable` for Android SDKs under 22?

I agree that we should rework the solution with sharing with result. Not sure I can jump right into it (because an Android dev for quite some time and might be the right person to take care of it) as I am quite occupied with some activities here in Ukraine due to infamous events, so don't have much time. But let's see how it goes.

packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/MethodCallHandler.kt
```diff
@@ -5,71 +5,74 @@ import io.flutter.plugin.common.MethodChannel
...
+                }
+            }
+            "shareWithResult" -> {
+                expectMapArguments(call)
```
ekuleshov, line 43: could these `*withResults(..)` methods extended with check like

```
if(Build.VERSION.SDK_INT > Build.VERSION_CODES.LOLLIPOP_MR1)
  result.notImplemented()
else ...
```

Then it won't be necessary to bump the `minSdkVersion` version for Android platform and it could be kept at the level supported in `3.x` version of plugin.

Coronon:
> How about falling back to the regular non-result based share operations and just returning `ShareResultStatus.unavailable` for Android SDKs under 22?

That sounds like a good idea, as we already have that status and would not need to break any existing APIs as of version 4.0.1

I might even try to implement that tomorrow

ekuleshov:
> How about falling back to the regular non-result based share operations and just returning `ShareResultStatus.unavailable` for Android SDKs under 22?

Thank you for looking into this.

Now, if some platform would return `ShareResultStatus.unavailable`, perhaps it would be more consistent to do the same on web, linux and windows too instead of tossing not supported runtime exception on those?

Coronon:
I will fork this fork to base my PR on the better formatted code

vbuberen:
> I will fork this fork to base my PR on the better formatted code

I wouldn't bother yourself with that to depend on this PR, since I don't know for sure as to when it gets a review.

If it is just about formatting - auto formatting should work fine for you no matter which IDE you use as the project has an editorconfig file in its root, so code format should be forced according to riles in that file.

Coronon:
> If it is just about formatting - auto formatting should work fine for you no matter which IDE you use as the project has an editorconfig file in its root, so code format should be forced according to riles in that file.

I only use melos format, which seems to not yield the same result as your formatting, but ok, will depend on the main branch then

Coronon:
Just pushed my changes, please try this example project with your low API version devices/emulators: https://github.com/Coronon/share_plus_result_example

miquelbeltran (approved):
Thank you @vbuberen !

packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/MethodCallHandler.kt
miquelbeltran: I think so. If anyone can submit a separated PR with a that change, that would be awesome!

miquelbeltran:
and published

PR #832, opened 2022-04-11
[all] Fix formatting issues

miquelbeltran:
## Description

Runs `melos run format` and fixes multiple formatting issues

The CI has been adapted to accept the formatting output of `melos run format` by removing the `--style=Google` parameter in the clang-format call for Objective-C classes.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,6 +7,10 @@ insert_final_newline = true
 trim_trailing_whitespace = true
 charset = utf-8
 
+[*.cpp]
+indent_style = space
+indent_size = 2
+
 [*.dart]
 indent_style = space
 indent_size = 2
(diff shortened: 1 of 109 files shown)


miquelbeltran:
Objective-C formatting seems to give different results when running `melos format` on my machine and on CI, I will take a look at that

miquelbeltran:
The problem with Objective-C, is that `melos run format` formats it in the default `clang-format` style, but CI requires the "Google" style. This will cause a problem in the future when we run `melos run format`.

To simplify thing, I am changing CI to format using the default formatting like the `melos run format`.

vbuberen (commented):
As I understand the main change is in `all_plugins.yaml` and other things are just reformatting.

Suggest to also modify `editorconfig` file to include rules to Objective-C.

miquelbeltran:
Correct, the `all_plugins.yaml` is the only relevant change.

I will look at adding a configuration for `.m` on `editorconfig`. Would it be necessary for this PR?

vbuberen:
> Would it be necessary for this PR?

I would do it, considering that it is directly related as well and pretty trivial change.

PR #911, opened 2022-06-15
update version of shareplus to the latest version

Faezehyas:
update the version of shareplus to the latest version in the overview.mdx

### DIFF
--- a/docs/share_plus/overview.mdx
+++ b/docs/share_plus/overview.mdx
@@ -20,7 +20,7 @@ on iOS.
 dependencies:
   flutter:
     sdk: flutter
-  share_plus: "^{{ plugins.share_plus }}"
+  share_plus: ^4.0.8
 ```
 
 #### 2. Download dependency


docs/share_plus/overview.mdx
```diff
@@ -20,7 +20,7 @@ on iOS.
 dependencies:
   flutter:
     sdk: flutter
-  share_plus: "^{{ plugins.share_plus }}"
```
miquelbeltran, line 23: I am not sure this should be replaced by a hardcoded value

Faezehyas: Then how do you think it is better to change it?
I use this package in my codes and saw the documentation version is so old.

miquelbeltran: I believe versions are updated when the website is redeployed, as it is generated automatically from pub.dev.

miquelbeltran:
@mhadaily Do you know how to trigger a re-deploy of the website, so versions are up to date?

mhadaily:
If we can just make a simple change in the website folder it will deploy according to this ´https://github.com/fluttercommunity/plus_plugins/blob/main/.github/workflows/build_deploy_website.yaml´

otherwise, we can run it locally by running ´npm run deploy´ in believe,


I can take care of that, should I deploy now?

miquelbeltran:
Yes, I guess we need to do re-deploys once in a while, if not the versions listed on the site will be outdated

mhadaily:
Ok, let's actually change GitHub action to deploy when we have a new tag available, I'll send a PR today

PR #918, opened 2022-06-19
[device_info_plus] Remove AndroidId, update Windows dependencies

### LINKED ISSUE #824: Could we exclude Settings.Secure.ANDROID_ID
I recently submitted an app to Google Play Store and it seems that from June onwards you have to declare if you use tracking id's.

The error message from Google Play is not easy to understand, I was confused for quite some time before I found the usage of ANDROID_ID[ inside device_info_plus](https://github.com/fluttercommunity/plus_plugins/blob/main/packages/device_info_plus/device_info_plus/android/src/main/java/dev/fluttercommunity/plus/device_info/MethodCallHandlerImpl.java#L109).

Since our app does not need ANDROID_ID, I prefer not to declare that our app uses tracking identifiers since I think this will show as a badge in the Google Play store which is unnecessary if the app does not need it.

For now I have just forked device_info_plus and removed it and Google is happy but given that this plugin is widely used, perhaps it is worth looking for a solution for this to not force apps not using tracking identifiers to declare that they do?

vbuberen:
## Description

According to report in #824 Google Play's new policies discourage usage of unique ids, by requiring to explicitly declare usage of such ids in Data Safety section. I recently had similar experience with `flutter_segment` package, which used segment_analytics Android library, which was accessing AndroidId as well: https://github.com/la-haus/flutter-segment/issues/30
I believe we should exclude AndroidId from set of data returned by `device_info_plus`, so people, who use the package don't have to declare collection of unique IDs in Data Safety section on their Google Play Store listing. Bumped the major version, since removal is a breaking change from my point of view. Those, who would still need to use AndroidId for some reason could use older version of `device_info_plus`.

Additionally, needed to update some dependencies for Windows plugin as building `example` app was failing for me. Due to changes in requirements on `ffi` and `win32` had to bump min Dart version to `2.17` in Windows plugin.

## Related Issues

Closes #824

## Checklist

## Breaking Change

- [x] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).

### DIFF
--- a/packages/device_info_plus/device_info_plus/CHANGELOG.md
+++ b/packages/device_info_plus/device_info_plus/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 4.0.0
 
-- Remove `AndroidId` getter to avoid Google Play policies violations
+- **Breaking change** Remove `AndroidId` getter to avoid Google Play policies violations
 - Update flutter_lints to 2.0.1
 - Remove explicit `test` dependency to use `flutter_test` from Flutter SDK
 
--- a/packages/device_info_plus/device_info_plus/CHANGELOG.md
+++ b/packages/device_info_plus/device_info_plus/CHANGELOG.md
@@ -1,6 +1,12 @@
+## 4.0.0
+
+- Remove `AndroidId` getter to avoid Google Play policies violations
+- Update flutter_lints to 2.0.1
+- Remove explicit `test` dependency to use `flutter_test` from Flutter SDK
+
 ## 3.2.4
 
-- Update the description of getAndroidId method  
+- Update the description of getAndroidId method
 
 ## 3.2.3
 
--- a/packages/device_info_plus/device_info_plus/android/src/main/java/dev/fluttercommunity/plus/device_info/MethodCallHandlerImpl.java
+++ b/packages/device_info_plus/device_info_plus/android/src/main/java/dev/fluttercommunity/plus/device_info/MethodCallHandlerImpl.java
@@ -4,12 +4,10 @@
 
 package dev.fluttercommunity.plus.device_info;
 
-import android.annotation.SuppressLint;
 import android.content.ContentResolver;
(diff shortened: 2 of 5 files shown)


miquelbeltran:
Yep, good catch, makes sense to me!

miquelbeltran (approved):

miquelbeltran:
new versions published now

bhanuka96:
It's better to mark depreciated labe

vbuberen:
> What ID now can use as a unique ID?

You should refer to the official documentation for that https://developer.android.com/training/articles/user-data-ids

> It's better to mark depreciated labe

Because new policy is going to be enforced on 9th of August already we had no time to do a proper deprecation cycle. However, nobody stops you from using pre 4.0.0 version of the plugin if you want to still rely on AndroidId as there are no new features of critical fixes so far.

feinstein:
@vbuberen since the other issue got locked I will reply it in here.

I wasn't demanding anyting, this is a community effort, so of course no one is obligated to nothing. My point was to make an open discussion so we can reach an agreement. The decision to extract `AndroidId` impacted several apps, many financial apps like mine, and having to import another package just to get this variable seems excessive to me.

I made a suggestion on how we can circumvent this, but the reason why I didn't send any of it as code is because I am forbidden from sending any code by my employment contract. All the code that I make is the property of my employer. What I am trying to do right now is to get an exception, or to find someone else that can open a PR with my suggestion.

feinstein:
@vbuberen I did more reseach into the source code and at first I thought this was a flag under the Manifest, but now I see it's just a method call. I thought the Play Store was filtering apps that declared access to the ANDROID_ID in their Manifests, but now I see that's not the case.

I don't remember this being mentioned by the maintainers, but this information can justify a new package, as the Play Store could search the app's source code, looking for calls to retrieve the `ANDROID_ID`.

Issue #949, opened 2022-07-07
[Request]: make toMap() json compatible


iapicca:
### Plugin

device_info_plus

### Use case

renaming `toMap()` as `toJson` would improve compatibility with `dart:convert` 
see referece at https://api.dart.dev/be/175791/dart-convert/jsonEncode.html

> If toEncodable is omitted, it defaults to a function that returns the result of calling .toJson() on the unencodable object.

### Proposal

rename BaseDeviceInfo method toMap() as toJson() and to all overrides across the package

iapicca:
after discussion in the [PR](https://github.com/fluttercommunity/plus_plugins/pull/950)
 `toJson` will not be implemented, but  `toMap` will be rendered json parsing compatible

Yczar:
Fixed this issue here #1132

iapicca:
@mhadaily I'm not very happy of seeing my PR
on which I spent several hours and being on hold and without / without an answer for 2 months
being superseded for "hacktoberfest" sake

iapicca:
@miquelbeltran could you please take a look at my [question](https://github.com/fluttercommunity/plus_plugins/pull/950#issuecomment-1193315008)
I see that there is a new PR, but I'm still waiting for a feedback

iapicca:
> @iapicca sorry, I forgot that we had your PR on this, I apologize for not having followed up.
> [...] 

no problem :)

> As I though originally, I am not in favor of adding extra functionality to the plugins that should be external to the plugin, like JSON support.

what about fixing the crash that occurs using the current implementation of "toMap" using `jsonEncode()`
(caused by mapping an enum not as string)

in the question I asked I proposed 2 possible approaches

> @miquelbeltran I think we have to option for the web plugin implementation:
> 
> a) remove `browserName` from both `toMap` and `fromMap`
> 
> b) keep `browserName` in both `toMap` and `fromMap` ensuring consistency with `userAgent` data
> 
> `b` is the current implementation and requires the[ committed changes](https://github.com/fluttercommunity/plus_plugins/pull/950/files#diff-99f10196cb030bdc5fb387fc806693089a9fb6ce15e46961c7034fab0bb1fa83), I'm not sure what changes you wish to be limited

miquelbeltran:
Sorry @iapicca for not following up.

[...]

PR #950, opened 2022-07-07
[device_info_plus] make toMap() json compatible

iapicca:
## Description

> rename `toMap()` as `toJson()` to improve compatibility with `dart:convert`
> see [jsonEncode]( https://api.dart.dev/be/175791/dart-convert/jsonEncode.html)
> 
> > If toEncodable is omitted, it defaults to a function that returns the result of calling .toJson() on the unencodable object.

updated: makes `toMap` compatible with json parsing

## Related Issues

fix https://github.com/fluttercommunity/plus_plugins/issues/949

## Checklist

## Breaking Change

- [] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [x ] No, this is *not* a breaking change.

TODO:

after publishing, 
bump `device_info_plus_platform_interface` to `^2.3.1` in  `device_info_plus`'s pubspec


miquelbeltran:
Thank you! I am positive about this change. The only issue is that this would be a major version bump, so the platform plugin should be increased to 3.0.0, and all its dependencies as well. The change should be well documented in the respective Chanelogs.

Any thoughts? @vbuberen

iapicca:
@miquelbeltran
thanks for looking up the PR

@vbuberen 
besides the originally stated reason, allowing `jsonEncode` to be cast seamlessly
during the implementation I also noticed that in some platform running
`jsonEncode(info.toMap())` is broken because is the `Map` doesn't always have json compatible types (eg: `enum`)

miquelbeltran:
> during the implementation I also noticed that in some platform `toMap` is "broken" more specifically `jsonEncode(info.toMap())` throws because toMap doesn't always return json compatible types (eg: `enum`) this PR address this problem as well

mmhhh... I am more inclined now to make toJson() a new method rather than modifying the toMap(), that way this is not going to be a breaking change.

What do you think about that @iapicca ?

iapicca:
> mmhhh... I am more inclined now to make toJson() a new method rather than modifying the toMap(), that way this is not going to be a breaking change.
> 
> What do you think about that @iapicca ?

@miquelbeltran 
no problem, I re-added the toMap method straight from the current upstream `main`
in most cases I can just `toJson() => toMap()`, but the web platform interface was messed up
and there is a different behavior for both `toJson` and `fromMap`
please take a look

iapicca:
I see that a [package_info_plus test](https://github.com/fluttercommunity/plus_plugins/blob/main/.github/workflows/package_info_plus.yaml) fails because dependencies conflicts with [flutter stable](https://github.com/fluttercommunity/plus_plugins/blob/main/.github/workflows/scripts/install-flutter.sh) (no [flutter-action](https://github.com/subosito/flutter-action)?) 
seems outside the scope of this PR, but let me know if I need to take any action

```console
Because package_info_plus_macos requires the Flutter SDK, version solving failed.
Running "flutter pub get" in package_info_plus_macos...         
Flutter users should run `flutter pub get` instead of `dart pub get`.
---- Log transcript ----
FINE: Pub 2.17.6
MSG : Resolving dependencies...
SLVR: fact: package_info_plus_macos is 1.3.0
SLVR: derived: package_info_plus_macos
SLVR: fact: package_info_plus_macos requires the Flutter SDK
SLVR: conflict: package_info_plus_macos requires the Flutter SDK
SLVR: Version solving took 0:00:00.080[109](https://github.com/fluttercommunity/plus_plugins/runs/7388981363?check_suite_focus=true#step:5:110) seconds.
    | Tried 1 solutions.
FINE: Resolving dependencies finished (0.1s).
ERR : Because package_info_plus_macos requires the Flutter SDK, version solving failed.
    | 
    | Flutter users should run `flutter pub get` instead of `dart pub get`.
FINE: Exception type: SolveFailure
FINE: package:pub/src/solver/version_solver.dart 311:5               VersionSolver._resolveConflict
    | package:pub/src/solver/version_solver.dart 132:27              VersionSolver._propagate
    | package:pub/src/solver/version_solver.dart 96:11               VersionSolver.solve.<fn>
    | ===== asynchronous gap ===========================
    | dart:async                                                     Future.catchError
    | package:pub/src/utils.dart 109:52                              captureErrors.wrappedCallback
    | package:stack_trace                                            Chain.capture
    | package:pub/src/utils.dart [122](https://github.com/fluttercommunity/plus_plugins/runs/7388981363?check_suite_focus=true#step:5:123):11                              captureErrors
    | package:pub/src/command.dart 183:13                            PubCommand.run
    | package:args/command_runner.dart 209:27                        CommandRunner.runCommand
    | package:pub/src/command_runner.dart 174:24                     PubCommandRunner.runCommand
    | package:pub/src/command_runner.dart 159:20                     PubCommandRunner.run
    | package:dartdev/dartdev.dart 45:56                             runDartdev
    | /b/s/w/ir/cache/builder/sdk/pkg/dartdev/bin/dartdev.dart 11:9  main
---- End log transcript ----
pub get failed (1; ---- End log transcript ----)
BootstrapException: Failed to install.: package_info_plus_macos at /home/runner/work/plus_plugins/plus_plugins/packages/package_info_plus/package_info_plus_macos.
Error: Process completed with exit code 1.
```

miquelbeltran:
I am sorry, I am not sure anymore if we need this. If a user wants to serialize the contents of a `DeviceInfo` object, they should do it on their own app data model. We shouldn't be maintaining a `toJson` implementation on our side, even if convenient, it is not our job. Like, I am not even sure why are we providing a toMap method even, I guess that's the legacy from when we got the plugins from Google.

iapicca:
> I am sorry, I am not sure anymore if we need this. If a user wants to serialize the contents of a `DeviceInfo` object, they should do it on their own app data model. We shouldn't be maintaining a `toJson` implementation on our side, even if convenient, it is not our job. Like, I am not even sure why are we providing a toMap method even, I guess that's the legacy from when we got the plugins from Google.

@miquelbeltran 
can I at least [fix the web implementation](https://github.com/fluttercommunity/plus_plugins/pull/950#issuecomment-1185270273)?
 It's literally the only case  where it cannot be use as `jsonEncode(info.toMap())`

iapicca:
I did merge properly [before](https://github.com/fluttercommunity/plus_plugins/pull/950#issuecomment-1187271679
), now it's up to date

> > [...] can I at least [fix the web implementation](https://github.com/fluttercommunity/plus_plugins/pull/950#issuecomment-1185270273)? It's literally the only case where it cannot be use as `jsonEncode(info.toMap())`
> 
> Yes, I agree with that.
> 
> The Readme.md says:
> 
> > You then use it's toMap method to serialize all known properties to a Map. Your backend service should be prepared to handle new properties, which can be added to this plugin in the future.
> 
> If the serialization using toMap doesn't work because it uses enums, then it must be fixed!

@miquelbeltran 
that's good to hear!
So I removed the `toJson` and ported the fix for the web implementation
I think this way is more than usable and if someone (like me) really wants `toJson`
can just do 

```dart
extension BaseDeviceInfoToJsonX on BaseDeviceInfo {
  Map<String, Object?> Function() get toJson => toMap;
}
```

miquelbeltran:
Thanks! can you cleanup the PR, and limit the changes to the web plugin implementation?

packages/device_info_plus/device_info_plus_platform_interface/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.3.1+0
+
+- breaking change: toMap renamed as toJson for dart:convert compatibility
```
miquelbeltran, line 3: message outdated, the toMap should not be renamed to toJson

miquelbeltran (changes requested):
All changes not related to the fix should be reverted

packages/device_info_plus/device_info_plus_platform_interface/CHANGELOG.md
iapicca: I will update it in a couple of days, sorry about that (I'm a bit stuffed)

iapicca: fixed in [a880118](https://github.com/fluttercommunity/plus_plugins/pull/950/commits/a8801184ce028c19f23c8e54443391a7723b0c07)

iapicca:
> Thanks! can you cleanup the PR, and limit the changes to the web plugin implementation?

> All changes not related to the fix should be reverted

@miquelbeltran 
I think we have to option for the web plugin implementation:

a)  remove `browserName` from both `toMap` and `fromMap`

b) keep `browserName` in both `toMap` and `fromMap` ensuring consistency with `userAgent` data

`b` is the current implementation and requires the[ committed changes](https://github.com/fluttercommunity/plus_plugins/pull/950/files#diff-99f10196cb030bdc5fb387fc806693089a9fb6ce15e46961c7034fab0bb1fa83),
I'm not sure what changes you wish to be limited

iapicca:
as I didn't receive any response I'm closing this PR

PR #975, opened 2022-07-24
test(device_info_plus): Integration test

Chever-John:
## Description

This PR is an integration test written for device_info_plus. I have completed the Android & iOS parts o far. If everything works, it will be finished today at 18:00 UTC.

## Related Issues

This is a GSoC task, so there are no associated issues for now.

## Checklist

- [ ] I updated the version in `pubspec.yaml` and `CHANGELOG.md`.
- [ ] I updated/added relevant documentation (doc comments with `///`).

## Breaking Change


packages/device_info_plus/device_info_plus/example/integration_test/device_info_plus_test.dart
```diff
@@ -65,4 +66,121 @@ void main() {
...
+      expect(iosInfo.identifierForVendor, isNotNull);
+      expect(iosInfo.isPhysicalDevice, isNotNull);
+
+      expect(iosInfo.utsname.sysname, equals('Darwin'));
+      expect(iosInfo.utsname.nodename, isNotNull);
+      expect(iosInfo.utsname.release, isNotNull);
+      expect(iosInfo.utsname.version, isNotNull);
+      expect(iosInfo.utsname.machine, isNotNull);
+    }
+  });
```
miquelbeltran, line 87: At the moment, there's no iOS integration testing setup, only Android is set up, so we cannot have this test (as it will do nothing)

packages/device_info_plus/device_info_plus/example/integration_test/device_info_plus_test.dart
```diff
@@ -65,4 +66,121 @@ void main() {
...
+
+  testWidgets('android', (WidgetTester tester) async {
```
miquelbeltran, line 89: Can you change the test name to something else? For example: "check android info values are set".

packages/device_info_plus/device_info_plus/example/integration_test/device_info_plus_test.dart
```diff
@@ -65,4 +66,121 @@ void main() {
...
+    if (Platform.isAndroid) {
+      expect(androidInfo.version.baseOS, isNotNull);
+      expect(androidInfo.version.codename, isNotNull);
+      expect(androidInfo.version.incremental, isNotNull);
+      expect(androidInfo.version.previewSdkInt, isNotNull);
+      expect(androidInfo.version.release, isNotNull);
+      expect(isInteger(androidInfo.version.sdkInt), true);
+      expect(androidInfo.version.securityPatch, isNotNull);
```
miquelbeltran, line 97: Instead of checking for `notNull`, which is not bad, but could be better, can you check against actual values?

for example: `expect(androidInfo.version.sdkInt, 30)`

We can do that because we know that the Emulator has version 30

packages/device_info_plus/device_info_plus/example/integration_test/device_info_plus_test.dart
```diff
@@ -65,4 +66,121 @@ void main() {
...
+      expect(macosInfo.arch, isNotNull);
+      expect(macosInfo.model, isNotNull);
+      expect(macosInfo.kernelVersion, isNotNull);
+      expect(macosInfo.osRelease, isNotNull);
+      expect(macosInfo.activeCPUs, isNotNull);
+      expect(macosInfo.memorySize, isNotNull);
+      expect(macosInfo.cpuFrequency, isNotNull);
+      expect(macosInfo.systemGUID, isNotNull);
+    }
+  });
```
miquelbeltran, line 180: same as with the iOS, we don't have CI integration that runs tests on these platforms, so the test will never run. I'd suggest to remove them at the moment.

packages/device_info_plus/device_info_plus/example/integration_test/device_info_plus_test.dart
```diff
@@ -65,4 +66,121 @@ void main() {
...
 }
+
+bool isInteger(num value) =>
+    value is int || value == value.roundToDouble() || value != null;
+
+bool isString(String value) => value != null;
```
miquelbeltran, line 186: These shouldn't be necessary. The `isString` method is strange, as `String value` can never be `null`. Same with the `isInteger` method.

packages/device_info_plus/device_info_plus/example/integration_test/device_info_plus_test.dart
```diff
@@ -65,4 +66,121 @@ void main() {
...
+      expect(iosInfo.utsname.sysname, equals('Darwin'));
+      expect(iosInfo.utsname.nodename, isNotNull);
+      expect(iosInfo.utsname.release, isNotNull);
+      expect(iosInfo.utsname.version, isNotNull);
+      expect(iosInfo.utsname.machine, isNotNull);
+    }
+  });
+
+  testWidgets('android', (WidgetTester tester) async {
+    if (Platform.isAndroid) {
```
miquelbeltran, line 90: Instead of having an if condition here, you can set the `skip:` parameter of the `testWidgets` to `skip: !Platform.isAndroid` so the test will only run when running on an Android device. I think this is worth trying.

miquelbeltran (changes requested):
The Android test is a good idea,  but I think it can be improved.

The other tests should be removed as there's currently no integration with those platforms, so the tests would always pass, which is not ideal.

you can also use the `skip:` parameter in the `testWidgets` method to skip a test conditionally.

packages/device_info_plus/device_info_plus/example/integration_test/device_info_plus_test.dart
Chever-John: Got~

Chever-John:
@miquelbeltran 
I think you're right, so I'm going to focus on optimizing the integration tests for Android. Doing everything I can to think of scenarios and test them /:)

packages/device_info_plus/device_info_plus/example/integration_test/device_info_plus_test.dart
Chever-John: I found this [usage](https://github.com/dart-lang/test/issues/545#:~:text=failure%27)%3B%0A%20%20%20%20%7D-,%2C%20skip%3A%20false,-)%3B%0A%20%20%7D%2C%20skip%3A) through google, not sure if it is what you are describing.
If so, I'm curious what the rationale for using this method is? Because I'm using Android studio's automation tools to run a command similar to the following.
`/Users/cheverjohn/SE/envs/flutter/flutter3/bin/flutter --no-color test --machine --start-paused --name "Can get non\-null device model(\s*\( variant: . *\))? $" -d 24c0f2776eb8a90283ef5f22dfad19984fb7615e`
integration_test/device_info_plus_test.dart

Seems simpler?

Chever-John: Oh, I see. 
Maybe This is an unnecessary judgment statement.

miquelbeltran: I am not sure if we are talking about the same.

The `skip` parameter from the `testWidgets` allows you to disable a test from running given a certain condition. The problem is that if you run this test on a non-Android machine it will always pass, even if it did nothing, giving a false positive. It would be better if instead the test was skipped completely.

packages/device_info_plus/device_info_plus/example/integration_test/device_info_plus_test.dart
Chever-John: Here's the thing, this is just an attempt on my part. Trying to go and add some logic to my integration tests.

miquelbeltran (approved):

packages/device_info_plus/device_info_plus/example/integration_test/device_info_plus_test.dart
```diff
@@ -65,4 +65,39 @@ void main() {
...
+    expect(androidInfo.supported32BitAbis, isNotNull);
+    expect(androidInfo.supported64BitAbis, isNotNull);
+    expect(androidInfo.supportedAbis, isNotNull);
+
+    expect(androidInfo.tags, isNotNull);
+    expect(androidInfo.type, isNotNull);
+    expect(androidInfo.isPhysicalDevice, isNotNull);
+    expect(androidInfo.androidId, isNull);
+    expect(androidInfo.systemFeatures, isNotNull);
+  }, skip: !Platform.isAndroid);
```
miquelbeltran, line 102: @Chever-John this is what I mean with the `skip` parameter

Issue #939, opened 2022-06-29
[Bug]:  Can't get androidID


kaichun831:
### Platform

Android 11

### Plugin

device_info_plus

### Version

4.0.0 

### Flutter SDK

2.10.3

### Steps to reproduce

null

### Code Sample

```dart
var myDeviceInfo = DeviceInfoPlugin();     
var device = readAndroidBuildData(await myDeviceInfo.androidInfo);
mobileId = device['androidId'].toString();  // null
```

### Logs

```shell
[...]

PR #1021, opened 2022-08-14
Redo changes in device_info_plus_platform_interface 2.6.0

miquelbeltran:
## Description

- Revert "Revert 4.1.0 device_info_plus changes (#1020)"
- Version bumps

## Related Issues

https://github.com/fluttercommunity/plus_plugins/issues/939#issuecomment-1213044548

### DIFF
--- a/packages/device_info_plus/device_info_plus/CHANGELOG.md
+++ b/packages/device_info_plus/device_info_plus/CHANGELOG.md
@@ -1,7 +1,3 @@
-## 4.1.1
-
-- Revert changes in 4.1.0
-
 ## 4.1.0
 
 - Remove `androidId` (that already got removed from the method channel in 4.0.0, thus always returned null)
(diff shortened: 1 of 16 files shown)


miquelbeltran:
Maybe we need to update all the dependencies of the macos, windows, linux and web federated plugins.

nohli (approved):
LGTM

miquelbeltran:
> Maybe we need to update all the dependencies of the macos, windows, linux and web federated plugins.

so here's the catch:

- device_info_plus 4.1.2 depends on platform_inferface 3.0.0
- device_info_plus 4.1.2 depends on macos 2.2.3, that depends on platform_interface 2.2.0
- same issue for windows, linux and web

Would that be a problem? Should be bump all the versions of the other plugins to be on the safe side?

miquelbeltran:
I don't trust having different dependencies, so I bumped all versions.

nohli:
Thanks! This makes sure we have a stable release of everything.

nohli (approved):

PR #1027, opened 2022-08-17
Update to make android and IOS return the same expected wifi name.

EgHubs:
without these changes users has to check for platform before calling getWifiName which is not the best situation.


miquelbeltran:
Unfortunately that will disturb with SSID that contains double quotes in the name.

miquelbeltran:
You are right, however, in my opinion we shouldn't be overriding what the OS provides, as there's a reason why Android sometimes returns SSIDs with quotes and sometimes doesn't:

```
    /**
     * Returns the service set identifier (SSID) of the current 802.11 network.
     * If the SSID can be decoded as UTF-8, it will be returned surrounded by double
     * quotation marks. Otherwise, it is returned as a string of hex digits. The
     * SSID may be {@code null} if there is no network currently connected.
     * @return the SSID
     */
```

IMO, we cannot decide for the developer, and we should let them choose how they want to handle this.

We would be happy to accept a PR that adds more documentation to the method, otherwise.

Issue #730, opened 2022-02-10
share_plus: errors on any share : ios / iPhone


espbee:
## System info

Platform the Issue occurs on: ios / iPhone
Plugin name: share_plus
Plugin version: 3.0.4

On the first run, there is a long lag before the share window pops up and although the share is successful, there is a messy error console log. The error continues each time though the lag seems to improve. **The lag itself does not happen on release builds, only debug.**

```
[ShareSheet] connection invalidated
[default] LaunchServices: store (null) or url (null) was nil: Error Domain=NSOSStatusErrorDomain Code=-54 "process may not map database" UserInfo={NSDebugDescription=process may not map database, _LSLine=264, _LSFunction=-[_LSDReadClient getServerStoreWithCompletionHandler:]}
[default] Attempt to map database failed: permission was denied. This attempt will not be retried.
[db] Failed to initialize client context with error Error Domain=NSOSStatusErrorDomain Code=-54 "process may not map database" UserInfo={NSDebugDescription=process may not map database, _LSLine=264, _LSFunction=-[_LSDReadClient getServerStoreWithCompletionHandler:]}
[default] -imageForImageDescriptor: can do IO please adopt -imageForDescriptor: for IO free drawing or -prepareImageForDescriptor: if IO is allowed. (This will become a fault soon.)
```

## Steps to Reproduce

Just run the simplest example of the share_plus plugin, like so:

```Dart


import 'package:flutter/material.dart';
import 'package:share_plus/share_plus.dart';


void main() {
  runApp(const MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({Key? key}) : super(key: key);

  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      title: 'Flutter Demo',
      theme: ThemeData(
        primarySwatch: Colors.blue,
      ),
      home: const MyHomePage(title: 'Flutter Demo Home Page'),
    );
  }
}

class MyHomePage extends StatefulWidget {
  const MyHomePage({Key? key, required this.title}) : super(key: key);
  final String title;

  @override
  State<MyHomePage> createState() => _MyHomePageState();
}

class _MyHomePageState extends State<MyHomePage> {

  void _share() {
    Share.share('check out my website https://example.com');
  }

  @override
  Widget build(BuildContext context) {
    return Scaffold(
      appBar: AppBar(
        title: Text(widget.title),
      ),
      body: Center(
          child: GestureDetector(
              onTap: () => _share(),
[...]

PR #1055, opened 2022-09-06
(share_plus) fix text share on iOS

miquelbeltran:
## Description

Following the issue reported in #730 and performing some tests, I've found that the share functionality on iOS doesn't seem to work well.

Trying to share text caused the app to show an empty share sheet (no apps listed).

Seems that the issue is caused by attempting to encode the incoming text string as a NSURL object.

I don't understand why we need to do that, but this causes issues.

This change removes that line of code, and sharing works again.

The following has been tested and worked:

- Sharing text with Telegram
- Sharing text + Image with Telegram (both appear)
- Sharing text + subject on Mail (text and subject are set correctly)
- Sharing text + subject + image on Mail (all three appear correctly - see screenshot in comments)

Gmail seems to have issues supporting the "subject" field, reported here: https://github.com/fluttercommunity/plus_plugins/issues/1056

## Related Issues

#730 

## Breaking Change

### DIFF
--- a/packages/share_plus/share_plus/CHANGELOG.md
+++ b/packages/share_plus/share_plus/CHANGELOG.md
@@ -1,3 +1,13 @@
+## 4.1.0
+
+- iOS: Fix text sharing.
+  - Previously, the text was being encoded as a URL, this caused the share sheet to appear empty.
+  - Now the shared text is not encoded as a URL anymore but rather shared as plain text.
+  - Sharing text + subject + attachments should work on apps that support that (e.g. Mail app).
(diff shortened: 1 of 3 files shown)


miquelbeltran:
The Subject issue is only present with GMAIL, but not with Apple MAIL app. So this is a bug on Gmail side :) 

Sharing Text + Subject + Image attachment works on the Mail app!

miquelbeltran:
photo_2022-09-06_14-37-11
photo_2022-09-06_14-37-07

PR #1057, opened 2022-09-06
Remove Codecov coverage reports from PRs

miquelbeltran:
## Description

Reporting code coverage sounds like a good idea, but it has some drawbacks:

1. Codecov only considers the coverage of Unit Tests.
2. Our code is mainly located in the native side of the plugin.
3. We rely on Integration Tests, which aren't considered for code coverage.
4. The report gives misleading results, marking a job as failed because the coverage changed.

Because of these, I'd like to remove the Codecov coverage reports.


mhadaily (approved):
That makes sense to me

PR #1059, opened 2022-09-07
Update manifest with explicit minimum sdk level

ChristianEdwardPadilla:
Need minimum of 22 to support Intent#createChooser.

## Breaking Change

### DIFF
--- a/packages/share_plus/share_plus/android/src/main/AndroidManifest.xml
+++ b/packages/share_plus/share_plus/android/src/main/AndroidManifest.xml
@@ -11,4 +11,5 @@
           android:resource="@xml/flutter_share_file_paths"/>
       </provider>
     </application>
+  <uses-sdk android:minSdkVersion="22" />
 </manifest>


miquelbeltran:
Hi @ChristianEdwardPadilla can you elaborate on this change?

My guess is that the plugin was originally using `Intent.createChooser(Intent, CharSequence)` https://developer.android.com/reference/android/content/Intent#createChooser(android.content.Intent,%20java.lang.CharSequence), which is available since API 1, but now it uses https://developer.android.com/reference/android/content/Intent#createChooser(android.content.Intent,%20java.lang.CharSequence,%20android.content.IntentSender) `Intent.createChooser(Intent, CharSequence, IntentSender) for the `shareWithResult` method.

So, if an app is still using `share`, they are fine without the API 22 requirement. However, if an app wants to use `shareWithResult` then it is mandatory to use API 22 at least.

I think it would be better to document this and/or add an exception if someone tries to use `shareWithResult` on API 21 or less.

vbuberen:
Closing as we are not merging this PR.

Issue #1081, opened 2022-09-20
[Bug]: Crash on sharing Sub-sub level url


maitycyrus:
### Platform

iOS

### Plugin

share_plus

### Version

4.3.0

### Flutter SDK

3.0.5

### Steps to reproduce

Share.share("https://developer.apple.com");

### Code Sample

```dart
Share.share("https://developer.apple.com");
```

### Logs

```shell
[assertion] Error acquiring assertion: <Error Domain=RBSServiceErrorDomain Code=1 "target is not running or doesn't have entitlement com.apple.runningboard.assertions.webkit" UserInfo={NSLocalizedFailureReason=target is not running or doesn't have entitlement com.apple.runningboard.assertions.webkit}>
[ProcessSuspension] 0x11b004f60 - ProcessAssertion::acquireSync Failed to acquire RBS assertion 'ConnectionTerminationWatchdog' for process with PID=36398, error: Error Domain=RBSServiceErrorDomain Code=1 "target is not running or doesn't have entitlement com.apple.runningboard.assertions.webkit" UserInfo={NSLocalizedFailureReason=target is not running or doesn't have entitlement com.apple.runningboard.assertions.webkit}
[assertion] Error acquiring assertion: <Error Domain=RBSAssertionErrorDomain Code=2 "Specified target process does not exist" UserInfo={NSLocalizedFailureReason=Specified target process does not exist}>
[ProcessSuspension] 0x11b004fc0 - ProcessAssertion::acquireSync Failed to acquire RBS assertion 'WebProcess Background Assertion' for process with PID=36398, error: Error Domain=RBSAssertionErrorDomain Code=2 "Specified target process does not exist" UserInfo={NSLocalizedFailureReason=Specified target process does not exist}
[assertion] Error acquiring assertion: <Error Domain=RBSAssertionErrorDomain Code=2 "Specified target process does not exist" UserInfo={NSLocalizedFailureReason=Specified target process does not exist}>
[ProcessSuspension] 0x11b005020 - ProcessAssertion::acquireSync Failed to acquire RBS assertion 'WebProcess Suspended Assertion' for process with PID=36398, error: Error Domain=RBSAssertionErrorDomain Code=2 "Specified target process does not exist" UserInfo={NSLocalizedFailureReason=Specified target process does not exist}
-[NSNull length]: unrecognized selector sent to instance 0x1f6ba0558
[...]

PR #1079, opened 2022-09-18
[share_plus] Fix Instagram does not show up in provider list for web links

neokn:
## Description

iOS: Fix Instagram does not show up in provider list for web links
  - issue #459 appear again
  - put back NSURL for the shareText, when text is pure URL
  - using LPMetadataProvider to get LPLinkMetadata make the user experience better

## Related Issues

[#459](https://github.com/fluttercommunity/plus_plugins/issues/459)

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`).
This will ensure a smooth and quick review process.

- [X] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [X] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [X] All existing and new tests are passing.
- [X] I updated the version in `pubspec.yaml` and `CHANGELOG.md`.
- [X] I updated/added relevant documentation (doc comments with `///`).
- [X] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [X] I read and followed the [Flutter Style Guide].
- [X] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [X] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/fluttercommunity/plus_plugins/blob/main/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning

PR #1082, opened 2022-09-21
Revert "[share_plus] Fix Instagram does not show up in provider list for web links (#1079)"

miquelbeltran:
## Description

Revert patch 4.2.0 on share_plus

## Related Issues

 - #1081
 - #1079


miquelbeltran:
Published under 4.4.0

PR #1095, opened 2022-09-26
Enable integration_test run on iOS Simulator

miquelbeltran:
## Description

Doing some work we did not finish with the GSOC.

I am enabling the integration_test to run on iOS.


miquelbeltran:
A problem may be that Linux doesn't have a desktop environment, so we cannot really launch apps to test them.

I may try MacOS instead of Linux

miquelbeltran:
So iOS works, time to enable all tests then :)

miquelbeltran:
Only one failed job and it was android ;)

PR #1110, opened 2022-09-29
[website] Add explicit entry point for Firebase Deploy Action

vbuberen:
## Description

Looks like Firebase Action doesn't respect the `defaults` syntax as it didn't find `firebase.json` file: https://github.com/fluttercommunity/plus_plugins/actions/runs/3154184239/jobs/5131506340

Trying to fix it by using available `entryPoint` parameter.


vbuberen:
Going to merge without review to check fix faster. Sorry.

PR #1113, opened 2022-09-30
feat: Add integration tests to alarm manager

LukeMoody01:
## Description

This PR adds integration tests to the android_alarm_manager_plus plugin and also adds these tests into the CI job on an android emulator

## Related Issues

This PR addresses the open issue #1103 

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`).
This will ensure a smooth and quick review process.

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [x] All existing and new tests are passing.
- [x] I updated the version in `pubspec.yaml` and `CHANGELOG.md`.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [x] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/fluttercommunity/plus_plugins/blob/main/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning

PR #1116, opened 2022-09-30
[Github Actions] Add ignored paths for Build website preview workflow

vbuberen:
## Description

Adding ignored paths to avoid triggering website preview build when non website related files are changed, so we don't have issues, like in #1113 where checks are marked as failed because of the failing website preview workflow.

## Breaking Change

### DIFF
--- a/.github/workflows/build_preview_website.yaml
+++ b/.github/workflows/build_preview_website.yaml
@@ -1,13 +1,9 @@
 name: Build website preview
 on:
   pull_request:
-    paths-ignore:
-      - "packages/**"
-      - ".github/**"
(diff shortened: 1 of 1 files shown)


nohli:
What about adding the other files in the root folder?

vbuberen:
I added paths that are most likely to be edited. But, probably, it is better to do the other way - add paths which should trigger the workflow

vbuberen:
Updated, so the job starts only when `docs` and `website` folders have changes.

nohli (approved):
Great ✌🏽

miquelbeltran (approved):

PR #1132, opened 2022-10-02
make toMap() json compatible

### LINKED ISSUE #949: [Request]: make toMap() json compatible

Yczar:
## Description

*Rendered toMap on the models to be jsonCompatible*

## Related Issues

* fix #949 

## Checklist

## Breaking Change

### DIFF
--- a/packages/device_info_plus/device_info_plus_platform_interface/lib/model/android_device_info.dart
+++ b/packages/device_info_plus/device_info_plus_platform_interface/lib/model/android_device_info.dart
@@ -138,6 +138,7 @@ class AndroidDeviceInfo implements BaseDeviceInfo {
   }
 
   /// `toJson()` returns a JSON string representation of [AndroidDeviceInfo] Model.
+  @override
   String toJson() => jsonEncode(toMap());
 
   /// Deserializes from the message received from [_kChannel].
--- a/packages/device_info_plus/device_info_plus_platform_interface/lib/model/android_device_info.dart
+++ b/packages/device_info_plus/device_info_plus_platform_interface/lib/model/android_device_info.dart
@@ -2,6 +2,8 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:convert';
+
 import 'base_device_info.dart';
 
 /// Information derived from `android.os.Build`.
@@ -135,6 +137,9 @@ class AndroidDeviceInfo implements BaseDeviceInfo {
     };
   }
 
+  /// `toJson()` returns a JSON string representation of [AndroidDeviceInfo] Model.
+  String toJson() => jsonEncode(toMap());
+
   /// Deserializes from the message received from [_kChannel].
   static AndroidDeviceInfo fromMap(Map<String, dynamic> map) {
     return AndroidDeviceInfo(
@@ -219,6 +224,9 @@ class AndroidBuildVersion {
       'securityPatch': securityPatch,
     };
   }
+
+  /// `toJson()` returns a JSON string representation of [AndroidBuildVersion] Model.
+  String toJson() => jsonEncode(toMap());
 
   /// Deserializes from the map message received from [_kChannel].
   static AndroidBuildVersion _fromMap(Map<String, dynamic> map) {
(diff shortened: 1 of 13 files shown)


miquelbeltran:
An `@overrides` is missing, check the analyze output.

Would it be possible to add unit tests for the toJson methods?

Yczar:
> An `@overrides` is missing, check the analyze output.
> 
> Would it be possible to add unit tests for the toJson methods?

Fixed this in the new push @miquelbeltran

Yczar:
> An `@overrides` is missing, check the analyze output.
> 
> Would it be possible to add unit tests for the toJson methods?

Fixed these on the latest push @miquelbeltran

iapicca:
@miquelbeltran 
shouldn't "toJson" return a json rather than a string?

Yczar:
> @miquelbeltran shouldn't "toJson" return a JSON rather than a string?

toJson should return a parseable string otherwise it is just toMap with a change of name. @iapicca

iapicca:
> > @miquelbeltran shouldn't "toJson" return a JSON rather than a string?
> 
> toJson should return a parseable string otherwise it is just toMap with a change of name. @iapicca

I'm saying that this PR doesn't the issue you linked 
the current problem is that `toMap` breaks dart:convert as it returns an `enum` somewhere

I fixed in this [PR](https://github.com/fluttercommunity/plus_plugins/pull/950), 
but I didn't get any answer in 2 months

your approach seems to be adding a `toJson` method rather than fixing `toMap`
similar approach as been previously described as [not needed](https://github.com/fluttercommunity/plus_plugins/pull/950#issuecomment-1185264150) by @vbuberen

and [quoting](https://github.com/fluttercommunity/plus_plugins/pull/950#issuecomment-1187329379) @miquelbeltran 

> I am sorry, I am not sure anymore if we need this. If a user wants to serialize the contents of a `DeviceInfo` object, they should do it on their own app data model. We shouldn't be maintaining a `toJson` implementation on our side, even if convenient, it is not our job. Like, I am not even sure why are we providing a toMap method even, I guess that's the legacy from when we got the plugins from Google.

Yczar:
> > > @miquelbeltran shouldn't "toJson" return a JSON rather than a string?
> > 
> > 
> > toJson should return a parseable string otherwise it is just toMap with a change of name. @iapicca
> 
> I'm saying that this PR doesn't the issue you linked the current problem is that `toMap` breaks dart:convert as it returns an `enum` somewhere
> 
> I fixed in this [PR](https://github.com/fluttercommunity/plus_plugins/pull/950), but I didn't get any answer in 2 months
> 
> your approach seems to be adding a `toJson` method rather than fixing `toMap` similar approach as been previously described as [not needed](https://github.com/fluttercommunity/plus_plugins/pull/950#issuecomment-1185264150) by @vbuberen
> 
> and [quoting](https://github.com/fluttercommunity/plus_plugins/pull/950#issuecomment-1187329379) @miquelbeltran
> 
> > I am sorry, I am not sure anymore if we need this. If a user wants to serialize the contents of a `DeviceInfo` object, they should do it on their own app data model. We shouldn't be maintaining a `toJson` implementation on our side, even if convenient, it is not our job. Like, I am not even sure why are we providing a toMap method even, I guess that's the legacy from when we got the plugins from Google.

All checks passed on this end... I feel we should leave it to the regulators to decide which approach is right or wrong... I noticed the enum was redundant Because there is already a function and a getter that generates the enum in the first place. So there was never a need for the breaks to happen. But if the maintainers find your approach interesting then I believe they would do the justice.

miquelbeltran:
@iapicca sorry, I forgot that we had your PR on this, I apologize for not having followed up.

Since the original issue has been closed, I will simply close this PR as well.

As I though originally, I am not in favor of adding extra functionality to the plugins that should be external to the plugin, like JSON support.

vbuberen:
> similar approach as been previously described as not needed by @vbuberen

Would be good if you could calm down a little bit first, because seems like the miscommunication issue and lack of response from the reviewer of your PR made you feel offended.

Secondly, I wasn't stating anything there, but was genuinely curios to find out what is the reason behind your changes. So would be also good to understand that such question might have a different meaning than what you implied.

PR #1138, opened 2022-10-02
[package_info_plus] add support for chrome-extensions

### LINKED ISSUE #1134: [Request]: Changes are required to allow Package Info Plus to work with browser extensions as well
### Plugin

package_info_plus_web

### Use case

https://github.com/fluttercommunity/plus_plugins/blob/e7391f8f8224c020ff60bb59524788bc1d5280ed/packages/package_info_plus/package_info_plus_web/lib/package_info_plus_web.dart#L24-L31

Only http and https are allowed because `baseUri.origin` is used.

`_SimpleUri.origin` is supposed to cause an error when a scheme other than http and https comes in.

However, Browser Extension has the following URL: `chrome-extension://`

Why don't we change the features to make them available on more diverse platforms?

### Proposal

https://github.com/fluttercommunity/plus_plugins/blob/e7391f8f8224c020ff60bb59524788bc1d5280ed/packages/package_info_plus/package_info_plus_web/lib/package_info_plus_web.dart#L26

Instead of `origin`, please provide various schemas as acceptable properties

werainkhatri:
## Description

`dart:core`'s `Uri.origin` does not support schemes other than `http` and `https`. This PR adds a workaround by extending the `Uri` class with an `_origin` getter which returns the origin for `chrome-extensions://` w/o throwing `StateError` by combining `Uri.scheme` and `Uri.host`.

This will enable ONLY `chrome-extensions` other than `http` and `https`, nothing else, going along the same lines as dart sdk authors did by preventing anything other than `http` and `https`. If there is some other use case, adding support for that will be trivial.

## Related Issues

Fixes #1134

## Checklist

## Breaking Change

### DIFF
--- a/packages/package_info_plus/package_info_plus_web/CHANGELOG.md
+++ b/packages/package_info_plus/package_info_plus_web/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.0.6
+
+- Add support for chrome extensions.
+
 ## 1.0.5
 
 - Resolve package_name
--- a/packages/package_info_plus/package_info_plus_web/CHANGELOG.md
+++ b/packages/package_info_plus/package_info_plus_web/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 1.0.6
 
-- Add support for chrome extensions.
+- Add support for chrome extensions
 
 ## 1.0.5
 
--- a/packages/package_info_plus/package_info_plus_web/lib/package_info_plus_web.dart
+++ b/packages/package_info_plus/package_info_plus_web/lib/package_info_plus_web.dart
@@ -64,7 +64,7 @@ extension _UriOrigin on Uri {
   /// Get origin.
   ///
   /// This is different from [Uri.origin] because that has checks to prevent
-  /// non- http/https use cases.
+  /// non-http/https use-cases.
   String get _origin {
     if (isScheme('chrome-extension')) {
       return '$scheme://$host';
--- a/packages/package_info_plus/package_info_plus_web/lib/package_info_plus_web.dart
+++ b/packages/package_info_plus/package_info_plus_web/lib/package_info_plus_web.dart
@@ -20,10 +20,30 @@ class PackageInfoPlugin extends PackageInfoPlatform {
     PackageInfoPlatform.instance = PackageInfoPlugin();
   }
 
+  /// Get origin from [uri].
+  ///
+  /// This is different from [Uri.origin] because that does not support
+  /// non- http / https use cases. This function is a workaround,
+  /// specifically for chrome extensions.
+  String origin(Uri uri) {
+    try {
+      return uri.origin;
+    } on StateError catch (e) {
+      String getErrorMessage([bool to = false]) =>
+          'Origin is only applicable ${to ? 'to ' : ''}schemes http and https: chrome-extension://';
+
+      if (e.message.startsWith(getErrorMessage()) ||
+          e.message.startsWith(getErrorMessage(true))) {
+        return '${uri.scheme}://${uri.authority}';
+      }
+      rethrow;
+    }
+  }
+
   /// Get version.json full url.
   Uri versionJsonUrl(String baseUrl, int cacheBuster) {
     final baseUri = Uri.parse(baseUrl);
-    final originPath = '${baseUri.origin}${baseUri.path}';
+    final originPath = '${origin(baseUri)}${baseUri.path}';
     final versionJson = 'version.json?cachebuster=$cacheBuster';
     return Uri.parse(originPath.endsWith('/')
         ? '$originPath$versionJson'
--- a/packages/package_info_plus/package_info_plus_web/lib/package_info_plus_web.dart
+++ b/packages/package_info_plus/package_info_plus_web/lib/package_info_plus_web.dart
@@ -20,30 +20,10 @@ class PackageInfoPlugin extends PackageInfoPlatform {
     PackageInfoPlatform.instance = PackageInfoPlugin();
   }
 
-  /// Get origin from [uri].
-  ///
-  /// This is different from [Uri.origin] because that does not support
-  /// non- http / https use cases. This function is a workaround,
-  /// specifically for chrome extensions.
-  String origin(Uri uri) {
-    try {
-      return uri.origin;
-    } on StateError catch (e) {
-      String getErrorMessage([bool to = false]) =>
-          'Origin is only applicable ${to ? 'to ' : ''}schemes http and https: chrome-extension://';
-
-      if (e.message.startsWith(getErrorMessage()) ||
-          e.message.startsWith(getErrorMessage(true))) {
-        return '${uri.scheme}://${uri.authority}';
-      }
-      rethrow;
-    }
-  }
-
   /// Get version.json full url.
   Uri versionJsonUrl(String baseUrl, int cacheBuster) {
     final baseUri = Uri.parse(baseUrl);
-    final originPath = '${origin(baseUri)}${baseUri.path}';
+    final originPath = '${baseUri._origin}${baseUri.path}';
     final versionJson = 'version.json?cachebuster=$cacheBuster';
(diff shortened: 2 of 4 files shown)


miquelbeltran:
I don't have the necessary web-dev knowledge to review this, asking the rest of maintainers/community if they can chip in

vbuberen:
> I don't have the necessary web-dev knowledge to review this

Same here.

nohli:
@werainkhatri is there maybe another workaround except catching the `StateError`?
Thanks for adding tests 🙂

packages/package_info_plus/package_info_plus_web/lib/package_info_plus_web.dart
```diff
@@ -20,10 +20,30 @@ class PackageInfoPlugin extends PackageInfoPlatform {
...
+  /// specifically for chrome extensions.
+  String origin(Uri uri) {
+    try {
+      return uri.origin;
+    } on StateError catch (e) {
+      String getErrorMessage([bool to = false]) =>
+          'Origin is only applicable ${to ? 'to ' : ''}schemes http and https: chrome-extension://';
+
+      if (e.message.startsWith(getErrorMessage()) ||
+          e.message.startsWith(getErrorMessage(true))) {
```
nohli, line 36: The bool `to` could be difficult to read, as well as the method generating both error messages.

How would you like this?
```suggestion
      const errorPart1 = 'Origin is only applicable';
      const errorPart2 = 'http and https: chrome-extension://';

      if (e.message.contains(errorPart1) && e.message.contains(errorPart2) {
```

werainkhatri:
> is there maybe another workaround except catching the StateError?

i have instead created sort of a replica of the `Uri.origin` function, without the checks. this seems to me, a better work-around.

nohli:
Nice!

If we replicate `origin`, shouldn't we replicate the other checks, too?

Screen Shot 2022-10-04 at 18 00 43

`origin()` is only used in this library, correct? Then it should be private.
Or, what about an extension?

```
extension _UriOrigin on Uri {
  String get _origin { // or a better name
    if (isScheme('HTTP') || isScheme('HTTPS')) {
      return origin;
    } else if (port == 0) {
      // some comment what port == 0 means
      return '$scheme://$host';
    } else {
      return '$scheme://$host:$port';
    }
  }
}
```

```
buildSignature: Uri.parse(window.document.baseUri!)._origin,
```

werainkhatri:
I had this doubt initially. if we anyway wish to support "only" chrome extensions for now, might as well add a check specifically for it and rest be handled by `Uri.origin`.

```dart
extension _UriOrigin on Uri {
  String get _origin {
    if (isScheme('chrome-extension')) {
      return '$scheme://$host:$port';
    }
    return origin;
  }
}
```

packages/package_info_plus/package_info_plus_web/lib/package_info_plus_web.dart
```diff
@@ -58,3 +58,16 @@ class PackageInfoPlugin extends PackageInfoPlatform {
...
 }
+
+extension _UriOrigin on Uri {
+  /// Get origin.
+  ///
+  /// This is different from [Uri.origin] because that has checks to prevent
+  /// non- http/https use cases.
+  String get _origin {
+    if (isScheme('chrome-extension')) {
+      return '$scheme://$host:$port';
```
nohli, line 69: Don't we need this logic here?
```
int? port = _port;
if (port == null) return "$scheme://$host";
return "$scheme://$host:$port";
```

packages/package_info_plus/package_info_plus_web/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.0.6
+
+- Add support for chrome extensions.
```
nohli, line 3: Nit:
```suggestion
- Add support for chrome extensions
```

packages/package_info_plus/package_info_plus_web/test/package_info_plus_web_test.dart
```diff
@@ -16,6 +16,7 @@ void main() {
     'build_number': '1',
     'package_name': 'io.flutter.plugins.packageinfoexample',
     'version': '1.0',
+    'build_signature': '',
```
nohli, line 19: Something with a value would be better to test the function.

```suggestion
    'build_signature': 'test_signature',
```

nohli (changes requested):
Only a few small questions/nit picking.

Would be nice to update the description as well (as this PR isn't catching the error any more, but enabling something for the case `isScheme('chrome-extension')` that just threw an exception before).

From my side it looks good 🙂

packages/package_info_plus/package_info_plus_web/test/package_info_plus_web_test.dart
werainkhatri: yes, but the code does not set this variable.

nohli: Ok, just found it, nevermind.
`Contains the signing key signature (hex) on Android and empty string on everything else.`
Do we need to 'test' this?

packages/package_info_plus/package_info_plus_web/lib/package_info_plus_web.dart
werainkhatri: a few observations:
1. i don't think chrome extensions support ports, although i wasn't able to find anything backing this. (one theory is that it's not a server, so what will the port manager be? unlike nginx used in backend)
2. actually, having and not having the port seems to be same in case of chrome extensions. if you pass a port in the url, chrome just removes it once you press enter.

so given this, i think it's safe to remove the port from here too. (i didn't do this before cause i wasn't completely sure of the above two observations).

packages/package_info_plus/package_info_plus_web/test/package_info_plus_web_test.dart
werainkhatri: since it should be empty in case of web, i don't see a downside in testing it. bonus to prevent the mistake like i made of not reverting.

<details>
<summary>what and why did i do that</summary>

well, i wanted to verify what the different functions return in case of chrome extensions, but i can't run the via flutter run as it's that. so i thought of using one of the PackageInfo model fields to get the url. but after testing, i forgot to remove it.
<details>

packages/package_info_plus/package_info_plus_web/lib/package_info_plus_web.dart
```diff
@@ -58,3 +59,16 @@ class PackageInfoPlugin extends PackageInfoPlatform {
     }
   }
 }
+
+extension _UriOrigin on Uri {
+  /// Get origin.
+  ///
+  /// This is different from [Uri.origin] because that has checks to prevent
+  /// non- http/https use cases.
```
nohli, line 67: nit:
```suggestion
  /// non-http/https use-cases.
```

nohli:
@werainkhatri just more one small comment for improving the docs

From my side it's ready to go, good job! 💪🏽

nohli (approved):
LGTM!
Thanks for your contribution ❤️

packages/package_info_plus/package_info_plus_web/pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: package_info_plus_web
 description: Web platform implementation of package_info_plus
-version: 1.0.5
+version: 1.0.6
```
werainkhatri, line 3: i just realized, since this is a new feature, would it make sense to increase the minor version instead? `1.1.0`? not sure how this should work

werainkhatri: also, should have updated the version in `package_info_plus` package. will raise a separate PR for that after resolving my previous doubt.

nohli: No worries 😊

@miquelbeltran possibly one more federated version fix for you ✌🏽

PR #1153, opened 2022-10-03
[share_plus] Enable Web implementations for `shareFiles`

### LINKED ISSUE #212: Missing share_plus share files implementation in web
There is a share files method in the html.navigator, see [here](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share)

ThexXTURBOXx:
## Description

This PR is a continuation of #318.
There were a few issues in that original implementation:
 - The int arrays were passed as a concatenated int string to the sharing program
 - File name being forgotten
 - Mime Types missing (causing `Permission Denied` errors)
 - The `withResult` implementations were missing
 - The examples were not updated properly
 - Tests were missing

This PR aims at fixing these issues as they seemed to be the remaining ones.
Also, I have tested the feature on Chrome (Android + PC) and Edge (PC) and it worked in my test cases (multiple files, different mime types etc.).
Firefox is sadly not supported (due to Browser limitations, see [here](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share#browser_compatibility) for details).

## Related Issues

Fixes #212

## Checklist

## Breaking Change

### DIFF
--- a/packages/share_plus/share_plus/lib/share_plus.dart
+++ b/packages/share_plus/share_plus/lib/share_plus.dart
@@ -66,6 +66,7 @@ class Share {
   ///
   /// May throw [PlatformException] or [FormatException]
   /// from [MethodChannel].
+  @Deprecated("Use shareXFiles instead.")
   static Future<void> shareFiles(
     List<String> paths, {
     List<String>? mimeTypes,
@@ -78,42 +79,6 @@ class Share {
     return _platform.shareFiles(
       paths,
       mimeTypes: mimeTypes,
-      subject: subject,
-      text: text,
-      sharePositionOrigin: sharePositionOrigin,
-    );
-  }
-
-  /// Summons the platform's share sheet to share multiple files.
-  ///
-  /// Wraps the platform's native share dialog. Can share a file.
-  /// It uses the `ACTION_SEND` Intent on Android and `UIActivityViewController`
-  /// on iOS.
-  ///
-  /// Android supports all natively available MIME types (wildcards like image/*
-  /// are also supported) and it's considered best practice to avoid mixing
-  /// unrelated file types (eg. image/jpg & application/pdf). If MIME types are
-  /// mixed the plugin attempts to find the lowest common denominator. Even
-  /// if MIME types are supplied the receiving app decides if those are used
-  /// or handled.
-  /// On iOS image/jpg, image/jpeg and image/png are handled as images, while
-  /// every other MIME type is considered a normal file.
-  ///
-  /// The optional `sharePositionOrigin` parameter can be used to specify a global
-  /// origin rect for the share sheet to popover from on iPads and Macs. It has no effect
-  /// on other devices.
-  ///
-  /// May throw [PlatformException] or [FormatException]
-  /// from [MethodChannel].
-  static Future<void> shareXFiles(
-    List<XFile> files, {
-    String? subject,
-    String? text,
-    Rect? sharePositionOrigin,
-  }) {
-    assert(files.isNotEmpty);
-    return _platform.shareXFiles(
-      files,
       subject: subject,
       text: text,
       sharePositionOrigin: sharePositionOrigin,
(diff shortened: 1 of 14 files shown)


ThexXTURBOXx:
Sorry for the odd title at first. Something on my end appearently went wrong there. Fixed it, though!

daniel-shuy:
The `share_plus` example currently uses [image_picker](https://pub.dev/packages/image_picker), which does not work on Flutter Web. The example should be updated to use [image_picker_web](https://pub.dev/packages/image_picker_web) if the platform is web, so that this feature can be tested with the `share_plus` example using a mobile/emulator browser, by running the example on Web Server.

ThexXTURBOXx:
I have successfully tested the example on web, even the image picker. Why should the image picker not work there?

packages/share_plus/share_plus_web/lib/share_plus_web.dart
```diff
@@ -61,7 +62,107 @@ class SharePlusPlugin extends SharePlatform {
...
+        context: DiagnosticsNode.message('while trying to share file(s)'),
+        library: 'share_plus_web',
+        informationCollector: () => [
+          DiagnosticsNode.message(
+            'The web share API is not available on the current browser. '
+            'Falling back to downloading the files. '
+            'Subject and text are discarded if given.',
+          ),
+        ],
+      ));
```
miquelbeltran, line 118: Should we call directly into the FlutterError error sink? 

I think it would be better to throw an exception instead and let the developer catch it on their side.

packages/share_plus/share_plus_platform_interface/lib/method_channel/method_channel_share.dart
```diff
@@ -69,6 +70,26 @@ class MethodChannelShare extends SharePlatform {
...
   }
 
+  /// Summons the platform's share sheet to share multiple files.
+  @override
+  Future<void> shareXFiles(
+    List<XFile> files, {
+    String? subject,
+    String? text,
+    Rect? sharePositionOrigin,
+  }) {
```
miquelbeltran, line 80: Is it necessary to have the `shareXFiles` method in the platform interface, rather than being a private method of the web implementation?

packages/share_plus/share_plus/lib/share_plus.dart
```diff
@@ -82,6 +84,42 @@ class Share {
...
+  /// or handled.
+  /// On iOS image/jpg, image/jpeg and image/png are handled as images, while
+  /// every other MIME type is considered a normal file.
+  ///
+  /// The optional `sharePositionOrigin` parameter can be used to specify a global
+  /// origin rect for the share sheet to popover from on iPads and Macs. It has no effect
+  /// on other devices.
+  ///
+  /// May throw [PlatformException] or [FormatException]
+  /// from [MethodChannel].
```
miquelbeltran, line 107: Could you document here instead what is the difference between `shareFile` and `shareXFile` and when should `shareXFile` be used?

I understand we need `shareXFile` because `XFile` is what web uses, correct?

miquelbeltran:
I am looking into it and added some questions, but overall I wonder if we could avoid adding a new share method for `XFiles`.

Also, don't create two new methods with and without result, but rather just one method with result. The share methods without result should be legacy.

The final goal is to have a unique universal share method as the current implementation is a mess.

ThexXTURBOXx:
@miquelbeltran Thank you very much for the feedback. I will implement it as soon as possible today. I will remove the shareXFiles function without result and deprecate the `shareFiles*` functions since they are obsolete now with the `shareXFiles*` functions.
However, I won't unify all functions into one in this PR since it is hugely out of scope and described in another issue.

miquelbeltran:
Sounds good to me, thanks!

packages/share_plus/share_plus/lib/share_plus.dart
ThexXTURBOXx: Basically yes, `XFile` is a cross-platform `File` implementation which makes it very handy to use on all platforms.

packages/share_plus/share_plus_platform_interface/lib/method_channel/method_channel_share.dart
ThexXTURBOXx: This is also solved because of the deprecations

packages/share_plus/share_plus_web/lib/share_plus_web.dart
ThexXTURBOXx: Done, the exceptions are now uncaught

miquelbeltran (approved):
LGTM! thanks for applying all suggestions

Issue #851, opened 2022-05-01
[network_info_plus] request location permission for getWifiBSSID() and getWifiName() on Android


ekuleshov:
## Use case

The iOS plugin relies on `requestLocationServiceAuthorization()` call to provide access to `getWifiName()` and `getWifiBSSID()`.

But on Android this functionality is not available and the app has to implement some Android-version specific logic to request corresponding android permissions.

## Proposal

The Android plugin can use the same `requestLocationServiceAuthorization()` API call to request required Android permissions. Implementation can request permissions as required for a given Android version, e.g. `ACCESS_FINE_LOCATION` on Android 10+ and `ACCESS_COARSE_LOCATION` for Android <10.

PR #1155, opened 2022-10-04
fix(network_info_plus): Mark iOS-specific permission methods as deprecated and update docs

shaynekellyii:
## Description

Updates README.md to explain users should use permission_handler with this package for both iOS and Android.

Marks iOS-specific methods as `@Deprecated`, telling developers to check the README.md for further instructions.

## Related Issues

https://github.com/fluttercommunity/plus_plugins/issues/851

## Checklist

- [ ] I updated the version in `pubspec.yaml` and `CHANGELOG.md`.

## Breaking Change


vbuberen:
I don't agree with enforcing additional permissions, which are not needed. In docs there is a list of cases when this new permission is required: https://developer.android.com/guide/topics/connectivity/wifi-permissions#check-for-apis-that-require-permission
As of today `Network Info Plus` uses neither of mentioned APIs, so this new permission should be removed.

Will do the proper review later.

shaynekellyii:
Ah, I misinterpreted that part of the Android docs. So it would be preferable to request `ACCESS_FINE_LOCATION` on anything about Android 10 and remove `NEARBY_WIFI_DEVICES` entirely?

miquelbeltran:
Hey! Due to changes in our contributor guidelines, all PRs should now:

- Have a title that follows the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0-beta.4/) format. e.g. `feat(package_name)!: feature description`
- **DO NOT** modify the CHANGELOG.md or the version in the `pubspec.yaml`. This is going to be an automated process now.

We would ask you kindly to update the PR following these changes.

Thanks!

packages/network_info_plus/network_info_plus/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 2.3.0
+- Android: Add `getLocationServiceAuthorization` and `requestLocationServiceAuthorization` permissions support
+
```
miquelbeltran, line 3: Please revert this change, check the new CONTRIBUTOR guidelines, you don't have to update the CHANGELOG anymore.

packages/network_info_plus/network_info_plus/README.md
```diff
@@ -41,15 +41,44 @@ var wifiGateway = await info.getWifiGatewayIP(); // 192.168.1.1
...
+...
+<!-- Android >=13 -->
+<uses-permission
+        android:name="android.permission.NEARBY_WIFI_DEVICES"
+        android:usesPermissionFlags="neverForLocation" />
+<!-- Android 10-12 -->
+<uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
+<!-- Android <10 -->
+<uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
+```
```
miquelbeltran, line 77: Thanks for updating the documentation!

packages/network_info_plus/network_info_plus/android/src/main/AndroidManifest.xml
```diff
@@ -1,5 +1,11 @@
...
-  <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
+    package="dev.fluttercommunity.plus.network_info">
+
+    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
+    <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
+    <uses-permission
+        android:name="android.permission.NEARBY_WIFI_DEVICES"
+        android:usesPermissionFlags="neverForLocation" />
+    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
+    <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
```
miquelbeltran, line 10: Do we need to have these in the AndroidManifest of the plugin? We are already telling the developer to add them in the README.md.

packages/network_info_plus/network_info_plus/android/src/main/kotlin/dev/fluttercommunity/plus/network_info/NetworkInfo.kt
```diff
@@ -1,15 +1,42 @@
...
+        grantResults: IntArray
+    ): Boolean {
+        if (requestCode == permissionRequestCode) {
+            val cachedResult = ongoingResult ?: return true
+            cachedResult.success(parseAuthStatus(grantResults).platformName)
+            ongoingResult = null
+        }
+        return true
+    }
+
```
miquelbeltran, line 39: I personally would not include code to request the permissions in the plugin. There's plugins that can already do that better than what we can: e.g. https://pub.dev/packages/permission_handler

miquelbeltran:
Hi @shaynekellyii thanks for your contribution, but I am sorry I am not sure if I understand the purpose of the PR.

Developers can already handle permissions using other plugins like https://pub.dev/packages/permission_handler

What advantage would it have to offer this functionality in this plugin?

shaynekellyii:
> Hi @shaynekellyii thanks for your contribution, but I am sorry I am not sure if I understand the purpose of the PR.
> 
> Developers can already handle permissions using other plugins like https://pub.dev/packages/permission_handler
> 
> What advantage would it have to offer this functionality in this plugin?

It was mentioned in the linked issue (https://github.com/fluttercommunity/plus_plugins/issues/851) that iOS has the method `requestLocationServiceAuthorization()` in this plugin, while devs have to use their own permission logic on Android.

That's okay, I can close this PR and comment on the issue that it would be better to use `permission_handler` plugin in this case.

miquelbeltran:
Ah yes, thanks, you explained it well in the description.

Yep, I am not sure, my guess is that the previous dev team thought it's necessary to have the iOS permission handling in the plugin.

Just a random idea: Would it better to refactor the `requestLocationServiceAuthorization` to use internally the permission_handler package?

Or if not, would you be able to write down on the README the instructions to set up the necessary permissions and how to use the permission_handler plugin to accomplish that? At least, I think other devs will find that very useful.

@vbuberen what do you think about this?

nohli:
permission_handler is very actively maintained and I am pro using it under the hood, or adding it to the readme (and removing the iOS-only methods).

In my opinion, it doesn't make sense to keep the method just for one platform in the long-run, while the majority of Flutter apps is cross-platform and would need permission_handler anyway.

shaynekellyii:
> In my opinion, it doesn't make sense to keep the method just for one platform in the long-run, while the majority of Flutter apps is cross-platform and would need permission_handler anyway.

I generally agree with this, and was thinking we should just use permission_handler if anything under the hood after I finished this PR 😅 

In my opinion, I also think the two methods could be removed and README updated to suggest using permission_handler.

miquelbeltran:
Thanks for the comments! I'd suggest:

- Update README.md to explain how to do the same using permission_handler with this package for both iOS and Android.
- Mark the iOS method as `@Deprecated`, tell developers that the method will be removed and  to check the README.md for further instructions.

vbuberen:
As a user of the plugin I would love to see `permission_handler` integrated, but from experience with some other plugins which integrate functionality based on another package, adding `permission_handler` as a dependency would mean that we have to keep up with releases of it so we don't have issues from people that version resolution failed.

If we are talking about native implementation, like suggested in this PR, we would also limit developers in possible options with where they ask permissions, if they want to show some additional dialog, etc.

So if we don't do what is suggested in this PR to complement the iOS implementation (don't know its history or why it was added by maintainers some time ago, by the way), I would do as @miquelbeltran suggests - mark iOS as deprecated and provide info on how devs should handle the permission topic.

shaynekellyii:
Marked both iOS-only methods as deprecated and updated README with a short description telling plugin users to use `permission_handler`. 

I did not add an example to the `network_info_plus` README as the `permission_handler` README has examples. (reduces the need for this package to update the README if `permission_handler` changes their API)

miquelbeltran (approved):
LGTM!

vbuberen (approved):
Looks good to me as well.

nohli (approved):

ekuleshov:
It is very unfortunate this feature got dropped and the general decision is to completely ignore permissions in the `network_info_plus` code and make users to deal with them.

The problem is that permissions are different on every platform AND also are different between platform's OS versions. The application developer who uses `network_info_plus` plugin not only has to use another 3rd party plugin and also have to add bunch of code to handle permission requests and not only for individual platforms, but also for for different OS versions. E.g. to access SSID, the location permission on Android is required since Android 12.

Such obscure platform-specific  requirements are significantly reducing simplicity of use of the cross-platform plugins like `network_info_plus`.

Also, the suggested `permission_nandler` plugin does not have support for MacOS and Web and in the past it took a while to see updates for the newly added platform-specific permissions.

YawarOsman:
still having the same issue on google pixel 4xl (android 13)

PR #1169, opened 2022-10-05
refactor(battery_plus)!: platform implementation refactor into a single package

miquelbeltran:
## Description

After multiple discussions of the maintainer team, we came to the conclusion that maintaining a federated plugin architecture was more painful than brought any advantages.

This PR includes the work to "unfederate" **battery_plus**.

I decided to pick this package since there are no open PRs targeting this plugin.

**Note to reviewers:** The PR is easier to see if you checkout the branch.

### Work Done

To perform the "unfederation":

- I moved all code from the respective plugins into the main battery_plus plugin.
- Then, I renamed all references to the external plugins to the main one, e.g. `battery_plus_linux` became `battery_plus`.
- Also, updated the references to the paths to be local to the package.
- Some unit tests that existed on the platform packages were moved.
- The package_interface package also suffered the same fate as the other platform plugins and the code was moved to the main plugin. The platform interface still exists and could theoretically be used by an external plugin that implements a new platform (e.g. Fuchsia?)

Update:

- battery_plus_package_interface will remain as independent package.

## Breaking Change

- [x] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).


miquelbeltran:
Windows build is not working, will try to fix tomorrow.

packages/battery_plus/lib/battery_plus.dart
```diff
@@ -4,12 +4,14 @@
...
 
-import 'package:battery_plus_platform_interface/battery_plus_platform_interface.dart';
+import 'package:battery_plus/battery_plus_platform_interface.dart';
 
 // Export enums from the platform_interface so plugin users can use them directly.
-export 'package:battery_plus_platform_interface/battery_plus_platform_interface.dart'
+export 'package:battery_plus/battery_plus_platform_interface.dart'
     show BatteryState;
 
+export 'src/battery_plus_linux.dart';
```
jpnurmi, line 13: i wish we could do something like
```
linux:
  dartPluginClass: BatteryPlusLinux
  fileName: src/battery_plus_linux.dart
```
in `pubspec.yaml` to avoid this export but it would probably require changes in the flutter tool.

miquelbeltran: I will check if that is possible

nohli:
Yayy the tests are passing 😊

packages/battery_plus/lib/battery_plus.dart
miquelbeltran: I tried that, but I was faced with this error:

```
Launching lib/main.dart on Linux in debug mode...
ERROR: .dart_tool/flutter_build/dart_plugin_registrant.dart:20:9: Error: Undefined name 'BatteryPlusLinux'.
ERROR:         BatteryPlusLinux.registerWith();
ERROR:         ^^^^^^^^^^^^^^^^
Building Linux application...                                           
Exception: Build process failed
```

When I removed the `export src/battery_plus_linux.dart`

miquelbeltran:
> I'm curious, is there a technical reason to have `battery_plus_web.dart` in `lib/` instead of `lib/src/`?

This has been addressed ;)

miquelbeltran:
> Yayy the tests are passing blush

I am happy that I didn't need to do any CI changes! Only to make sure that the projects compiled.

I had to switch to my Windows PC to fix the Windows build but that was it.

packages/battery_plus/lib/battery_plus.dart
jpnurmi: Well, da*m. Thanks for trying anyway. :) Looks like there's some inconsistent `filename`/`fileName` handling in [flutter_tools/lib/src/platform_plugins.dart](https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/platform_plugins.dart). Web uses `fileName` for Dart plugins, whereas Windows and Linux use `filename` but only for C++ plugins...

nohli:
The six folders can be deleted, right?

Why is there no linux folder and is `pluginClass: none` correct?

Screen Shot 2022-10-06 at 08 07 58

miquelbeltran:
> The six folders can be deleted, right? Edit: Nevermind, was only because they contained files not checked into git. Is there maybe a way to delete these folders in local clones although they contain untracked files?

The folders will still exist on your local folder, a `git clean -dxf` will take care of that for you.

> 
> Why is there no linux folder and is `pluginClass: none` correct?

I believe so, since the Linux implementation is purely in Dart. The `pluginClass: none` existed in the `battery_plus_linux` pubspec so I copied it over. I am not 100% sure if it is required or not.

jpnurmi:
> I believe so, since the Linux implementation is purely in Dart. The `pluginClass: none` existed in the `battery_plus_linux` pubspec so I copied it over. I am not 100% sure if it is required or not.

It was a temporary workaround that is probably not needed anymore: https://github.com/flutter/flutter/issues/57497

nohli:
How about moving everything into `src` folder?
And perhaps renaming `method_channel_battery_plus.dart` to `battery_plus_method_channel.dart`?

Screen Shot 2022-10-06 at 08 33 53

jpnurmi:
One more, is it possible to move `method_channel_battery_plus.dart` into `src/` as well so that we'd only have `battery_plus.dart` and the platform interface in the root?

jpnurmi:
> Edit: Do we need the platform interface in the lib folder (not in src)?

It might actually be better to leave the platform interface at the root (or at least export it?) for the theoretical case that someone might want to implement an external platform plugin because [implementation_imports](https://dart.dev/tools/linter-rules#implementation_imports) is enabled by default in lints/recommended and flutter_lints.

packages/battery_plus/lib/battery_plus.dart
```diff
@@ -4,11 +4,10 @@
...
 import 'dart:async';
 
-import 'package:battery_plus_platform_interface/battery_plus_platform_interface.dart';
+import 'src/battery_plus_platform_interface.dart';
 
-// Export enums from the platform_interface so plugin users can use them directly.
-export 'package:battery_plus_platform_interface/battery_plus_platform_interface.dart'
-    show BatteryState;
+export 'src/battery_plus_linux.dart';
+export 'src/battery_plus_platform_interface.dart';
```
nohli, line 11: Can this be removed because of line 14?

miquelbeltran:
That has been addressed and seems to work, now all files are inside `src` and the platform interface class is exported from the main `battery_plus.dart`

jpnurmi:
Sorry for going back and forth but now that I think of it, having a separate `battery_plus_platform_interface.dart` at the top level (instead of exporting all of the platform interface) would avoid polluting the usual import with `BatteryPlatform`.

jpnurmi:
Very nice! :+1: LGTM but we need Majid's blessing for this...

miquelbeltran:
> Sorry for going back and forth but now that I think of it, having a separate `battery_plus_platform_interface.dart` at the top level (instead of exporting all of the platform interface) would avoid polluting the usual import with `BatteryPlatform`.

Right. Also, we can now export the `enums.dart` directly instead of going through the platform_interface file.

packages/battery_plus/pubspec.yaml
```diff
@@ -14,31 +14,30 @@ flutter:
...
       linux:
-        default_package: battery_plus_linux
+        dartPluginClass: BatteryPlusLinux
       macos:
-        default_package: battery_plus_macos
+        pluginClass: BatteryPlusMacosPlugin
       web:
-        default_package: battery_plus_web
+        pluginClass: BatteryPlusPlugin
+        fileName: src/battery_plus_web.dart
```
nohli, line 22: Can the web plugin be renamed to `BatteryPlusWebPlugin` (and do we need `fileName`)?

packages/battery_plus/pubspec.yaml
```diff
@@ -14,31 +14,30 @@ flutter:
       ios:
         pluginClass: FLTBatteryPlusPlugin
```
nohli, line 15: Just curious - is this naming because we copied code from somewhere else?
(`BatteryPlusIosPlugin` would match the others)

packages/battery_plus/pubspec.yaml
miquelbeltran: Renamed! But the fileName seems to be necessary

nohli: Sorry, I updated my comment. It's no plugin, right? Could it just be `BatteryPlusWeb`?

nohli: I would either add it to the linux one, or remove it from web. Would be happy with either option.
Both would make sense (as they're pure Dart and therefore technically no plugins (if I'm correct), it could be omitted - but since all platform implementations are called plugin, it would also make sense to name all consistently).

miquelbeltran: I renamed the Linux one for consistency with the other 5 platforms

nohli: Thanks a lot 👍🏽

nohli:
LGTM! ❤️ 

Also letting @mhadaily give the approval 🙂

packages/battery_plus/battery_plus/lib/battery_plus.dart
```diff
@@ -37,7 +39,7 @@ class Battery {
 
   /// check if device is on battery save mode
   ///
-  /// Currently only impemented on Android, IOS and Windows.
+  /// Currently only implemented on Android, IOS and Windows.
```
nohli, line 42: Since we're fixing typos, `iOS` would be better

miquelbeltran: I couldn't let it go, been seeing the typo the whole day :joy:

miquelbeltran:
I wonder if we should modify this CI check:

```
 Downloading battery_plus_platform_interface 1.2.1...
Package validation found the following potential issue:
* Your dependency on "battery_plus_platform_interface" should allow more than one version. For example:
  
  dependencies:
    battery_plus_platform_interface: ^1.2.1
  
  Constraints that are too tight will make it difficult for people to use your package
  along with other packages that also depend on "battery_plus_platform_interface".
```

packages/battery_plus/battery_plus/lib/battery_plus.dart
nohli: That's the spirit!

vbuberen (approved):
Looks good to me

PR #1184, opened 2022-10-09
ci(android_alarm_manager_plus): Add matrix of versions for integration tests

vbuberen:
## Description

Introduced build matrix so tests run against multiple Android versions. It should help us to be sure that functionality of the plugin isn't broken by some change.

This PR has matrix only for Android alarm manager plus to get initial feedback. If other maintainers agree with suggested strategies in the next PR I will add similar changes for all plugins at once.

Initially I wanted to have 19 as a minimal version, but on 19 I saw such error:
Screenshot 2022-10-09 at 19 43 37

So it seems that `integration_test` doesn't work well with old APIs.
Mid version corresponds to Android O, which is something in the middle of the suggested range.
Max version is 32 as 33 acted strange during my tests on fork.

## Checklist

## Breaking Change

### DIFF
--- a/.github/workflows/android_alarm_manager_plus.yaml
+++ b/.github/workflows/android_alarm_manager_plus.yaml
@@ -29,7 +29,7 @@ jobs:
       - name: "Install Tools"
         run: ./.github/workflows/scripts/install-tools.sh
 
-      # Required for Android builds
+          # Required for Android builds
       - uses: actions/setup-java@v1
         with:
           java-version: "11"
@@ -40,15 +40,19 @@ jobs:
   android_integration_test:
     runs-on: macos-latest
     timeout-minutes: 30
+    strategy:
+      matrix:
+        android-api-level: [21, 26, 32]
+
     steps:
       - name: "Checkout repository"
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: "Install Flutter"
         run: ./.github/workflows/scripts/install-flutter.sh stable
       - name: "Install Tools"
         run: ./.github/workflows/scripts/install-tools.sh
 
-      # Required for Android builds
+          # Required for Android builds
       - uses: actions/setup-java@v1
         with:
           java-version: "11"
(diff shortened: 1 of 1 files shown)


.github/workflows/android_alarm_manager_plus.yaml
```diff
@@ -23,15 +23,16 @@ jobs:
...
+        uses: actions/checkout@v3
       - name: "Install Flutter"
         run: ./.github/workflows/scripts/install-flutter.sh stable
       - name: "Install Tools"
         run: ./.github/workflows/scripts/install-tools.sh
 
-      # Required for Android builds
-      - uses: actions/setup-java@v1
+        # Required for Android builds
+      - uses: actions/setup-java@v2
```
vbuberen, line 33: I experimented with removing this step as suggested in one of other PRs (because of https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md#java saying that such runners already have Java 11), but couldn't build the project due to missing Java 11. Thus, returned the step again.

.github/workflows/android_alarm_manager_plus.yaml
```diff
@@ -23,15 +23,16 @@ jobs:
...
         with:
+          distribution: "temurin"
```
vbuberen, line 35: setup-java in v2 requires to specify the distribution type, thus, this addition.

.github/workflows/android_alarm_manager_plus.yaml
```diff
@@ -40,26 +41,33 @@ jobs:
...
 
       - name: "Bootstrap Workspace"
         run: melos bootstrap
       - name: "Android Integration Test"
         uses: reactivecircus/android-emulator-runner@v2
         with:
-          api-level: 30
-          arch: x86_64
+          api-level: ${{ matrix.android-api-level }}
+          cores: 3
```
vbuberen, line 68: `MacOS` runner [has 3 cores available](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources), while `android-emulator-runner` uses 2 by default. Specifying this value in effort to get higher performance.

.github/workflows/android_alarm_manager_plus.yaml
```diff
@@ -40,26 +41,33 @@ jobs:
   android_integration_test:
     runs-on: macos-latest
     timeout-minutes: 30
+    strategy:
+      matrix:
+        android-api-level: [21, 26, 32]
```
miquelbeltran, line 46: Ohh that's cool, simpler than what I thought.

miquelbeltran (approved):
LGTM, let's see if the CI gods keep everything green ;)

PR #1185, opened 2022-10-10
ci(all): Enforce using Conventional Commit PR title

miquelbeltran:
## Description

Enforces the use of Conventional Commits on all PRs.

This ensures that all work can be managed using `melos version`, as it requires following Conventional Commits strictly.

Also makes reading PR changes easier.

I used the example provided by https://github.com/marketplace/actions/semantic-pull-request to configure it.

## Related Issues

None

## Checklist

## Breaking Change

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,0 +1,15 @@
+name: "Pull-Request Check"
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - synchronize
+
+jobs:
+  main:
+    name: Validate PR title
+    runs-on: ubuntu-latest
+    steps:
+      - uses: amannn/action-semantic-pull-request@v4
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -13,3 +13,30 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: amannn/action-semantic-pull-request@v4
+        id: lint_pr_title
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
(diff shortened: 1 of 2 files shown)


vbuberen:
I think we need to move this check into a separate workflow, because `all_plugins` runs not for all commits (see `ignored-paths`), while we would want to be sure in all incoming commits.

Also, I see that the referenced workflow uses https://github.com/marketplace/actions/semantic-pull-request action and I would just use it directly.

.github/workflows/pr.yml
```diff
@@ -0,0 +1,42 @@
+name: "Pull-Request Check"
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - synchronize
```
miquelbeltran, line 8: This workflow will run on PR changes, not code changes

.github/workflows/pr.yml
```diff
@@ -0,0 +1,42 @@
...
+
+jobs:
+  main:
+    name: Validate PR title
+    runs-on: ubuntu-latest
+    steps:
+      - uses: amannn/action-semantic-pull-request@v4
+        id: lint_pr_title
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
miquelbeltran, line 18: I am not sure if this is required, I copied it over from the example

miquelbeltran:
@vbuberen I have addressed your points. I am unable to check if this really works as it doesn't seem to trigger if it is in a branch and not on main, so we would have to wait until merged to see it in action.

mhadaily:
This is great

nohli (approved):
Great idea, let's see if it works 😊

vbuberen (approved):

PR #1190, opened 2022-10-11
ci(all): Add Dependabot config

vbuberen:
## Description

Introducing Dependabot config to automate dependencies updates for all Github Actions, website and plugins dependencies.

The config is that big because I had to specify exact folder for all `pubspec.yaml` files we have in the repo as Dependabot doesn't support multiple directories search yet: https://github.com/dependabot/dependabot-core/issues/2824

Current setup is basic with once in a week check, but we can customise quite a lot of params if needed:
https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates

Commits will follow a proper conventional commits naming, so it won't be breaking for the newly introduced contribution process:
Screenshot 2022-10-11 at 14 36 17

Be aware that during its first run Dependabot will open lots of PRs. If needed, we can set an upper limit for amounts of PRs as well to not spam our `Pull requests` section first time. But I would leave it as it is for now.

## Checklist

## Breaking Change

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,412 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+
+# Github Actions dependencies updates config
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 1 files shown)


miquelbeltran (approved):
LGTM!

vbuberen:
> Will there still be an opened PR (with failing lints), so we can manually fix the code?

Yes, such PR is still opened. Also, there is a badge with compatibility indicator, which is based on stats among other repos using this dependency: https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/about-dependabot-security-updates#about-compatibility-scores

PR #1242, opened 2022-10-15
test(device_info_plus): Add Android display metrics check, fix Android version check

vbuberen:
## Description

After adding an Android build matrix saw that [integration tests fail ](https://github.com/fluttercommunity/plus_plugins/actions/runs/3244740493/jobs/5321314320)for `device_info_plus` on newly added Andoid APIs. As it turned out there was a hardcoded version of Android (30), which caused test to fail on all other versions.
Also, saw that same test was missing check for a newly added device metrics data, so added it as well.

## Checklist

## Breaking Change

### DIFF
--- a/packages/device_info_plus/device_info_plus/example/integration_test/device_info_plus_test.dart
+++ b/packages/device_info_plus/device_info_plus/example/integration_test/device_info_plus_test.dart
@@ -78,14 +78,16 @@ void main() {
 
   testWidgets('Check all android info values are available',
       (WidgetTester tester) async {
-    expect(androidInfo.version.baseOS, isNotNull);
+    if (androidInfo.version.sdkInt >= 23) {
+      expect(androidInfo.version.baseOS, isNotNull);
+      expect(androidInfo.version.previewSdkInt, isNotNull);
+      expect(androidInfo.version.securityPatch, isNotNull);
+    }
+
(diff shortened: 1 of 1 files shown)


nohli (approved):
LGTM!

miquelbeltran (approved):

vbuberen:
Test was still failing on API 21 because some properties are not available till API 23:
https://developer.android.com/reference/android/os/Build.VERSION#BASE_OS
https://developer.android.com/reference/android/os/Build.VERSION#PREVIEW_SDK_INT
https://developer.android.com/reference/android/os/Build.VERSION#SECURITY_PATCH

And the implementation had the check for API version, thus, returning `null` on lower versions:
https://github.com/fluttercommunity/plus_plugins/blob/main/packages/device_info_plus/device_info_plus/android/src/main/java/dev/fluttercommunity/plus/device_info/MethodCallHandlerImpl.java#L73

Fixed it as well.

nohli:
Is it somewhere in our docs that they return null on lower versions?

vbuberen:
> Is it somewhere in our docs that they return null on lower versions?

No. we don't have it. I will create a separate PR as I see one more possible improvement for our docs on the website for this plugin.

nohli:
If even we don't really know, we should put it in the docs of the methods that return null for specific versions.

Issue #1270, opened 2022-10-19
[Bug]: No implementation found for method getAll on channel dev.fluttercommunity.plus/package_info


CodeDoctorDE:
### Platform

Linux (Fedora 36)

### Plugin

package_info_plus

### Version

3.0.0

### Flutter SDK

3.3.4

### Steps to reproduce

1. Upgrade to the latest package_info_plus 3.0.0
2. Use `await PackageInfo.fromPlatform()`

After downgrading to 2.0.0, everything works.

I cloned the example from the `main` branch and everything works there.

### Code Sample

```dart
Future<bool> hasNewerVersion() async {
    final packageInfo = await PackageInfo.fromPlatform();
    return packageInfo.version != kVersion;
}
const kVersion = "1.0.0";
```

### Logs

```shell
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: MissingPluginException(No implementation found for method getAll on channel dev.fluttercommunity.plus/package_info)
#0      MethodChannel._invokeMethod
package:flutter/…/services/platform_channel.dart:294
<asynchronous suspension>
#1      MethodChannel.invokeMapMethod
package:flutter/…/services/platform_channel.dart:499
<asynchronous suspension>
#2      MethodChannelPackageInfo.getAll
package:package_info_plus_platform_interface/method_channel_package_info.dart:13
<asynchronous suspension>
#3      PackageInfo.fromPlatform
package:package_info_plus/package_info_plus.dart:41
<asynchronous suspension>
#4      SettingsCubit.hasNewerVersion
package:butterfly/cubits/settings.dart:424
<asynchronous suspension>
#5      _ProjectPageState._showIntroduction
package:butterfly/views/main.dart:216
<asynchronous suspension>
```

### Flutter Doctor

```shell
[✓] Flutter (Channel stable, 3.3.4, on Fedora Linux 36 (Workstation Edition) 5.19.15-201.fc36.x86_64, locale de_DE.UTF-8)
    • Flutter version 3.3.4 on channel stable at /home/jan-andrej/Tools/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision eb6d86ee27 (vor 2 Wochen), 2022-10-04 22:31:45 -0700
    • Engine revision c08d7d5efc
    • Dart version 2.18.2
    • DevTools version 2.15.0

[✓] Android toolchain - develop for Android devices (Android SDK version 33.0.0)
[...]

Issue #1274, opened 2022-10-19
[Question]: Is generated_plugin_registrant.dart necessary for web for device_info_plus and package_info_plus?


naamapps:
### What is your question?

Hey, 
After upgrading `device_info_plus` and `package_info_plus` to latest versions,
the `generated_plugin_registrant.dart` file does not generate the `PackageInfoPlugin.registerWith(registrar);` and `DeviceInfoPlusPlugin.registerWith(registrar);` lines anymore (It used to generate those lines, so after upgrading I had this error
```
Undefined name 'PackageInfoPlugin'.
Try correcting the name to one that is defined, or defining the name
```

Is this the right behaviour?

Thanks.

### Checklist before submitting a question

- [X] I Google'd a solution and I couldn't find it
- [X] I searched on StackOverflow for a solution and I couldn't find it
- [X] I read the README.md file of the plugin
- [X] I am using the latest version of the plugin
- [X] All dependencies are up to date with `flutter pub upgrade`
- [X] I did a `flutter clean`
- [X] I tried running the example project

miquelbeltran:
Looks related to https://github.com/fluttercommunity/plus_plugins/issues/1270

miquelbeltran:
Please check if with the latest release of the package(s) still happens

naamapps:
Hey @miquelbeltran,
Still won't generate for me.

I think this line is the culprit
https://github.com/fluttercommunity/plus_plugins/blob/37abbf7b05a32bc1e7b440c19540b277e301fe5a/packages/device_info_plus/device_info_plus/lib/device_info_plus.dart#L26 

If I try to write `DeviceInfoPlusWebPlugin` manually inside the `generated_plugin_registrant` file, it won't recognize this class.

naamapps:
Same for PackageInfo.

miquelbeltran:
mhh, so adding the export for the web plugin implementation causes the generated_plugin_registrant not to work properly? If you can checkout the project locally, reference by path, and then find what causes it, it will be super helpful (also feel free to submit a PR if you find the solution).

naamapps:
Is the file web_plugin_registrant the same as generated_plugin_registrant?
I can't find it in my project at all..

I tried deleting the generated_plugin_registrant file and it didn't even regenerate itself.

miquelbeltran:
It is in  `.dart_tool/dartpad/web_plugin_registrant.dart` it appears for me when running the project

[...]

PR #1275, opened 2022-10-20
fix(all): Increase min Flutter version to fix dartPluginClass registration

miquelbeltran:
## Description

Fixes the issue where the `dart_plugin_registrant.dart` file is not generated correctly.

Tested this locally with packages referenced by path. When I increased the Flutter SDK version to 2.11 then the issue we saw in #1270 was gone.

## Related Issues

- Fix: https://github.com/fluttercommunity/plus_plugins/issues/1270
- Related: https://github.com/flutter/flutter/pull/96610
- Related: https://github.com/flutter/flutter/issues/113720
- Related: https://github.com/fluttercommunity/plus_plugins/issues/1274

## Checklist

## Breaking Change

### DIFF
--- a/packages/battery_plus/battery_plus/pubspec.yaml
+++ b/packages/battery_plus/battery_plus/pubspec.yaml
@@ -42,4 +42,4 @@ dev_dependencies:
 
 environment:
   sdk: ">=2.12.0 <3.0.0"
-  flutter: ">=1.20.0"
+  flutter: ">=2.11.0"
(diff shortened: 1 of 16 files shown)


miquelbeltran:
Since we are at it, I have also upgraded the platform interfaces (although not required for this fix) and removed some of the SDK requirements from the examples, as they are not required.

CodeDoctorDE:
Why do we need the platform_interface packages if there is only one package for everything?

miquelbeltran:
> Why do we need the platform_interface packages if we only need one package for every platform?

It has been a long discussion in the maintainers team, and the conclusion is that having a package interface as an external package allows for expandability. I'd personally get rid of it but not everyone agreed.

vbuberen (approved):
This is great that we are pushing things a little bit and getting rid of these really old versions as well as bumping all plugins at once, getting rid of differences in min supported version.

PR #1295, opened 2022-10-26
feat(share_plus): remove direct dependence of url_launcher

### LINKED ISSUE #1070: [Request]: share_plus android remove url_launcher
### Plugin

share_plus

### Use case

url_launcher is not required when compiling android but it will also be loaded into the app's gradle project

### Proposal

Remove url_launcher when compiling android

SirusCodes:
## Description

Remove direct dependence of url_launcher, instead depending on platform specific packages.

## Related Issues

resolves #1070

## Checklist

- [ ] I did not modify the `CHANGELOG.md` nor the `pubspec.yaml` files.

## Breaking Change

### DIFF
--- a/packages/share_plus/share_plus/lib/src/share_plus_linux.dart
+++ b/packages/share_plus/share_plus/lib/src/share_plus_linux.dart
@@ -4,10 +4,13 @@ library share_plus_linux;
 import 'dart:ui';
 
 import 'package:share_plus_platform_interface/share_plus_platform_interface.dart';
-import 'package:url_launcher/url_launcher.dart';
+import 'package:url_launcher_linux/url_launcher_linux.dart';
+import 'package:url_launcher_platform_interface/url_launcher_platform_interface.dart';
 
 /// The Linux implementation of SharePlatform.
 class SharePlusLinuxPlugin extends SharePlatform {
+  final urlLauncher = UrlLauncherLinux();
+
   /// Register this dart class as the platform implementation for linux
   static void registerWith() {
     SharePlatform.instance = SharePlusLinuxPlugin();
@@ -35,7 +38,11 @@ class SharePlusLinuxPlugin extends SharePlatform {
           .join('&'),
     );
 
-    await launchUrl(uri);
+    if (await urlLauncher.canLaunch(uri.toString())) {
+      await urlLauncher.launchUrl(uri.toString(), const LaunchOptions());
+    } else {
+      throw Exception('Unable to share on web');
+    }
   }
 
   /// Share files.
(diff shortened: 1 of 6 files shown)


miquelbeltran:
Is this PR still in progress?

SirusCodes:
Yes, I just need to fix some test.

miquelbeltran (approved):
I like the idea. I don't fully understand why it was a problem to have url_launcher as a dependency originally, but I like that now it is injected as param, easier to mock or replace even.

miquelbeltran:
I will have to test on Monday before merging

SirusCodes:
> I don't fully understand why it was a problem to have url_launcher as a dependency originally

Someone wanted to save a few KBs in their app, so why not.

miquelbeltran:
Thanks!

packages/share_plus/share_plus/lib/src/share_plus_linux.dart
```diff
@@ -35,7 +40,11 @@ class SharePlusLinuxPlugin extends SharePlatform {
           .join('&'),
     );
 
-    await launchUrl(uri);
+    if (await urlLauncher.canLaunch(uri.toString())) {
+      await urlLauncher.launchUrl(uri.toString(), const LaunchOptions());
+    } else {
+      throw Exception('Unable to share on web');
+    }
   }
```
stuartmorgan, line 48: FYI: changing to this construction will break functionality, and has no advantages; see https://pub.dev/packages/url_launcher#checking-supported-schemes

SirusCodes: So, it's better to have only `launchUrl` in try catch block?

SirusCodes: [Checking the return of launchUrl](https://pub.dev/packages/url_launcher#checking-supported-schemes:~:text=if%20(!await%20launchUrl(uri))%20%7B%0A%20%20throw%20%27Could%20not%20launch%20%24uri%27%3B%0A%7D) is better construct?

stuartmorgan: > So, it's better to have only `launchUrl` in try catch block?

I'm not sure what the goal of a try/catch would be given that you are currently throwing.

> [Checking the return of launchUrl](https://pub.dev/packages/url_launcher#checking-supported-schemes:~:text=if%20(!await%20launchUrl(uri))%20%7B%0A%20%20throw%20%27Could%20not%20launch%20%24uri%27%3B%0A%7D) is better construct?

If you don't have a specific need to know in advance of trying (which is rare, and certainly isn't the case given the code here), checking what actually happened is better than using a potentially inaccurate prediction of what might happen, yes.

nohli: @miquelbeltran FYI

miquelbeltran: If anyone can send a PR with the fix, I can review and merge asap

SirusCodes: Yeah I can do that.

PR #1314, opened 2022-10-31
feat(share_plus): Show destination for share with result in example, update example UI

vbuberen:
## Description

Saw that example app for `share_plus` shows only type of result in case with `shareWithResult` call, but not the destination. So adding a destination info as well to showcase the feature better for package users.

Additionally decided to give the example app UI a little visual improvement, so made it to use Material 3 and did small changes to make it look more modern and clean to use:

## Checklist

## Breaking Change

### DIFF
--- a/packages/share_plus/share_plus/example/lib/main.dart
+++ b/packages/share_plus/share_plus/example/lib/main.dart
@@ -37,110 +37,118 @@ class DemoAppState extends State<DemoApp> {
   Widget build(BuildContext context) {
     return MaterialApp(
       title: 'Share Plus Plugin Demo',
+      theme: ThemeData(
+        useMaterial3: true,
+        colorScheme: ColorScheme.fromSwatch(primarySwatch: Colors.indigo),
+      ),
       home: Scaffold(
-          appBar: AppBar(
-            title: const Text('Share Plus Plugin Demo'),
-          ),
-          body: SingleChildScrollView(
-            child: Padding(
-              padding: const EdgeInsets.all(24.0),
-              child: Column(
-                crossAxisAlignment: CrossAxisAlignment.start,
-                children: <Widget>[
-                  TextField(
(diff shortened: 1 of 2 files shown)


packages/share_plus/share_plus/example/lib/main.dart
```diff
@@ -208,15 +237,18 @@ class DemoAppState extends State<DemoApp> {
...
     );
 
-    scaffoldMessenger.showSnackBar(SnackBar(
-      content: Text("Share result: ${result.status}"),
-    ));
+    scaffoldMessenger.showSnackBar(
+      SnackBar(
+        content: Text("Share result: ${result.status}"),
+      ),
+    );
```
miquelbeltran, line 252: Only one possible improvement, it would be to refactor this showSnackBar with the one in line 217 into one method, so you can call both passing the result object.

miquelbeltran (approved):
LGTM, just left a comment on a possible improvement

nohli (approved):

packages/share_plus/share_plus/example/lib/main.dart
vbuberen: Updated

packages/share_plus/share_plus/example/lib/main.dart
```diff
@@ -180,43 +199,54 @@ class DemoAppState extends State<DemoApp> {
...
       sharePositionOrigin: box!.localToGlobal(Offset.zero) & box.size,
     );
 
-    scaffoldMessenger.showSnackBar(SnackBar(
-      content: Text("Share result: ${result.status}"),
-    ));
+    scaffoldMessenger.showSnackBar(getResultSnackBar(shareResult));
+  }
+
+  SnackBar getResultSnackBar(ShareResult result) {
```
nohli, line 239: Could be private

PR #1315, opened 2022-11-01
fix(share_plus): remove `canLaunch` check

SirusCodes:
## Description

Reverting back to old construct as @stuartmorgan suggested https://github.com/fluttercommunity/plus_plugins/pull/1295#discussion_r1009240852

Also remove `canLaunch` check.

## Checklist

## Breaking Change

### DIFF
--- a/packages/share_plus/share_plus/lib/src/share_plus_linux.dart
+++ b/packages/share_plus/share_plus/lib/src/share_plus_linux.dart
@@ -40,11 +40,7 @@ class SharePlusLinuxPlugin extends SharePlatform {
           .join('&'),
     );
 
-    if (await urlLauncher.canLaunch(uri.toString())) {
-      await urlLauncher.launchUrl(uri.toString(), const LaunchOptions());
-    } else {
-      throw Exception('Unable to share on web');
-    }
+    await urlLauncher.launchUrl(uri.toString(), const LaunchOptions());
   }
 
   /// Share files.
--- a/packages/share_plus/share_plus/lib/src/share_plus_linux.dart
+++ b/packages/share_plus/share_plus/lib/src/share_plus_linux.dart
@@ -44,7 +44,7 @@ class SharePlusLinuxPlugin extends SharePlatform {
       const LaunchOptions(),
     );
     if (!launchResult) {
-      throw Exception('Failed to launch mailto: URI');
+      throw Exception('Failed to launch $uri');
     }
   }
 
--- a/packages/share_plus/share_plus/lib/src/share_plus_linux.dart
+++ b/packages/share_plus/share_plus/lib/src/share_plus_linux.dart
@@ -19,7 +19,6 @@ class SharePlusLinuxPlugin extends SharePlatform {
   }
 
   /// Share text.
-  /// Throws a [PlatformException] if `mailto:` scheme cannot be handled.
   @override
   Future<void> share(
     String text, {
@@ -40,7 +39,13 @@ class SharePlusLinuxPlugin extends SharePlatform {
           .join('&'),
     );
 
-    await urlLauncher.launchUrl(uri.toString(), const LaunchOptions());
+    final launchResult = await urlLauncher.launchUrl(
+      uri.toString(),
+      const LaunchOptions(),
+    );
+    if (!launchResult) {
+      throw Exception('Failed to launch mailto: URI');
+    }
   }
 
   /// Share files.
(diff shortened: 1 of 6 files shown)


stuartmorgan:
Why only Linux? Other implementations were using the same construction.

packages/share_plus/share_plus/lib/src/share_plus_linux.dart
```diff
@@ -40,11 +40,7 @@ class SharePlusLinuxPlugin extends SharePlatform {
           .join('&'),
     );
 
-    if (await urlLauncher.canLaunch(uri.toString())) {
-      await urlLauncher.launchUrl(uri.toString(), const LaunchOptions());
-    } else {
-      throw Exception('Unable to share on web');
-    }
+    await urlLauncher.launchUrl(uri.toString(), const LaunchOptions());
```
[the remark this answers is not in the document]
SirusCodes: I just reverted it back. Initially it was not returning anything. Also, the API is not returning anything to the user.

SirusCodes:
> Why only Linux? Other implementations were using the same construction.

Ahh... missed it. Thought it was only used in Linux. Will change them as well.

packages/share_plus/share_plus/lib/src/share_plus_linux.dart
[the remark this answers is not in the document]
SirusCodes: cc: @miquelbeltran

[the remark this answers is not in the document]
miquelbeltran: For what I can see, we can check the result and throw an exception if it was false.

I think using a plain `Exception` would be fine here. I will push the changes.

packages/share_plus/share_plus/lib/src/share_plus_linux.dart
```diff
@@ -19,7 +19,6 @@ class SharePlusLinuxPlugin extends SharePlatform {
   }
 
   /// Share text.
-  /// Throws a [PlatformException] if `mailto:` scheme cannot be handled.
```
miquelbeltran, line 22: Removed for consistency with the other platforms

packages/share_plus/share_plus/lib/src/share_plus_linux.dart
```diff
@@ -40,10 +39,12 @@ class SharePlusLinuxPlugin extends SharePlatform {
...
-    if (await urlLauncher.canLaunch(uri.toString())) {
-      await urlLauncher.launchUrl(uri.toString(), const LaunchOptions());
-    } else {
-      throw Exception('Unable to share on web');
+    final launchResult = await urlLauncher.launchUrl(
+      uri.toString(),
+      const LaunchOptions(),
+    );
+    if (!launchResult) {
+      throw Exception('Failed to launch mailto: URI');
```
miquelbeltran, line 47: As commented, the launchUrl can either return false or fail with an exception, in the case of returning false we throw an Exception.

I did not use `PlatformException` here because I don't think we should be creating them outside platform code, so I think just a generic `Exception` is fine here.

packages/share_plus/share_plus/test/url_launcher_mock.dart
```diff
@@ -0,0 +1,30 @@
...
+        required bool useSafariVC,
+        required bool useWebView,
+        required bool enableJavaScript,
+        required bool enableDomStorage,
+        required bool universalLinksOnly,
+        required Map<String, String> headers,
+        String? webOnlyWindowName,
+      }) async {
+    this.url = url;
+    return canLaunchMockValue;
```
miquelbeltran, line 28: By returning the `canLaunchMockValue` here, we can test that the package indeed throws an exception when the `launch` method returns false.

packages/share_plus/share_plus/lib/src/share_plus_linux.dart
SirusCodes: Yeah, it makes sense.

packages/share_plus/share_plus/lib/src/share_plus_windows.dart
```diff
@@ -3,6 +3,7 @@ library share_plus_windows;
 
 import 'dart:ui';
 
+import 'package:flutter/services.dart';
```
SirusCodes, line 6: Removing this redundant import

SirusCodes:
@miquelbeltran thanks for helping out. This feels fine to me

nohli (approved):
Looks good, just not sure if the thrown messages should include $uri instead of mailto: URI.

vbuberen (approved):
LGTM

PR #1330, opened 2022-11-11
fix(package_info_plus): remove html file parts from version url

### LINKED ISSUE #855: [package_info_plus][web] can not load version.json when URI path contains index.html
Our Flutter web app is embedded in iframe of an external web platform using some URI pattern like `https://external-domain.com/index.html`. Because the path contains `index.html` package_info_plus_web fails to load version information from `version.json`.

## System info

Platform the Issue occurs on: web
Plugin name: package_info_plus_web
Plugin version: 1.0.5

## Steps to Reproduce

Create a local Flutter app with a sample page which is available as 
`http://127.0.0.1:8081/#/my-sample-page`

The same page should be available when we modify the path like

`http://127.0.0.1:8081/index.html#/my-sample-page`

Using the second URI the path in `versionJsonUrl` is wrong so the method fails to load `version.json`

In our case it would be sufficient to just remove the special case of `index.html` from the path in `versionJsonUrl`, e.g.:
```dart
  /// Get version.json full url.
  Uri versionJsonUrl(String baseUrl, int cacheBuster) {
    final baseUri = Uri.parse(baseUrl);
    final originPath =
        '${baseUri.origin}${baseUri.path.replaceAll('index.html', '')}';
    final versionJson = 'version.json?cachebuster=$cacheBuster';
    return Uri.parse(originPath.endsWith('/')
        ? '$originPath$versionJson'
        : '$originPath/$versionJson');
  }
```

But maybe this issue needs a more general solution?

kaptnkoala:
## Description

This pull request tries to provide a more general approach to remove the concrete reference an html file added to the path where the main flutter web entry point is hosted. We have a setup a facebook instant games, where the flutter index html is not served directly at the server root folder and is also explicitly called with the full url http://some-url.com/index.html. In this case the versionJsonUrl is not correct and the json file can not be loaded.

## Related Issues

- *Fix #855*

## Checklist

## Breaking Change


miquelbeltran:
I read the original thread and I think it makes sense, it shouldn't break existing webapps.

The PR is missing tests, so I'd like to ask you to implement a unit test that showcases this functionality using the `versionJsonUrl` method.

kaptnkoala:
> I read the original thread and I think it makes sense, it shouldn't break existing webapps.
> 
> The PR is missing tests, so I'd like to ask you to implement a unit test that showcases this functionality using the `versionJsonUrl` method.

Where should the test for the extension of the `versionJsonUrl` method go? Is the `package_info_plus_web_test_disabled.dart` test disabled as the name suggests? Please give more concrete direction, where i should implement the requested test!

vbuberen:
@kaptnkoala Thanks for pinging and getting back to this PR after such a long time. From the first look it looks fine to me. I will do a proper review a bit later today and if all goes well this PR will go into the next release which I plan to do today/tomorrow.

vbuberen (approved):
Looks fine to me.

aytunch:
Hi,
In the documentation of this plugin for `version` getter, it says:
`String version = packageInfo.version;`
> The package version. CFBundleShortVersionString on iOS, versionName on Android. 
And nothing about Web Platform. But when I saw this `version.json` under the buiId/web folder, I wonder if we can get version on web and it is just unupdated documentation.
So can you clarify if we can use `packageInfoversion` on web platform reliably?

Issue #857, opened 2022-05-07
[connectivity_plus] [macOS] ConnectivityResult.none when VPN is turned on


deimantasa:
## System info

Platform the Issue occurs on: **macOS**
Plugin name: **connectivity_plus**
Plugin version: **2.3.0**

## Steps to Reproduce

1. Run the app with VPN -> `ConnectivityResult.none`
2. Turn off VPN -> `ConnectivityResult.wifi`
3. Turn on VPN -> `ConnectivityResult.none`

<details>
<summary>Flutter doctor output</summary>

```
[✓] Flutter (Channel stable, 2.10.5, on macOS 12.3.1 21E258 darwin-arm, locale en-HK)
    • Flutter version 2.10.5 at /Users/Aurimas/FlutterProject/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 5464c5bac7 (3 weeks ago), 2022-04-18 09:55:37 -0700
    • Engine revision 57d3bac3dd
    • Dart version 2.16.2
    • DevTools version 2.9.2

[✓] Android toolchain - develop for Android devices (Android SDK version 30.0.2)
    • Android SDK at /Users/Aurimas/Library/Android/sdk
    • Platform android-32, build-tools 30.0.2
    • ANDROID_SDK_ROOT = /Users/mac/Library/Android/sdk
    • Java binary at: /Applications/Android Studio.app/Contents/jre/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 11.0.11+0-b60-7772763)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 13.3)
    • Xcode at /Applications/Xcode-13.3-beta.app/Contents/Developer
    • CocoaPods version 1.11.2

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 2021.1)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build 11.0.11+0-b60-7772763)

[✓] VS Code (version 1.51.1)
    • VS Code at /Applications/Visual Studio Code.app/Contents
[...]

PR #1335, opened 2022-11-15
fix(connectivity_plus):  Do not return ConnectivityResult.none on iOS and MacOS with VPN

mit-73:
## Description

Added support of `ConnectivityResult.vpn` for iOS (12+) and MacOS (10.14+).

## Related Issues

- *Related #857*

## Checklist

## Breaking Change


mit-73:
@nohli @miquelbeltran ?

miquelbeltran:
sorry, I was going to review and then this fell off my plate, will check now

packages/connectivity_plus/connectivity_plus/ios/Classes/PathMonitorConnectivityProvider.swift
```diff
@@ -19,6 +19,11 @@ public class PathMonitorConnectivityProvider: NSObject, ConnectivityProvider {
         // .wiredEthernet is available in simulator
         // but for consistency it is probably correct to report .wifi
         return .wifi
+      } else if path.usesInterfaceType(.other) {
+        // path.status == .satisfied means that the network is available
+        // and .other is available in simulator
+        // for consistency, it would probably be correct to report .vpn
+        return .vpn
```
miquelbeltran, line 26: I am not convinced about this. Of course, "none" is not correct either, but as far as I understand we don't have the certainty that this is a VPN.

Maybe adding another type `other` would be better?

mit-73: I wouldn't want to break the common interface with the android version. What's here, what's in android, vpn doesn't guarantee network connectivity.

It says "other" in the documentation:`The network interface type used for communication over virtual networks or networks of unknown types.`

Anyway, I agree with you that "other" is not "vpn", waiting for a final decision from you.

miquelbeltran: > The network interface type used for communication over virtual networks or networks of unknown types.

Thanks for pointing this out! VPNs do indeed appear as "other", so it kind of makes sense as well.

Any thoughts @nohli @vbuberen ?

nohli: > Anyway, I agree with you that "other" is not "vpn", waiting for a final decision from you.

@miquelbeltran I don't get the last sentence. I would be fine with changing line 26 to `other`.
If it should be added *additionally* to `vpn`, which argument would be passed to `usesInterfaceType()` for `other`?

mit-73: @nohli @miquelbeltran should I change `vpn` to `other` for iOS/MacOS?

vbuberen: No need to tag people all the time. Those, who are in a thread will get an email anyway.

Yes, agree with suggestion from Miguel to have `other` returned for such cases.

mit-73: Okay, I changed it to `other`.

packages/connectivity_plus/connectivity_plus_platform_interface/lib/src/enums.dart
```diff
@@ -16,5 +16,10 @@ enum ConnectivityResult {
   none,
 
   /// VPN: Device connected to a VPN
-  vpn
+  ///
+  /// Note that on iOS and macOS comes [other], not [vpn]
+  vpn,
+
+  /// OTHER: Device connected to any unknown network
```
nohli, line 23:
```suggestion
  /// Other: Device is connected to an unknown network
```

packages/connectivity_plus/connectivity_plus_platform_interface/lib/src/enums.dart
```diff
@@ -16,5 +16,10 @@ enum ConnectivityResult {
   none,
 
   /// VPN: Device connected to a VPN
-  vpn
+  ///
+  /// Note that on iOS and macOS comes [other], not [vpn]
```
nohli, line 20: Could you write an explanation what this exactly means? In case of vpn it returns other? Only in a simulator, or on any device?

mit-73: I don't quite understand how to describe it correctly. Yes in case of vpn it returns other, on any device.

[Documentation describing type other](https://developer.apple.com/documentation/network/nwinterface/interfacetype/other).

mit-73: Would that be more correct?

`Note for iOS and macOS: in the case of [vpn] it returns [other], both in the simulator and on any device.`

mit-73:
@nohli I have corrected all your comments.

One test went down with this problem:`Error: Timeout waiting for emulator to boot.`

packages/connectivity_plus/connectivity_plus_platform_interface/lib/src/enums.dart
```diff
@@ -16,5 +16,13 @@ enum ConnectivityResult {
   none,
 
   /// VPN: Device connected to a VPN
-  vpn
+  ///
+  /// Note for iOS and macOS:
+  /// There is no separate network interface type for [vpn],
+  /// it returns [other] type.
+  /// Both in the simulator and on any device.
```
nohli, line 23:
```suggestion
  /// There is no separate network interface type for [vpn].
  /// It returns [other] on any device (also simulator).
```

nohli (approved):
Nice work, thanks for your contribution.
I've added one suggestion for improving the docs.

crespinb:
Hi, when can we expect this to be merged?

vbuberen (approved):

PR #1338, opened 2022-11-17
fix(connectivity_plus): Does not crash in debug mode of iOS 14+

tianxiawoyougood:
## Description

In ios 14+, debug mode Flutter apps can only be launched from Flutter tooling, IDEs with Flutter plugins or from Xcode.
So, registrar will be nil.
However, we can't make it crash in its own plugin.
Determine whether the registrar is null.

## Checklist

## Breaking Change

### DIFF
--- a/packages/connectivity_plus/connectivity_plus/ios/Classes/ConnectivityPlusPlugin.m
+++ b/packages/connectivity_plus/connectivity_plus/ios/Classes/ConnectivityPlusPlugin.m
@@ -10,6 +10,8 @@
 
 @implementation ConnectivityPlusPlugin
 + (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar> *)registrar {
-  [SwiftConnectivityPlusPlugin registerWithRegistrar:registrar];
+    if (registrar) {
+        [SwiftConnectivityPlusPlugin registerWithRegistrar:registrar];
+    }
 }
 @end
--- a/packages/connectivity_plus/connectivity_plus/ios/Classes/ConnectivityPlusPlugin.m
+++ b/packages/connectivity_plus/connectivity_plus/ios/Classes/ConnectivityPlusPlugin.m
@@ -10,8 +10,8 @@
 
 @implementation ConnectivityPlusPlugin
 + (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar> *)registrar {
-    if (registrar) {
-        [SwiftConnectivityPlusPlugin registerWithRegistrar:registrar];
-    }
+  if (registrar) {
+      [SwiftConnectivityPlusPlugin registerWithRegistrar:registrar];
+  }
 }
 @end
--- a/packages/connectivity_plus/connectivity_plus/ios/Classes/ConnectivityPlusPlugin.m
+++ b/packages/connectivity_plus/connectivity_plus/ios/Classes/ConnectivityPlusPlugin.m
@@ -11,7 +11,7 @@
 @implementation ConnectivityPlusPlugin
 + (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar> *)registrar {
   if (registrar) {
-      [SwiftConnectivityPlusPlugin registerWithRegistrar:registrar];
+    [SwiftConnectivityPlusPlugin registerWithRegistrar:registrar];
   }
 }
 @end


packages/connectivity_plus/connectivity_plus/ios/Classes/ConnectivityPlusPlugin.m
```diff
@@ -10,6 +10,8 @@
 
 @implementation ConnectivityPlusPlugin
 + (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar> *)registrar {
-  [SwiftConnectivityPlusPlugin registerWithRegistrar:registrar];
+    if (registrar) {
+        [SwiftConnectivityPlusPlugin registerWithRegistrar:registrar];
+    }
```
nohli, line 15:
```suggestion
  if (registrar) {
      [SwiftConnectivityPlusPlugin registerWithRegistrar:registrar];
  }
```

packages/connectivity_plus/connectivity_plus/ios/Classes/ConnectivityPlusPlugin.m
```diff
@@ -10,6 +10,8 @@
 
 @implementation ConnectivityPlusPlugin
 + (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar> *)registrar {
-  [SwiftConnectivityPlusPlugin registerWithRegistrar:registrar];
+  if (registrar) {
+      [SwiftConnectivityPlusPlugin registerWithRegistrar:registrar];
+  }
```
nohli, line 15:
```suggestion
  if (registrar) {
    [SwiftConnectivityPlusPlugin registerWithRegistrar:registrar];
  }
```

miquelbeltran:
Is there a valid case when the `registrar` should be null? I'd rather have it crash than fail silently.

tianxiawoyougood:
> `registrar`当应该为空时是否存在有效情况？我宁愿让它崩溃也不愿默默地失败。

在iOS 14+，当debug模式不是通过IDE运行的时候，一般会出现下面的页面，而不是崩溃。
A650EE9B-95FA-40AB-B70C-5F806C911C27

miquelbeltran:
I mean, if you are launching the app in debug mode from the launcher, but it is not supported by the OS, then it is not supported. Failing silently, it is not the right way to go here.

tianxiawoyougood:
> I mean, if you are launching the app in debug mode from the launcher, but it is not supported by the OS, then it is not supported. Failing silently, it is not the right way to go here.

Two aspects:
1. From a code point of view, registrar is nullable in Objective. The Swift method receives a null-safe parameter. You need to judge whether it is empty before assigning a value.
2. In the official Flutter SDK, a prompt is made instead of letting it crash.

miquelbeltran:
I am not convinced about approving this one, I'd rather see the plugin migrated entirely to Swift if that is possible instead of having the objc -> swift call, but just because the objc method is passing a nullable value doesn't mean we should accept that as a valid option. Is there any official Flutter documention explaining how this situation should be handled?

tianxiawoyougood:
> 我不相信批准这个，如果可能的话，我宁愿看到插件完全迁移到 Swift 而不是调用 objc -> swift，但仅仅因为 objc 方法传递一个可为 null 的值并不意味着我们应该接受它作为一个有效的选择。是否有任何官方 Flutter 文档解释应如何处理这种情况？

I think this has to be dealt with. Should this crash happen? do you have any other solution?

nohli:
The Objective-C part of this plugin are only 19 LoC.
Does migrating it to Swift also solve the issue?

miquelbeltran:
Just to give you an idea, in path_provider they don't check for `register` nullability either: 

https://github.com/flutter/plugins/blob/main/packages/path_provider/path_provider_ios/ios/Classes/FLTPathProviderPlugin.m#L20

tianxiawoyougood:
> The Objective-C part of this plugin are only 19 LoC. Does migrating it to Swift also solve the issue?

NO. Migrating it to Swift doesn't solve the problem

tianxiawoyougood:
The creation of messageChannel requires a nonnull registrar in swift.

tianxiawoyougood:
image

miquelbeltran:
See the linked code from the path_provider, they don't check if the `registar` is null either.

If the Framework is passing in a null `registar`, that's a bug in the framework, not in our code.

Sorry, I don't think we should accept this one. I'd like to hear what other maintainers think as well.

tianxiawoyougood:
> See the linked code from the path_provider, they don't check if the `registar` is null either.
> 
> If the Framework is passing in a null `registar`, that's a bug in the framework, not in our code.
> 
> Sorry, I don't think we should accept this one. I'd like to hear what other maintainers think as well.

I don't think you understand the question.
It doesn't crash without any processing in OC code. Many pure OC plug-ins are not handled like path_provider, and there is no problem.
The reason for the crash is that a nil is passed to the method of the Swift code.What this Swift method expects is a null safe parameter.
Therefore, make a judgment call when passing to Swift. 
It's that simple.

I think I've explained it clearly.

miquelbeltran:
The implementation of path_provider is accessing to `registar.messenger` without checking if `registar` is null or not.

If `registar` is null in path_provider, that would simply crash.

If the framework is providing a null `registar` that's a bug on the framework.

tianxiawoyougood:
> The implementation of path_provider is accessing to `registar.messenger` without checking if `registar` is null or not.
> 
> If `registar` is null in path_provider, that would simply crash.
> 
> If the framework is providing a null `registar` that's a bug on the framework.

No crash if registar in path_provider is null. Can you try running the code? Don't guess.

miquelbeltran:
Can you please indicate in the documentation where it says that the registar can be null? https://api.flutter.dev/objcdoc/Protocols/FlutterPlugin.html#/c:objc(pl)FlutterPlugin(cm)registerWithRegistrar:

tianxiawoyougood:
On iOS 14+, when the debug mode is not running through the IDE. 
Don't always trust the documentation.
Flutter of course hopes not to return null, but Apple does not allow it.

tianxiawoyougood:
> 

On iOS 14+, when the debug mode is not running through the IDE.
You can run the program yourself according to the information I provided to reproduce the scene.

miquelbeltran:
> On iOS 14+, when the debug mode is not running through the IDE.
> Flutter of course hopes not to return null, but Apple does not allow it.

Yes, I understand that, but that's not a problem of the plugin, that's a problem of the Flutter framework.

If the Flutter framework is passing a null registar when it explicitly says in the documentation it is nonnull, we should instead report that as a bug to the Flutter team.

If we fail silently instead, we are hiding an error that otherwise should have been reported.

> I'm going to sleep. Have you read Pride and Prejudice? Recommended for you.

I am honestly not sure what you mean, but I guess this ends our conversation here.

PR #1340, opened 2022-11-17
fix(share_plus): Fix the error of requestCode value range.

ipcjs:
## Description

When the application's activity is `FragmentActivity`, requestCode can only use the lower 16 bits.

## Related Issues

- *Related https://github.com/juliansteenbakker/mobile_scanner/pull/279*

## Checklist

## Breaking Change

### DIFF
--- a/packages/share_plus/share_plus/example/android/app/src/main/AndroidManifest.xml
+++ b/packages/share_plus/share_plus/example/android/app/src/main/AndroidManifest.xml
@@ -8,7 +8,7 @@
         android:icon="@mipmap/ic_launcher"
         android:label="share_example">
         <activity
-            android:name="io.flutter.embedding.android.FlutterActivity"
+            android:name="io.flutter.embedding.android.FlutterFragmentActivity"
             android:configChanges="orientation|keyboardHidden|keyboard|screenSize|locale|layoutDirection"
             android:exported="true"
             android:hardwareAccelerated="true"
--- a/packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/ShareSuccessManager.kt
+++ b/packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/ShareSuccessManager.kt
@@ -69,7 +69,11 @@ internal class ShareSuccessManager(private val context: Context) : ActivityResul
      * the share result.
      */
     companion object {
-        const val ACTIVITY_CODE = 17062003
+        /**
+         * When the application's activity is [androidx.fragment.app.FragmentActivity], requestCode can only use the lower 16 bits.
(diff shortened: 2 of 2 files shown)


packages/share_plus/share_plus/example/android/app/src/main/AndroidManifest.xml
```diff
@@ -8,7 +8,7 @@
         android:icon="@mipmap/ic_launcher"
         android:label="share_example">
         <activity
-            android:name="io.flutter.embedding.android.FlutterActivity"
+            android:name="io.flutter.embedding.android.FlutterFragmentActivity"
```
miquelbeltran, line 11: Is there any reason why the example should be changed?

miquelbeltran:
According to https://developer.android.com/reference/androidx/fragment/app/FragmentActivity#validateRequestPermissionsRequestCode(int) that should be deprecated and "there are no longer any restrictions on permissions requestCodes.". 

Since this is an `androidx` class, and not part of the OS, I assume that using an up-to-date package should be enough?

ipcjs:
`io.flutter:flutter_embedding_release` depends on `androidx.fragment:fragment:1.1.0`, which does not support the new `ActivityResultRegistry` API

https://github.com/flutter/engine/blob/main/tools/androidx/files.json#L31

packages/share_plus/share_plus/example/android/app/src/main/AndroidManifest.xml
ipcjs: In order to test that the `requestCode` is in the correct range.

miquelbeltran:
I always find confusing to see when a method was deprecated in the Android ecosystem, like I cannot find anywhere when was `validateRequestPermissionsRequestCode` deprecated. Anyway, I am fine with approving this change. I'd just ask that we don't change the example implementation if possible.

ipcjs:
Ok, I reverted it.

ipcjs:
This API was deprecated in [1.3.0-alpha05](https://developer.android.com/jetpack/androidx/releases/fragment?hl=en#version_120_3:~:text=thus%20removing%20the%20restriction%20on%20using%20only%20the%20lower%20bits).

miquelbeltran:
> This API was deprecated in [1.3.0-alpha05](https://developer.android.com/jetpack/androidx/releases/fragment?hl=en#version_120_3:~:text=thus%20removing%20the%20restriction%20on%20using%20only%20the%20lower%20bits).

OMG, released in 2020, and Flutter still using 1.1.0.

Thanks for pointing this out, btw.

miquelbeltran (approved):

diogo10:
Hi guys, can you confirm that this fix will be included in the next release? Can I ask you when the next release will come? Thanks for this lib!

PR #1373, opened 2022-12-04
fix: Data type issue - type 'String' is not a subtype of type 'bool'

vvvirani:
type 'String' is not a subtype of type 'bool'

Screenshot 2022-12-04 at 9 20 45 PM

Changes:
packages/device_info_plus/device_info_plus/lib/src/model/android_device_info.dart

Line number 143

Screenshot 2022-12-04 at 9 29 29 PM

### DIFF
--- a/packages/device_info_plus/device_info_plus/lib/src/model/android_device_info.dart
+++ b/packages/device_info_plus/device_info_plus/lib/src/model/android_device_info.dart
@@ -2,7 +2,6 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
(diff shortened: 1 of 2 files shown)


vbuberen:
How have you got this exception? Plugin puts boolean into that parameter, not string. Otherwise there would be exceptions for users in production and we would see reports from people about this issue already.

The suggested comparison is also wrong because bool won't be equal to string. Closing as invalid.

So I assume it is something with analyzer on your side.

vvvirani:
Hello vbuberen

image

You can see `init()` in the screenshot. I have just call `init()` function and catch error

PR #1389, opened 2022-12-10
docs(all): Updates for READMEs and website pages

vbuberen:
## Description

Updating docs to cleanup them a little bit and add missing info on either website or in README files (for example, added info from #389 about package_info_plus).

Also, started to use fresh Github features for admonitions in README files for better visualisation:
Screenshot 2022-12-11 at 00 43 17

Will do some more docs updates in further PRs.

## Related Issues

Closes #389 

## Checklist

- [x] I read the Contributor Guide and followed the process outlined there for submitting PRs.
- [x] I titled the PR using [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0).
- [x] I did not modify the `CHANGELOG.md` nor the `pubspec.yaml` files.
- [x] All existing and new tests are passing.
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate that with a `!` in the title as explained in [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0)).
- [x] No, this is *not* a breaking change.

PR #1391, opened 2022-12-11
chore(deps): Update Docusaurus dependncies

vbuberen:
## Description

Was looking through logs of preview build and saw that there are some errors due to Docusaurus components having different versions:
Screenshot 2022-12-11 at 15 00 48

This PR addresses these issues.

## Related Issues

-

## Checklist

## Breaking Change

### DIFF
--- a/.github/workflows/build_preview_website.yaml
+++ b/.github/workflows/build_preview_website.yaml
@@ -23,7 +23,7 @@ jobs:
       - name: Checkout Repo
         uses: actions/checkout@v3
         with:
-          ref: refs/pull/${{ github.event.number }}/merge
+          ref: ${{ github.ref }}
       - name: Install Dependencies
         run: npm ci
       - name: Build website
@@ -43,7 +43,7 @@ jobs:
       - name: Checkout Repo
         uses: actions/checkout@v3
         with:
-          ref: refs/pull/${{ github.event.number }}/merge
+          ref: ${{ github.ref }}
       - name: Download Artifact
         uses: actions/download-artifact@v3
         with:
--- a/.github/workflows/build_preview_website.yaml
+++ b/.github/workflows/build_preview_website.yaml
@@ -23,7 +23,7 @@ jobs:
       - name: Checkout Repo
         uses: actions/checkout@v3
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/build_preview_website.yaml
```diff
@@ -22,6 +22,8 @@ jobs:
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
```
nohli, line 26: What is this for?

vbuberen: I was researching the reason for the workflows not building the latest website versions like here: https://github.com/fluttercommunity/plus_plugins/pull/1389#issuecomment-1345403167

Same thing happened in this PR as well and, I guess, in any other PR we had so far for this workflow (potentially, for all other workflows as well). To see that not the latest version is built see logs for the `build website preview` workflow:
Screenshot 2022-12-12 at 12 08 13

As you might see the website version is still `0.2.1` even though I bumped it in this PR as well as older Docusaurus dependencies are used.

While looking through logs I found out that the issue is in checkout action, which checks out latest commit from the `main` branch, not the latest commit from the PR. 

As it turned out it happens for workflows that get restarted via `Re-run jobs`, for example: https://github.com/actions/checkout/issues/1036

Thus, I am experimenting with workflow updates to see what will fix the issue for website preview builds for us. Additionally, as we use checkout action everywhere all other workflows are affect by the same issue as well. So as soon as I find out the solution I will create another PR to fix the issue.

nohli: Cool, thanks for the explanation ❤️

vbuberen:
Seems like the fix for the checkout should work after I merge this PR. So going to merge this and check how it changes for other PRs.

.github/workflows/build_preview_website.yaml
vbuberen: As it turned out the issue with `checkout` happens only when workflow is triggered by event like `pull_request_target` which we use for this preview deployment workflow only. Thus, my theory about other workflows being affected as well isn't correct.

PR #1393, opened 2022-12-12
fix(connectivity_plus): add onCapabilitiesChanged callback

NANASHI0X74:
## Description

The android implementation for the onconnectivitychanged stream uses a network callback to emit new values every time onAvailable is Called. From Android M on, it determines which type of network is active by calling `connectivityManager.getNetworkCapabilities(network)`, from `connectivity.getNetworkType` but in the [documentation for the onAvailable callback](https://developer.android.com/reference/android/net/ConnectivityManager.NetworkCallback#onAvailable(android.net.Network)), It says:
```
Do NOT call ConnectivityManager.getNetworkCapabilities(android.net.Network) or ConnectivityManager.getLinkProperties(android.net.Network) or other synchronous ConnectivityManager methods in this callback as this is prone to race conditions (there is no guarantee the objects returned by these methods will be current).
Instead, wait for a call to onCapabilitiesChanged(android.net.Network, android.net.NetworkCapabilities) and onLinkPropertiesChanged(android.net.Network, android.net.LinkProperties) whose arguments are guaranteed to be well-ordered with respect to other callbacks.
```

So, it seems getNetworkType should be called from onCapabilitiesChanged instead of from onAvailable. Since I'm not sure how that would affect Platforms prior to Android M, I left the onAvailable Callback in.

The motivation for this is that we have discovered a bug where sometimes after our phone has been in standby for some time, onConnectivityChanged wouldn't update correctly, still reporting no connectivity even though a wifi connection was active and we're hoping this might fix that.

## Checklist

## Breaking Change

### DIFF
--- a/packages/connectivity_plus/connectivity_plus/android/src/main/java/dev/fluttercommunity/plus/connectivity/ConnectivityBroadcastReceiver.java
+++ b/packages/connectivity_plus/connectivity_plus/android/src/main/java/dev/fluttercommunity/plus/connectivity/ConnectivityBroadcastReceiver.java
@@ -10,6 +10,7 @@ import android.content.Intent;
 import android.content.IntentFilter;
 import android.net.ConnectivityManager;
 import android.net.Network;
+import android.net.NetworkCapabilities;
 import android.os.Build;
 import android.os.Handler;
 import android.os.Looper;
@@ -45,6 +46,12 @@ public class ConnectivityBroadcastReceiver extends BroadcastReceiver
           new ConnectivityManager.NetworkCallback() {
             @Override
             public void onAvailable(Network network) {
+              sendEvent();
+            }
+
+            @Override
+            public void onCapabilitiesChanged(
+                Network network, NetworkCapabilities networkCapabilities) {
               sendEvent();
             }
 


NANASHI0X74:
sorry, I accidentally submitted the PR early, I'm still editing...

NANASHI0X74:
an ios integration test failed, but since I only touched android source files this can't have been my doing 🤔

vbuberen:
> The motivation for this is that we have discovered a bug where sometimes after our phone has been in standby for some time, onConnectivityChanged wouldn't update correctly, still reporting no connectivity even though a wifi connection was active and we're hoping this might fix that.

It happens in lots of Android native apps as well. I have it on one banking app I use on a daily basis and some of my ex-colleagues from one of previous job having the same issue in their Android native projects.

> Since I'm not sure how that would affect Platforms prior to Android M, I left the onAvailable Callback in.

Do you have a confirmation or some proofs that this change you suggest fixes the issue?

NANASHI0X74:
Our QA engineer said it happens every day over night, so I'm going to build our app with this patched version of the package and see if it changes. But either way, the way this is implemented violates the guidance from the android documentation, so maybe it should be updated regardless

vbuberen:
> I'm going to build our app with this patched version of the package and see if it changes tomorrow

Looking forward to see results.

> the way this is implemented violates the guidance from the android documentation, so maybe it should be updated regardless

I would say that `violates` is too loud, but agree that if the suggested by the documentation way works fine on all APIs we should update the plugin to use it 👍🏻 
I will test this change as well in nearest days.

NANASHI0X74:
Unfortunately this didn't fix the bug

vbuberen:
> Unfortunately this didn't fix the bug

Thanks for sharing updates. As I mentioned in one of my replies I see the same issue in some native apps from the Play Store as well, so, I guess, it is quite a popular one with no obvious fix.

Also, as to the suggested change - it turned out that `onCapabilitiesChanged()` is not different from the already used `onAvailable()` according to [the official docs](https://developer.android.com/reference/android/net/ConnectivityManager.NetworkCallback#onCapabilitiesChanged(android.net.Network,%20android.net.NetworkCapabilities)) Thus, this change doesn't change anything as it seems to me, so I would like to close this PR.

NANASHI0X74:
What I understood from the docs is that you shouldn't call getNetworkCapabilities from onAvailable because the returned object may not be up to date at that point. I was assuming calling getNetworkCapabilities is fine from oncapabilitieschanged because the capabilities must have been updated at that point. But it's true it doesn't explicitly state that there

PR #1425, opened 2022-12-29
docs(sensor_plus): improve description of accelerometer

cristian-barbarosie:
## Description

I believe the description of the accelerometer in docs/sensors_plus/usage.mdx is not accurate.

I changed the description of AccelerometerEvent and UserAccelerometerEvent. See https://en.wikipedia.org/wiki/Accelerometer

The text is now rather long; I was unable to compress the information I find relevant in a shorter text.

I changed the order of the topics because I feel the presentation is more clear if we explain first the concept of acceleration in itself, and only after that we explain that the gravity is indistinguishable from an accelerated movement. I also changed the order in packages/sensors_plus/sensors_plus/example/lib/main.dart (perhaps this wasn't necessary).

I wonder if the descriptions of GyroscopeEvent and MagnetometerEvent are accurate and complete. I think at least the physical measure units should be mentioned. I suspect the GyroscopeEvent describes the rate of change of the rotation (angular velocity) rather than the rotation of the device. Or perhaps the angular acceleration ? Since I don't have enough knowledge on these topics, I left them unchanged.

## Related Issues

Discussion #1419

## Checklist

## Breaking Change

### DIFF
--- a/docs/sensors_plus/usage.mdx
+++ b/docs/sensors_plus/usage.mdx
@@ -12,23 +12,22 @@ file](https://plus.fluttercommunity.dev/docs/overview).
 This will expose such classes of sensor events through a set of streams:
 
 - `UserAccelerometerEvent` describes the acceleration of the device, in m/s<sup>2</sup>.
-If the device is still, or is moving along a straight line at constant speed,
-the reported acceleration is zero.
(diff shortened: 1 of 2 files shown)


nohli (approved):
LGTM!
Thanks for your contribution! ❤️

PR #1477, opened 2023-01-27
docs: Fixed markdown table formatting in README

Jebiel:
Table wasn't rendered properly, since the row with line separators were on the same line as the row of titles.


vbuberen (approved):
Thanks for catching and fixing it.

PR #1560, opened 2023-02-21
feat(android_intent_plus): Add support for mixing ArrayLists and Arrays and add support for nested bundles.

jsroest:
The current `arguments` parameter and `arrayArguments` parameter only support flat arrays. On top of that, `arguments` only support lists of type `ArrayList`, and `arrayArguments` only support lists of type `Array`.

This PR adds support for bundles that can contain lists of both types and also adds support for nested bundles.

I need this behavior to support the complex bundle structure as described in this document: [Nested bundles](https://techdocs.zebra.com/datawedge/11-4/guide/api/setconfig/#nestedbundles).

Sample usage:
```dart
final intent = AndroidIntent(
  action: 'com.symbol.datawedge.api.ACTION',
  extras: [
    Bundle(
      value: [
        PutBundle(
          key: 'com.symbol.datawedge.api.SET_CONFIG',
          value: [
            PutString(
              key: 'PROFILE_NAME',
              value: 'com.dalosy.scanner_by_intents',
            ),
            PutString(
              key: 'PROFILE_ENABLED',
              value: 'true',
            ),
            PutString(
              key: 'CONFIG_MODE',
              value: 'CREATE_IF_NOT_EXIST',
            ),
            PutParcelableArray(
              key: 'APP_LIST',
              value: [
                appList(),
              ],
            ),
            PutParcelableArrayList(
              key: 'PLUGIN_CONFIG',
              value: [
                barcodePluginConfig(),
                intentPluginConfig(),
                keystrokePluginConfig(),
              ],
            ),
          ],
        ),
      ],
    ),
  ],
);
intent.sendBroadcast();
}
```

## Checklist

## Breaking Change


packages/android_intent_plus/example/lib/main.dart
```diff
@@ -109,6 +115,141 @@ class MyHomePage extends StatelessWidget {
...
+                  value: [
+                    appList(),
+                  ],
+                ),
+                PutParcelableArrayList(
+                  key: 'PLUGIN_CONFIG',
+                  value: [
+                    barcodePluginConfig(),
+                    intentPluginConfig(),
+                    keystrokePluginConfig(),
```
nohli, line 151: Is there a benefit of using methods over objects (also `appList`)?

nohli:
@jsroest thanks a lot for your contribution. I love the updated tests and example!

The `onReceive` callback is not triggered on my emulator. Is this a bug?
https://github.com/fluttercommunity/plus_plugins/assets/43643339/0cacb0c0-8b36-4c99-9be8-8a1120e111be

Sorry for the delay in reviewing this PR. I'm curious if you're already using this in an app and whether it works as intended there. Could you share your use-case? I'm interested in understanding where this feature could be particularly beneficial.

packages/android_intent_plus/example/lib/main.dart
jsroest: Hi, 
The sample code is purely based on the use case described here:
https://techdocs.zebra.com/datawedge/11-4/guide/api/setconfig/#nestedbundles.

The nesting goes a whole lot deeper than is visualized in my example. The functions represent a seperate part of the data structure and are only there to improve the readability and seperate concerns.

nohli: Makes sense.

I meant, instead of this:

```
  Bundle appList() => Bundle(
        value: [
          PutString(
            key: 'PACKAGE_NAME',
            value: 'com.dalosy.scanner_by_intents',
          ),
          PutStringArray(
            key: 'ACTIVITY_LIST',
            value: ['*'],
          ),
        ],
      );
```

like this:
```
  final appList = Bundle(
    value: [
      PutString(
        key: 'PACKAGE_NAME',
        value: 'com.dalosy.scanner_by_intents',
      ),
      PutStringArray(
        key: 'ACTIVITY_LIST',
        value: ['*'],
      ),
    ],
  );
```

jsroest: Hi,
Ah, right. No, I don't think there is any benefit in using one or the other. (Is there?)

nohli: The function creates a new object every time it's called, an object would be reused.

Since we only need to create the object once and don't need to change it (no parameters), declaring a variable could be simpler and more efficient.

Another suggestion: They could be private.

jsroest:
> @jsroest thanks a lot for your contribution. I love the updated tests and example!
> 
> The `onReceive` callback is not triggered on my emulator. Is this a bug? https://github.com/fluttercommunity/plus_plugins/assets/43643339/0cacb0c0-8b36-4c99-9be8-8a1120e111be
> 
> Sorry for the delay in reviewing this PR. I'm curious if you're already using this in an app and whether it works as intended there. Could you share your use-case? I'm interested in understanding where this feature could be particularly beneficial.

The `onReceive` callback is not triggered because it was not intended to do so. The code creates a profile in the datawedge environment on Zebra devices. That's how I tested if a complex bundle structure worked in real life on an actual device. The sample code does nothing on non-zebra devices as the broadcast is not listened to.

Because this PR did not seem to get any attention, I moved on to make this change available on a broader scale. Now I'm working on a broadcast **sender and receiver** plugin that works with the same bundle classes and makes it available to the consumer of the plugin from only the Dart side. But it's a lot of work and I don't know when/if it will be ready.

I believe that this change is still beneficial for the community as it can replace both the `arguments` and `arrayArguments` parameters and provide a universal approach.

packages/android_intent_plus/example/lib/main.dart
jsroest: Of course, you are right. This is part of the sample code and I did not think about optimizing. It just came out of my fingers that way.

nohli:
Ok I hope you're successful with your project! In the meantime, you wish to proceed with this PR, right?

Is there any way to test it on a device/emulator?

jsroest:
> Ok I hope you're successful with your project! In the meantime, you wish to proceed with this PR, right?
> 
> Is there any way to test it on a device/emulator?

No, the current test for the `extras` parameters does nothing on a regular (non Zebra brand) device or an emulator.
Maybe we should rewrite the tests, so it shows that it can do the same as the `arrays` and `arrayArguments` samples?

nohli:
The German train company uses Zebra devices 😁

image

PR #1593, opened 2023-03-04
chore(all): Refresh Dependabot config

vbuberen:
## Description

I see that for some reason Dependabot ignores some of packages we have in this repo, so, for example, `share_plus` uses not the most up-to-date dependencies. After some research I found out that it might be a bug in Dependabot: https://github.com/dependabot/dependabot-core/issues/2574

As suggested in that issue any commit to yaml config will make Dependabot re-evaluate it and, hopefully, will resume PRs creation. So trying my luck here.

In case Dependabot continues ignoring packages I would try to swap it to https://github.com/renovatebot/renovate as an experiment.

## Checklist

- [x] I read the [Contributor Guide](https://github.com/fluttercommunity/plus_plugins/blob/main/CONTRIBUTING.md) and followed the process outlined there for submitting PRs.
- [x] I titled the PR using [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0).
- [x] I did not modify the `CHANGELOG.md` nor the `pubspec.yaml` files.
- [x] All existing and new tests are passing.
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate that with a `!` in the title as explained in [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0)).
- [x] No, this is *not* a breaking change.

PR #1596, opened 2023-03-04
chore(share_plus)!: Update multiple dependencies, bump min Flutter to 3.0

vbuberen:
## Description

Updating a set of dependencies for `share_plus` as it seems that Dependabot didn't really bump dependencies and was fine with constraints we had specified for the plugin. 
We also have a bit lower score on pub.dev due to not having latest versions: https://pub.dev/packages/share_plus/score
Unfortunately, I couldn't bump `win32` to 4.x versions as `path_provider` used in platform interface still uses old `win32` version.

This PR is a breaking change because a few dependencies, like `cross_file` bumped their min Flutter version to 3.0: https://pub.dev/packages/cross_file/changelog
As it wasn't only one dependency that did so, I decided to update our min Flutter version as well.

## Checklist

## Breaking Change


jpnurmi:
I wonder if it would be enough to bump the Flutter version and let dependabot take care of the rest. :thinking:

vbuberen:
> I wonder if it would be enough to bump the Flutter version and let dependabot take care of the rest. 🤔

I think Dependabot will do nothing after reading about some issues today: https://github.com/fluttercommunity/plus_plugins/pull/1593#issuecomment-1454704145

jpnurmi:
> Unfortunately, I couldn't bump `win32` to 4.x versions as path_provider used in platform interface still uses old `win32` version.

I have a feeling that this is the only version constraint that causes pub.dev to reduce points. This is the only constraint that creates a conflicts if apps try use the latest available package version together with plus plugins. All other version constraints are compatible with the latest package versions.

vbuberen:
Alright, I think that since we can't update `win32` and `meta` anyway we won't get score bump, so pushing min version of Flutter might too much at this point. Thus, closing the PR till bumping min Flutter version is justified (for some bug fix or when we can do `win32` update).

PR #1613, opened 2023-03-11
fix(share_plus): Set exported=false for BroadcastReceiver on Android

### LINKED ISSUE #1608: [Vulnerability]: Security scanner find vulnerability in broadcast receiver for share plus plugin
### Platform

Android

### Plugin

Share plus

### Version

6.0.1

### Flutter SDK

3.7.5

### Steps to reproduce

Just install plugin, it will get report from scanner about vulnerability. 

Scanner reports vulnerability:

Findings: `Broadcast Receiver (dev.fluttercommunity.plus.share.SharePlusPendingIntent) is not Protected.`
Severity: high
Description: 

```
[android:exported=true] A Broadcast Receiver is found to be
 shared with other apps on the device therefore leaving it 
accessible to any other application on the device.
```

### Code Sample

_No response_

### Logs

```shell
not needed for this ticket
```

### Flutter Doctor

```shell
[✓] Flutter (Channel stable, 3.7.5, on macOS 13.2.1 22D68 darwin-arm64, locale en-GB)
    • Flutter version 3.7.5 on channel stable 
    • Framework revision c07f788888 (2 weeks ago), 2023-02-22 17:52:33 -0600
    • Engine revision 0f359063c4
    • Dart version 2.19.2
    • DevTools version 2.20.1

[✓] Android toolchain - develop for Android devices (Android SDK version 31.0.0)
    • Platform android-33, build-tools 31.0.0
    • Java binary at: /Applications/Android Studio.app/Contents/jbr/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 11.0.15+0-b2043.56-8887301)
```
```


- [X] I did a `flutter clean`

### PR DESCRIPTION
**Pull request by vbuberen**
## Description

Addressing an issue reported in #1608. After looking at the documentation and code saw that we indeed don't need to have receiver exposed as only catch the intent that we launch from the app that uses `share_plus`

Docs on topic: 
- https://developer.android.com/guide/topics/manifest/receiver-element
- https://developer.android.com/training/permissions/restrict-interactions#broadcast-receivers

Tested the change - no issues with getting result of share in example app.

## Related Issues

Closes #1608

## Checklist

## Breaking Change

### DIFF
--- a/packages/share_plus/share_plus/android/src/main/AndroidManifest.xml
+++ b/packages/share_plus/share_plus/android/src/main/AndroidManifest.xml
@@ -14,7 +14,7 @@
       </provider>
       <!-- This manifest declared broadcast receiver allows us to use an explicit
            Intent when creating a PendingItent to be informed of the user's choice -->
-      <receiver android:name=".SharePlusPendingIntent" android:exported="true">
+      <receiver android:name=".SharePlusPendingIntent" android:exported="false">
         <intent-filter>
           <action android:name="EXTRA_CHOSEN_COMPONENT" />
         </intent-filter>

### CODE REVIEW

File: packages/share_plus/share_plus/android/src/main/AndroidManifest.xml [hunk 1]
```diff
@@ -14,7 +14,7 @@
       </provider>
       <!-- This manifest declared broadcast receiver allows us to use an explicit
            Intent when creating a PendingItent to be informed of the user's choice -->
-      <receiver android:name=".SharePlusPendingIntent" android:exported="true">
+      <receiver android:name=".SharePlusPendingIntent" android:exported="false">
```
charafau, line 17: Do we know why it was exported ? Was it on purpose or just a miss / was created exported by default?

vbuberen: Is it that big deal? You can ping author and ask. I guess, it it because the author isn't a mobile dev, but I don't see why we should care about it, considering that the issue is resolved.

nohli (approved):
Maybe it was added during the Android v2 migration.
Also, it's true by default.

Are we sure users won't run into issues like [this](https://stackoverflow.com/questions/71438370/flutter-android-you-uploaded-an-apk-which-has-activity-but-without-the-a)?

vbuberen:
> Maybe it was added during the Android v2 migration.
> Also, it's true by default.

It was added while adding share result functionality. As to default or not - I have attached the docs in the PR description about why it is Ok to set to `false`.

> Are we sure users won't run into issues like [this](https://stackoverflow.com/questions/71438370/flutter-android-you-uploaded-an-apk-which-has-activity-but-without-the-a)?

Yes, I am sure. The issue you linked is about not having `exported` property defined at all for one of components, not about `true` or `false`.

charafau:
thanks @vbuberen  for PR !

PR #1615, opened 2023-03-11
ci(all): Fix Melos version on 2.9.0

vbuberen:
## Description

CI workflow was activating latest version of Melos, so in #1614 and #1613 checks were already running with Melos 3.0.0, which was released just recently. Initially I wanted to follow the migration guide, but saw in changelogs that version 3.0.0 requires 2.18 as min Dart SDK: https://github.com/invertase/melos/blob/main/CHANGELOG.md#melos---v300
I wouldn't like us to bump plugins min Dart version for now, so modified the script to use 2.9.0 as it was used before.

## Checklist

## Breaking Change

### DIFF
--- a/.github/workflows/scripts/install-tools.sh
+++ b/.github/workflows/scripts/install-tools.sh
@@ -3,7 +3,7 @@
 set -e
 
 flutter config --no-analytics
-flutter pub global activate melos
+flutter pub global activate melos 2.9.0
 echo "$HOME/.pub-cache/bin" >> $GITHUB_PATH
 echo "$HOME/AppData/Local/Pub/Cache/bin" >> $GITHUB_PATH
 echo "$GITHUB_WORKSPACE/_flutter/.pub-cache/bin" >> $GITHUB_PATH


mhadaily (approved):
LGTM, however we just published Melos 3.0 which I think we can upgrade, that comes with lots of good new changes.

vbuberen:
> however we just published Melos 3.0 which I think we can upgrade

I would gladly use it, but as mentioned in the PR description don't want to bump min Dart to 2.18 as Melos 3.0.0 requires. After more dependencies we use start bumping min Dart to 2.18 and min Flutter to 3.0 we can get back to this. For now we don't really need any of new functionally that much to bump versions. Like 
I described here: https://github.com/fluttercommunity/plus_plugins/pull/1596#issuecomment-1455932342 

So would get back to this topic at the end of spring-beginning of summer.

nohli:
I think, increasing the minimum version for development tools is a very different use case than increases in the pubspec.

We're not limiting users and don't have breaking changes by increasing the minimum Flutter/Dart version for melos, right?

vbuberen:
> I think, increasing the minimum version for development tools is a very different use case than increases in the pubspec.

Didn't get what you mean here. We would need to bump min Dart version in plugins pubspec files to 2.18 if we want to switch to Melos 3.0.0. How it is not limiting users if we bump the requirements for plugins?

spydon:
> > I think, increasing the minimum version for development tools is a very different use case than increases in the pubspec.
> 
> Didn't get what you mean here. We would need to bump min Dart version in plugins pubspec files to 2.18 if we want to switch to Melos 3.0.0. How it is not limiting users if we bump the requirements for plugins?

You only have to bump it in the root workspace pubspec, not in the pubspec for the packages, like I did in #1619.

vbuberen:
Somehow I missed the part about workspace pubspec. We are good to go then.

PR #1619, opened 2023-03-13
ci(all): Migrate to Melos 3

spydon:
## Description

This migrates the repository and pipeline over to Melos 3.
The contributing guide doesn't have to be updated since that already installs the latest Melos version.

## Checklist

- [x] I read the [Contributor Guide](https://github.com/fluttercommunity/plus_plugins/blob/main/CONTRIBUTING.md) and followed the process outlined there for submitting PRs.
- [x] I titled the PR using [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0).
- [x] I did not modify the `CHANGELOG.md` nor the `pubspec.yaml` files. (Well, I did add the root pubspec file)
- [x] All existing and new tests are passing.
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [x] No, this is *not* a breaking change.

PR #1628, opened 2023-03-15
refactor(all): Remove all manual dependency_overrides

spydon:
## Description

Manual dependency_overrides in the examples shouldn't be needed since Melos already should handle those.

## Checklist

## Breaking Change

### DIFF
--- a/packages/android_alarm_manager_plus/example/pubspec.yaml
+++ b/packages/android_alarm_manager_plus/example/pubspec.yaml
@@ -8,8 +8,7 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
-  android_alarm_manager_plus:
-    path: ../
+  android_alarm_manager_plus: ^2.1.1
   shared_preferences: ^2.0.13
   path_provider: ^2.0.9
 
--- a/packages/android_intent_plus/example/pubspec.yaml
+++ b/packages/android_intent_plus/example/pubspec.yaml
@@ -8,8 +8,7 @@ dependencies:
   flutter:
     sdk: flutter
   platform: ^3.1.0
-  android_intent_plus:
-    path: ../
+  android_intent_plus: ^3.1.6
 
 dev_dependencies:
   flutter_driver:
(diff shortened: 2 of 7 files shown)


vbuberen:
Unfortunately it doesn't solve the problem I mentioned in https://github.com/fluttercommunity/plus_plugins/pull/1619#issuecomment-1470829311 because it is not about using overrides, but because `melos version` bumps version of `share_plus_platform_interface` and during `melos publish` dry run can't resolve the bumped version of platform interface for `share_plus`. 
There were a few mentions that the problem is about some changes in pub and it is not Melos related. 

Screenshot 2023-03-16 at 12 03 04

I have tried your PR with additional similar change in `share_plus` you did, but haven't seen any changes in behavior during melos bootstraping, etc. 
Additionally I would prefer to have dependencies from path in example apps as really use Melos only when need to do release, not during work on plugins.

spydon:
Makes sense that it didn't solve anything then.

> Additionally I would prefer to have dependencies from path in example apps as really use Melos only when need to do release, not during work on plugins.

You still have to run Melos bootstrap to work locally though, and then it would set up the examples to be dependent on the other packages in the repository?

vbuberen:
> You still have to run Melos bootstrap to work locally though, and then it would set up the examples to be dependent on the other packages in the repository?

No, I don't. I started to use Melos only recently and only for purpose of publishing. Before that just opened a project of a plugin in question and work on it.

spydon:
> No, I don't. I started to use Melos only recently and only for purpose of publishing. Before that just opened a project of a plugin in question and work on it.

Interesting, how do you work with when changing packages that are dependent on other packages in the repository, like the interfaces for example? Because if you haven't run melos bootstrap it would pull the interface package from pub instead of using the local one?

vbuberen:
> Interesting, how do you work with changing packages that are dependent on other packages in the repository, like the interfaces for example?

Probably, I haven't bumped into this case so far 🤷🏻

spydon:
@vbuberen should I close this or do you want to start using melos all the way? :)

vbuberen:
Well, let's merge it if Melos takes care of this stuff during versioning, but don't really see much benefits.

spydon:
@vbuberen something strange had happened when this branch was updated with main, but it's fixed now. The auto-complete was cancelled though.

vbuberen:
Yeah, had to manually restart some of our flaky android integration tests, but now everything seems fine.

vbuberen (approved):

PR #1649, opened 2023-03-28
feat(device_info_plus): add major, minor and patch versions to macos

goenning:
## Description

This PR adds major/minor/patch versions to mac info class, similar to what's available on windows.

With this fields we don't need to parse the version from of the osRelease field (which apple doesn't recommend).

https://developer.apple.com/documentation/foundation/operatingsystemversion

## Checklist

- [X] I did not modify the `CHANGELOG.md` nor the `pubspec.yaml` files.

## Breaking Change


goenning:
Android build is failing, but doesn't seem related?

vbuberen:
Thanks for your contribution.

I would also ask you to add these new fields to the `example` app, so they will be visible as well near that `osRelease` string.

> Android build is failing, but doesn't seem related?

Yes, correct. Android tests are flaky here. But seems like re-run helped.

goenning:
Done! :)
Screenshot 2023-04-04 at 22 05 00

nohli:
Should we also return a String with all, like version: 13.1.0?

goenning:
I thin that's something the user can easily concatenate on their side. It's the same with windows info, there are 3 separate fields.

packages/device_info_plus/device_info_plus/macos/Classes/DeviceInfoPlusMacosPlugin.swift
```diff
@@ -18,12 +18,17 @@ public class DeviceInfoPlusMacosPlugin: NSObject, FlutterPlugin {
     }
 
     private func handleDeviceInfo(result: @escaping FlutterResult)-> Void{
+        let osVersion = ProcessInfo.processInfo.operatingSystemVersion;
+
```
nohli, line 22: Is the whitespace needed?

```suggestion
        let osVersion = ProcessInfo.processInfo.operatingSystemVersion;
```

goenning: Removed, and also moved the osVersion closer to where it's used

nohli (approved):
LGTM!

goenning:
Thakns @nohli I couldn't find any info on the release process. How often does the package get released to pub.dev?

nohli:
Right now, there is a bug in our publishing tool, but vbueren will try to publish later today.

PR #1650, opened 2023-03-28
fix(share_plus_android): Fix strict compilation errors in MIME reduction function

ChristianEdwardPadilla:
## Description

Without this change I see the following errors with strict compilation flags:

```
Error: 'types' is captured by a lambda expression that is passed as immutable type(s) String but is of type List which is mutable [Immutable]
        if (types.size == 1) {
            ~~~~~
Error: 'reducedMimeType' is captured by a lambda expression that is passed as immutable type(s) String but is not final [Immutable]
          reducedMimeType = types.first()
          ~~~~~~~~~~~~~~~
```

## Checklist

## Breaking Change

### DIFF
--- a/packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/Share.kt
+++ b/packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/Share.kt
@@ -184,35 +184,27 @@ internal class Share(
         return uris
     }
 
-    /**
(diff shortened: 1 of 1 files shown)


vbuberen:
Thanks for your contribution. 
You have such errors only in `share_plus` because you use only this package? 
Also, do you mean [Android strict mode](https://developer.android.com/reference/android/os/StrictMode)?

vbuberen (approved):

PR #1711, opened 2023-04-24
update win32 dependencies

git-xiaocao:
Need to update dependency version due to win32 dependency in share_plus conflicts with file_picker.

`Because share_plus >=6.0.0 depends on win32 ^3.0.0 and file_picker >=5.2.11 depends on win32 ^4.1.3, share_plus >=6.0.0 is incompatible with file_picker >=5.2.11.`


vbuberen:
Use dependencies override as we can't update `win32` at the moment due to other packages constrains (see logs from CI checks).

PR #1731, opened 2023-04-28
chore(package_info_plus): Update Flutter dependencies, set Flutter >=3.3.0 and Dart to >=2.18.0 <4.0.0

vbuberen:
## Description

Updating used dependencies and aligning constraint with other plugins maintained by the Flutter team (for example, here: https://github.com/flutter/packages/blob/main/packages/url_launcher/url_launcher/pubspec.yaml#L8). Not marking as a breaking change as we already have namespace marked as breaking change.

## Checklist

## Breaking Change

### DIFF
--- a/packages/package_info_plus/package_info_plus/pubspec.yaml
+++ b/packages/package_info_plus/package_info_plus/pubspec.yaml
@@ -29,18 +29,18 @@ dependencies:
     sdk: flutter
   flutter_web_plugins:
     sdk: flutter
-  http: ^0.13.0
-  meta: ^1.3.0
-  path: ^1.8.0
+  http: ^0.13.5
+  meta: ^1.8.0
+  path: ^1.8.2
   package_info_plus_platform_interface: ^2.0.1
-  win32: ">=2.7.0 <5.0.0"
+  win32: ^4.1.3
 
 dev_dependencies:
   flutter_lints: ^2.0.1
   flutter_test:
     sdk: flutter
-  test: ^1.21.1
+  test: ^1.22.0
 
 environment:
-  sdk: ">=2.12.0 <3.0.0"
-  flutter: ">=2.11.0"
+  sdk: ">=2.18.0 <4.0.0"
+  flutter: ">=3.3.0"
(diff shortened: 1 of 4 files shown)


packages/package_info_plus/package_info_plus/pubspec.yaml
```diff
@@ -29,18 +29,18 @@ dependencies:
...
     sdk: flutter
-  http: ^0.13.0
-  meta: ^1.3.0
-  path: ^1.8.0
+  http: ^0.13.5
+  meta: ^1.8.0
+  path: ^1.8.2
   package_info_plus_platform_interface: ^2.0.1
-  win32: ">=2.7.0 <5.0.0"
+  win32: ^4.1.3
```
timsneath, line 36: As the author of `win32`, I'd recommend against setting such a wide set of constraints as was originally the case, but equally the replacement is unnecessarily narrow. Perhaps `^4.0.0` would be better?

packages/package_info_plus/package_info_plus/pubspec.yaml
```diff
@@ -29,18 +29,18 @@ dependencies:
     sdk: flutter
   flutter_web_plugins:
     sdk: flutter
-  http: ^0.13.0
```
timsneath, line 32: Is there a reason why you're tightening these three constraints to the latest version? This may cause problems for projects that contain other packages that have pinned a version for some reason. Unless there's a reason to migrate forward (e.g. you rely on a feature in meta v1.8), it's better to keep these looser.

packages/package_info_plus/package_info_plus/pubspec.yaml
timsneath: All the breaking changes in the 4.x series are to the experimental WinRT projection. The other changes are just adding new APIs, so ^4.0.0 should be great!

vbuberen: Alright, thanks for explanations.

packages/package_info_plus/package_info_plus/pubspec.yaml
vbuberen: The next release for Plus plugins will be a major bump because of some breaking changes (dropping support of Android Gradle Plugin < 4.2). Due to this I would like to bump dependencies versions as well to stimulate people move forward, considering that the release will be breaking anyway.

PR #1745, opened 2023-05-01
chore(device_info_plus): Remove redundant checks for PRODUCT strings with sdk

SamJakob:
There's already a check for PRODUCT contains "sdk", so there's no need to test various combinations of this. Perhaps it's also worth using case-insensitive tests and, for example, simply checking if MODEL then contains "sdk"?

## Description

There's already a check for PRODUCT contains "sdk", so there's no need to test various combinations of this. Also, added parenthesis around the initial conjunction for better readability (in terms of precedence).
Perhaps it's also worth using case-insensitive tests and, for example, simply checking if MODEL then contains "sdk"?

This is a bit of a micro-optimization, but I've noticed this code is [referenced on Stack Overflow, etc.,](https://stackoverflow.com/a/55355049/2872279) and this could potentially be simplified further. It should also stop additional checks being erroneously added.

## Checklist

- [X] I did not modify the `CHANGELOG.md` nor the `pubspec.yaml` files.

## Breaking Change

### DIFF
--- a/packages/device_info_plus/device_info_plus/android/src/main/kotlin/dev/fluttercommunity/plus/device_info/MethodCallHandlerImpl.kt
+++ b/packages/device_info_plus/device_info_plus/android/src/main/kotlin/dev/fluttercommunity/plus/device_info/MethodCallHandlerImpl.kt
@@ -107,7 +107,7 @@ internal class MethodCallHandlerImpl(
      * detection systems
      */
     private val isEmulator: Boolean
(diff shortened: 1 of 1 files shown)


vbuberen (approved):
Thanks for your contribution.

> Perhaps it's also worth using case-insensitive tests and, for example, simply checking if MODEL then contains "sdk"?

We are fine with the current state, so no need to add case-insensitive checks.

PR #1798, opened 2023-05-10
fix: Add jvm target compatibility to Kotlin plugins

davidmartos96:
## Description

When using Gradle 8+ and Kotlin version 1.8+, the build fails with the following error.
(Using the recently uploaded versions of the plugins)

image

I believe the correct fix is to indicate the `jvmTarget` in the `kotlinOptions` to all plugins that use Kotlin, same as it is done for the `compileOptions` with Java

```gradle
kotlinOptions {
    jvmTarget = '1.8'
}
```

Should the `compileOptions` be kept in Kotlin plugins as well?
```gradle
compileOptions {
    sourceCompatibility JavaVersion.VERSION_1_8
    targetCompatibility JavaVersion.VERSION_1_8
}
``` 

## Checklist

## Breaking Change

### DIFF
--- a/packages/battery_plus/battery_plus/CHANGELOG.md
+++ b/packages/battery_plus/battery_plus/CHANGELOG.md
@@ -1,7 +1,3 @@
-## 4.0.1
-
-- **FIX**: Support building with Kotlin 1.8+
-
 ## 4.0.0
 
 > Note: This release has breaking changes.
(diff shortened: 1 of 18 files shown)


vbuberen (approved):

vbuberen:
Good catch. It is due to using JDK 17 with switch to AGP 8 + Gradle 8. Forgot to add it while was adding target for Java recently. Thanks for opening a PR.

PR #1805, opened 2023-05-11
chore: Win32 dependency upgrade

timsneath:
## Description

This PR updates the dependent version of `package:win32` to the range `>=4.0.0 <6.0.0`. v4 and v5 are compatible for Win32 API calls, but incompatible for COM API calls. At present, every usage of `package:win32` is only for traditional Win32 API calls, so we're good. If at a later stage, a package adopts COM, then we can increase the range knowing that a v4-compatible version is still available.

The PR also migrates a few tests to the minimum version of 2.12 or later.

## Checklist

- [X] I did not modify the `CHANGELOG.md` nor the plugin version in `pubspec.yaml` files.

## Breaking Change


vbuberen (approved):
Thanks for another contribution.
Looks good to me.

Issue #1810, opened 2023-05-12
[Bug]: Dependencies targeting android 34 cannot use battery_plus


reidbaker:
### Platform

Android api 34

### Plugin

battery_plus

### Version

4.0.0

### Flutter SDK

Main

### Steps to reproduce

Android API 34 will enforce broadcast receivers to declare if they wish to be exported. 

Battery plus uses a broadcast receiver and does not specify if it is exported apps targeting api 34 will not be able to compile if using battery plus. 

I tried to find publicly available documentation or comment that this is happening but everything I found is internal. 
This was found because of a google customer of battery plus trying to compile against the latest 34 implementation. 

More context:
https://developer.android.com/guide/components/broadcasts#context-registered-receivers
https://developer.android.com/guide/topics/manifest/receiver-element
https://developer.android.com/reference/androidx/core/content/ContextCompat#registerReceiver(android.content.Context,android.content.BroadcastReceiver,android.content.IntentFilter,int)

### Logs

```shell
<can not share>
```

### Flutter Doctor

```shell
Irrelevant
```

### Checklist before submitting a bug

- [X] I Google'd a solution and I couldn't find it
- [X] I searched on StackOverflow for a solution and I couldn't find it
- [X] I read the README.md file of the plugin
- [X] I'm using the latest version of the plugin
- [X] All dependencies are up to date with `flutter pub upgrade`
- [X] I did a `flutter clean`
- [X] I tried running the example project

PR #1811, opened 2023-05-12
fix(battery_plus): Use context compat to set exported state

reidbaker:
## Description
Most of the context is in the bug but short summary is that broadcast receivers will require and export state if apps are targeting android api 34. Unfortunately I cant find a public document or post saying this but I see this on internal facing documents. 

There is an internal to google team using battery_plus who wishes to target api 34 and they reached out to me as the TL of flutter on android to help get their dependencies updated. You can see a similar issues and a pull request to fix our own url_launcher plugin below. 

I think the other examples of registerReceiver are ok because they are protected by a version check and not called on api 34. If that is incorrect I can fix those in a different pr. 

## Related Issues

https://github.com/fluttercommunity/plus_plugins/issues/1810
https://github.com/flutter/flutter/issues/126460
https://github.com/flutter/packages/pull/3973

## Checklist

## Breaking Change

### DIFF
--- a/packages/battery_plus/battery_plus/android/src/main/kotlin/dev/fluttercommunity/plus/battery/BatteryPlusPlugin.kt
+++ b/packages/battery_plus/battery_plus/android/src/main/kotlin/dev/fluttercommunity/plus/battery/BatteryPlusPlugin.kt
@@ -84,7 +84,6 @@ class BatteryPlusPlugin : MethodCallHandler, EventChannel.StreamHandler, Flutter
     @SuppressLint("WrongConstant") // Error in ContextCompat for RECEIVER_NOT_EXPORTED
     override fun onListen(arguments: Any?, events: EventSink) {
         chargingStateChangeReceiver = createChargingStateChangeReceiver(events)
-        // DO NOT MERGE, this alternates states. reidbaker debug before review.
         applicationContext?.let {
             ContextCompat.registerReceiver(
                 it, chargingStateChangeReceiver,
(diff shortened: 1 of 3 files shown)


reidbaker:
Not ready for review. I am debugging why modfying this code causes the state of charging to constantly change.

packages/battery_plus/battery_plus/example/android/build.gradle
```diff
@@ -26,6 +26,6 @@ subprojects {
     project.evaluationDependsOn(':app')
 }
 
-task clean(type: Delete) {
+tasks.register("clean", Delete) {
```
reidbaker, line 29: This change is an auto migration from a more recent version of flutter. Happy to remove if you want.

vbuberen:
> Unfortunately I cant find a public document or post saying this but I see this on internal facing documents.

I think you meant something like this: https://developer.android.com/about/versions/14/behavior-changes-14

packages/battery_plus/battery_plus/android/src/main/kotlin/dev/fluttercommunity/plus/battery/BatteryPlusPlugin.kt
```diff
@@ -77,9 +81,16 @@ class BatteryPlusPlugin : MethodCallHandler, EventChannel.StreamHandler, Flutter
         }
     }
 
+    @SuppressLint("WrongConstant") // Error in ContextCompat for RECEIVER_NOT_EXPORTED
```
reidbaker, line 84: Note I filed a bug against androidx for this behavior https://b.corp.google.com/issues/282596027

reidbaker:
FYI we saw https://github.com/flutter/flutter/issues/127014 after merging our version of this. I checked out battery_plus and rebuilt the example app and make sure that pressing back does not crash. So I believe this is safe but for transparency I wanted to add this comment.

PR #1824, opened 2023-05-16
fix: Value of key `appName` possibly not of type String on iOS.

markus-dot-fuereder:
### `package_info_plus`

After upgrading flutter SDK (see flutter doctor below) ...

in `method_channel_package_info.dart`:
**Value of key 'appName' possibly is not of type String on iOS.**

image

I think the type boils down to `Char` as a native type coming from the MethodChannel. I Would prefer to ensure the type in the native implementation, but I won't touch your Objective-C implementation.
The `?.toString()` should not hurt in any already expected case.
For whatever reason this only applies to the `appName`, but obviously the other values may cause the same error. So far I just fixed what is broken.

Let me know what you think.

### Environment
`pubspec.yaml`:
```
environment:
  sdk: ">=2.10.0 <3.0.0"

dependencies:
  package_info_plus: ^4.0.0
  ...
```

_Flutter Doctor:_
```
[!] Flutter (Channel stable, 3.7.12, on macOS 13.3.1 22E772610a darwin-arm64, locale en-US)
    • Flutter version 3.7.12 on channel stable at /Users/<USER>/.fvm/versions/3.7.12
    ! Warning: `dart` on your path resolves to /opt/homebrew/Cellar/dart/3.0.0/libexec/bin/dart, which is not inside your current Flutter SDK checkout at /Users/<USER>/.fvm/versions/3.7.12. Consider adding /Users/<USER>/.fvm/versions/3.7.12/bin to the front of your path.
    • Framework revision 4d9e56e694 (4 weeks ago), 2023-04-17 21:47:46 -0400
    • Engine revision 1a65d409c7
    • Dart version 2.19.6
    • DevTools version 2.20.1
    • If those were intentional, you can disregard the above warnings; however it is recommended to use "git" directly to perform update checks and upgrades.

[✓] Android toolchain - develop for Android devices (Android SDK version 33.0.0)
    • Android SDK at /Users/<USER>/Library/Android/sdk
    • Platform android-33, build-tools 33.0.0
    • ANDROID_HOME = /Users/<USER>/Library/Android/sdk
    • Java binary at: /Applications/Android Studio.app/Contents/jre/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 11.0.12+0-b1504.28-7817840)

[✓] Xcode - develop for iOS and macOS (Xcode 14.2)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 14C18
    • CocoaPods version 1.11.3

[✓] Android Studio (version 2021.2)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build 11.0.12+0-b1504.28-7817840)

[✓] VS Code (version 1.78.2)
    • VS Code at /Applications/Visual Studio Code.app/Contents
    • Flutter extension version 3.64.0
```

## Checklist

## Breaking Change

### DIFF
--- a/packages/package_info_plus/package_info_plus_platform_interface/lib/method_channel_package_info.dart
+++ b/packages/package_info_plus/package_info_plus_platform_interface/lib/method_channel_package_info.dart
@@ -12,7 +12,7 @@ class MethodChannelPackageInfo extends PackageInfoPlatform {
   Future<PackageInfoData> getAll() async {
     final map = await _channel.invokeMapMethod<String, dynamic>('getAll');
     return PackageInfoData(
-      appName: map!['appName'] ?? '',
+      appName: map!['appName']?.toString() ?? '',
       packageName: map['packageName'] ?? '',
       version: map['version'] ?? '',
       buildNumber: map['buildNumber'] ?? '',


vbuberen:
I don't see this issue even with Flutter 3.10.0 and 3.10.1, so I have doubts about the need of this change.

markus-dot-fuereder:
> I don't see this issue even with Flutter 3.10.0 and 3.10.1, so I have doubts about the need of this change.
Have you tried to reproduce with 3.7.12? Because I'm facing the issue^^.
As said, the cause of the issue probably is due to typing in the native iOS implementation. This change is just a shortcut.

vbuberen:
> Have you tried to reproduce with 3.7.12?

Example project builds fine on iOS with 3.7.12 and there are no linter issues on this line as well. 
You can't build the project?

> This change is just a shortcut.

If there is no problem there is no sense in adding code, even if it is a simple `toString()`. So it doesn't matter what type of `just` it is.

markus-dot-fuereder:
> > Have you tried to reproduce with 3.7.12?
> 
> Example project builds fine on iOS with 3.7.12 and there are no linter issues on this line as well. You can't build the project?

Yes, I can build the project, but it throws a `TypeError` in runtime, because the native iOS implementation is returning `MapEntry<String, List<Object?>>` instead of `MapEntry<String, String>` for the key "appName".
As said, my change is just a shortcut if you want to resolve this issue on the Flutter side. But the root cause happens is iOS.

So if you want to solve this issue on the iOS side, someone need to would ensure the correct type in iOS and revert my change.


> > This change is just a shortcut.
> 
> If there is no problem there is no sense in adding code, even if it is a simple `toString()`. So it doesn't matter what type of `just` it is.

Just because you cannot reproduce the issue on your machine does not mean that there is no problem. And build errors are different from runtime errors.

If you do not want to investigate that's fine, then I just dump the plugin and write the desired functionality by myself. I just need to know if I can use this plugin in the future or not.

vbuberen:
> Just because you cannot reproduce the issue on your machine does not mean that there is no problem. 
> If you do not want to investigate that's fine, then I just dump the plugin and write the desired functionality by myself. I just need to know if I can use this plugin in the future or not.

There would be reports from other users as this code wasn't touched for quite a long time. Considering that there is no reproducer and that I tried to build the example app of the package with 3 different Flutter versions already I believe that you have some local issue. I can claim the same that if the issue happens only on your machine it doesn't mean that there is a problem. 

If you don't have a reproducer sample that can confirm the issue by someone else I can't help you here and I definitely won't be accepting a PR fixing the issue without a reproducer. Feel free to dump the plugin or whatever. 

> And build errors are different from runtime errors.

Thanks for enlightening me, but could you think a bit more that if the person builds an example it means that the plugin code was called, namely the function `fromPlatform()` in question? 
Would be good to communicate in a bit more respectful manner even if the person you communicate with doesn't agree with you.

markus-dot-fuereder:
Ok, if you cannot reproduce then maybe it is a local issue, we'll see. At least now you heard it once.

I did not mean to be disrespectful, I just wanted you to understand the problem, since you did not answer my question if you were actually trying to reproduce. You just mentioned that it builds on different versions and you explained me some principles^^.

> Thanks for enlightening me, but could you think a bit more that if the person builds an example successfully it means that the plugin code was called on device/simulator, namely the function fromPlatform() in question?

Yes, I absolutely do, but I have a device where it fails and maybe you want to know about it.

> Would be good to communicate in a bit more respectful manner even if the person you communicate with doesn't agree with you.
Maybe do not take every word personal, passive aggressive or whatever. It probably is not meant that way. And this is true for this very sentence as well^^.

Anyways thanks for your time and effort. Have a good one ✌️

Issue #1821, opened 2023-05-15
[Bug]: Gradle failed building in Android


delizondo:
### Platform

Android

### Plugin

device_info_plus, package_info_plus

### Version

9.0.0

### Flutter SDK

3.10.0

### Steps to reproduce

Trying to build my Flutter app in Android, using Android Studio 2022.2.1

Running Gradle 8.0.0
Kotlin 1.8.0
Java v 17

### Logs

```shell
> Task :app:mergeExtDexDevelopmentDebug

FAILURE: Build completed with 2 failures.

1: Task failed with an exception.
-----------
* What went wrong:
Execution failed for task ':device_info_plus:compileDebugKotlin'.
> 'compileDebugJavaWithJavac' task (current target is 1.8) and 'compileDebugKotlin' task (current target is 17) jvm target compatibility should be set to the same Java version.
  Consider using JVM toolchain: https://kotl.in/gradle/jvm/toolchain


* Try:
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':device_info_plus:compileDebugKotlin'.
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.lambda$executeIfValid$1(ExecuteActionsTaskExecuter.java:149)
	at org.gradle.internal.Try$Failure.ifSuccessfulOrElse(Try.java:282)
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeIfValid(ExecuteActionsTaskExecuter.java:147)
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:135)
	at org.gradle.api.internal.tasks.execution.FinalizePropertiesTaskExecuter.execute(FinalizePropertiesTaskExecuter.java:46)
	at org.gradle.api.internal.tasks.execution.ResolveTaskExecutionModeExecuter.execute(ResolveTaskExecutionModeExecuter.java:51)
	at org.gradle.api.internal.tasks.execution.SkipTaskWithNoActionsExecuter.execute(SkipTaskWithNoActionsExecuter.java:57)
	at org.gradle.api.internal.tasks.execution.SkipOnlyIfTaskExecuter.execute(SkipOnlyIfTaskExecuter.java:74)
[...]

PR #1825, opened 2023-05-17
fix(device_info_plus, package_info_plus): fixed a bug!

Marco4763:
…rom the one defined in the project.

## Description

In this PR was added a condition to check if the java version defined in the project is different from the default version of the package, if it is different, we get the version from the project, if it is not, we define the default version.

## Related Issues

#1821

*e.g.*
- *Fix ##1821*

## Checklist

## Breaking Change

### DIFF
--- a/packages/device_info_plus/device_info_plus/android/build.gradle
+++ b/packages/device_info_plus/device_info_plus/android/build.gradle
@@ -30,8 +30,8 @@ android {
     namespace 'dev.fluttercommunity.plus.device_info'
 
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_1_8
-        targetCompatibility JavaVersion.VERSION_1_8
+        sourceCompatibility project.sourceCompatibility != JavaVersion.VERSION_1_8 ? project.sourceCompatibility : JavaVersion.VERSION_1_8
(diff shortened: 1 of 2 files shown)


Marco4763:
@vbuberen ??

packages/device_info_plus/device_info_plus/android/build.gradle
```diff
@@ -30,8 +30,8 @@ android {
     namespace 'dev.fluttercommunity.plus.device_info'
 
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_1_8
-        targetCompatibility JavaVersion.VERSION_1_8
+        sourceCompatibility project.sourceCompatibility != JavaVersion.VERSION_1_8 ? project.sourceCompatibility : JavaVersion.VERSION_1_8
+        targetCompatibility project.targetCompatibility != JavaVersion.VERSION_1_8 ? project.targetCompatibility : JavaVersion.VERSION_1_8
```
nohli, line 34: How is that logic different from the following?

```suggestion
        sourceCompatibility project.sourceCompatibility
        targetCompatibility project.targetCompatibility
```

Marco4763:
@nohli In the end the result would be the same as the condition only works if the version is different from the project, so now that I see it I agree that it's more logical to just set the value directly without the condition.

PR #1862, opened 2023-05-29

vbuberen:
## Description

New release. Mainly updated constraints for `http` and `file` packages to let people use Dart 3 versions of these packages.
Additionally `sensors_plus` potentially gets a fix for listening multiple sensors events on iOS.


vbuberen:
Looks like we would need to re-generate iOS part of example apps to resolve these issues, like some other plugins did to overcome failing builds on iOS and MacOS with Xcode 14.3

PR #1898, opened 2023-06-07
fix(all): Add namespace for compatibility with AGP <4.2

aliyazdi75:
## Description

These changes need for compatibility with AGP <4.2. For more information: https://developer.android.com/studio/build/agp-upgrade-assistant

## Checklist

## Breaking Change

### DIFF
--- a/packages/android_alarm_manager_plus/android/build.gradle
+++ b/packages/android_alarm_manager_plus/android/build.gradle
@@ -22,6 +22,10 @@ rootProject.allprojects {
 apply plugin: 'com.android.library'
 
 android {
+    // Conditional for compatibility with AGP <4.2.
+    if (project.android.hasProperty("namespace")) {
+        namespace 'dev.fluttercommunity.plus.androidalarmmanager'
+    }
     compileSdkVersion 31
 
     namespace 'dev.fluttercommunity.plus.androidalarmmanager'
--- a/packages/android_intent_plus/android/build.gradle
+++ b/packages/android_intent_plus/android/build.gradle
@@ -22,6 +22,10 @@ rootProject.allprojects {
 apply plugin: 'com.android.library'
 
 android {
+    // Conditional for compatibility with AGP <4.2.
+    if (project.android.hasProperty("namespace")) {
+        namespace 'dev.fluttercommunity.plus.androidintent'
+    }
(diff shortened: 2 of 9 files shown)


vbuberen:
Thanks for your contribution, but there are no plans to support AGP < 4.2 in Plug Plugins. Recently there was a release of major version for all plugins specifically because of dropping support of such old AGP versions. Developers should update their projects instead of staying on such an old version of the Android Gradle plugin,

aliyazdi75:
Almost all popular packages such as flutter's team packages, have the `namespace` property in their gradles. We want to upgrade our project and the upgrade process fails due to the lack of this.
>  Developers should update their projects instead of staying on such an old version of the Android Gradle plugin,

Also I don't understand this sentence, because we're trying to upgrade to the version 8 of gradle and it is not possible if you don't add these changes to the plugins. If this is not the case we have to clone the plus plugins unfortunately and add them on our side.

vbuberen:
> Almost all popular packages such as flutter's team packages, have the namespace property in their gradles. We want to upgrade our project and the upgrade process fails due to the lack of this.

There was a discussion about it, but long story short it is up to maintainers to decided on how to develop a plugin they work in their free time. Plus Plugins have `namespace` property in latest releases, so updating to Gradle 8 and Android Gradle Plugin 8.x isn't an issue. What you introduced in the PR is a workaround for projects with Android Gradle Plugin older than 4.2.

Just look at the diff of your PR - there is a namespace property already below:
Screenshot 2023-06-08 at 18 16 48

> Also I don't understand this sentence, because we're trying to upgrade to the version 8 of gradle and it is not possible if you don't add these changes to the plugins. 

This contradicts the previous sentence in your reply. If you want to update to Gradle 8 and Android Gradle Plugin 8.x you can use Plus Plugins and don't need such workarounds as `namespace` was introduced in Android Gradle Plugin 4.2 and available in any project using 4.2 or newer version.

aliyazdi75:
> Just look at the diff of your PR - there is a namespace property already below:
Oh sorry, it's right. I don't know why we faced gradle issues on plus plugin then. It might be related to the using old versions of plus plugins and I had an old clone version.
Thanks it's all good. We will try to update it.

PR #1292, opened 2022-10-26
docs(device_info_plus): Document toMap deprecation

miquelbeltran:
## Description

Documents the deprecation of the `toMap` method, explaining that it is not JSON compatible and shouldn't be used as such.

## Related Issues

- closes  #1291

## Checklist

- [x] I read the Contributor Guide and followed the process outlined there for submitting PRs.
- [x] I titled the PR using [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0).
- [x] I did not modify the `CHANGELOG.md` nor the `pubspec.yaml` files.
- [x] All existing and new tests are passing.
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate that with a `!` in the title as explained in [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0)).
- [x] No, this is *not* a breaking change.

PR #1142, opened 2022-10-02
[device_info_plus] deprecate "toMap" method

iapicca:
## Description

as the current implementation of toMap output an invalid json
(see discussion in this https://github.com/fluttercommunity/plus_plugins/issues/949#issue-1297230725 and https://github.com/fluttercommunity/plus_plugins/pull/950)
a suggested solution is to deprecate "toMap" in it's entirety

## Related Issues

fix https://github.com/fluttercommunity/plus_plugins/issues/1139

## Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [x] All existing and new tests are passing.
- [x] I updated the version in `pubspec.yaml` and `CHANGELOG.md`.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [x] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [ ] No, this is *not* a breaking change.

PR #1926, opened 2023-06-23
feat(package_info_plus): Added toMap method

mcorbelli:
## Description

This PR adds a `toMap()` function to the `PackageInfo` class in the `package_info_plus` plugin. Previously, this functionality was missing. The `toMap()` function returns a map representation of the `PackageInfo` instance. Notably, the `buildSignature` and `installerStore` fields are included in the map only if they are non-null and non-empty. This change ensures that the returned map only contains meaningful and useful data.

## Checklist

## Breaking Change

### DIFF
--- a/packages/package_info_plus/package_info_plus/lib/package_info_plus.dart
+++ b/packages/package_info_plus/package_info_plus/lib/package_info_plus.dart
@@ -121,22 +121,14 @@ class PackageInfo {
   }
 
   Map<String, dynamic> _toMap() {
-    final values = <String, dynamic>{
+    return {
       'appName': appName,
       'buildNumber': buildNumber,
       'packageName': packageName,
       'version': version,
+      if (buildSignature.isNotEmpty) 'buildSignature': buildSignature,
+      if (installerStore?.isNotEmpty ?? false) 'installerStore': installerStore,
     };
-
-    if (buildSignature.isNotEmpty) {
-      values['buildSignature'] = buildSignature;
-    }
-
-    if (installerStore != null && installerStore!.isNotEmpty) {
-      values['installerStore'] = installerStore;
-    }
-
-    return values;
   }
 
   /// Gets a map representation of the [PackageInfo] instance.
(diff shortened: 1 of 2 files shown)


nohli:
While I like it, we deprecated `toMap()` from `device_info_plus` last year.
https://github.com/fluttercommunity/plus_plugins/pull/950#issuecomment-1187329379
https://github.com/fluttercommunity/plus_plugins/issues/949#issuecomment-1264675131
https://github.com/fluttercommunity/plus_plugins/pull/1142
https://github.com/fluttercommunity/plus_plugins/pull/1292

mcorbelli:
I see the problem, can I create a data getter that retrieves data from the toMap as done for device_info_plus? @nohli

packages/package_info_plus/package_info_plus/lib/package_info_plus.dart
```diff
@@ -119,4 +119,26 @@ class PackageInfo {
...
+    if (buildSignature.isNotEmpty) {
+      values['buildSignature'] = buildSignature;
+    }
+
+    if (installerStore != null && installerStore!.isNotEmpty) {
+      values['installerStore'] = installerStore;
+    }
+
+    return values;
+  }
```
nohli, line 140: How about this?

```suggestion
  Map<String, dynamic> _toMap() {
    return {
      'appName': appName,
      'buildNumber': buildNumber,
      'packageName': packageName,
      'version': 'version',
      if (buildSignature.isNotEmpty) 'buildSignature': buildSignature,
      if (installerStore?.isNotEmpty ?? false) 'installerStore': installerStore,
    };
  }
```

nohli:
Thanks for the change. Makes sense to me, since we have it `device_info_plus`, too.

Left a code suggestion.

packages/package_info_plus/package_info_plus/lib/package_info_plus.dart
mcorbelli: The way it works is the same. Personally I tend to avoid conditions along the same lines in favor of more visible ones. But no problem.

nohli: In my opinion, the more concise version appears to be more commonly used these days.
People familiar with Dart will be able to read it quickly either way.

nohli:
Great, I like it.

Would you mind adding tests? Can set the data with `setMockInitialValues` and then expect the correct map (also with `buildSignature` and `installerStore` being empty/null)?

nohli (approved):
LGTM!

PR #1944, opened 2023-07-03
chore(deps): Update share_plus example

vbuberen:
## Description

Switched to a wider range of supported `image_picker` versions in share_plus example app as suggested in https://github.com/flutter/packages/blob/main/packages/image_picker/image_picker/CHANGELOG.md#100

Additionally did a small modification for example's text field hints.

## Checklist

## Breaking Change


nohli:
Widening the constrains is not necessary for an app - if `pub get` is successful.

The changelog in the link talks about `packages` - so that `pub get` still works when using the package:
If importing `image_picker` with a dependency of >= 1, and at the same time depending on `image_picker` < 1 through a transitive dependency of another package, `pub get` wouldn't work, that's why the dependencies get widened for packages.

Issue #1467, opened 2023-01-20
[Request]: Support ARM Simulator


spkersten:
### Plugin

connectivity_plus, device_info_plus, share_plus

### Use case

Currently, apps using `connectivity_plus` (or any other plugin from this repo) can't be build for ARM Simulators (on ARM Macs) because it explicitly requires only x86_64. For example:
https://github.com/fluttercommunity/plus_plugins/blob/f0158c7a167c329e0e454fd086c2d6af8f26ea34/packages/connectivity_plus/connectivity_plus/ios/connectivity_plus.podspec#L23

So Simulator has to run under Rosetta which is slower can running natively. `VALID_ARCHS` seems to be deprecated in favour of `EXCLUDED_ARCHS` (e.g., to exclude i863).

### Proposal

Support ARM Simulators by using `EXCLUDED_ARCHS`.

vbuberen:
Thanks for reporting. 

Opened PRs to fix it. Unfortunately, I don't know the historical reasons why the set of architectures was limited, but don't think that it still needed.

nohli:
@spkersten I can't reproduce it with the latest Flutter version, seems to build fine on Apple silicon.

Do you still encounter this issue?
If yes, or if you use an older Flutter version where you encounter this issue -> is the error fixed with the PR branch?

spkersten:
Judging from the linked PRs the issue should be fixed.

PR #2021, opened 2023-08-04
feat(connectivity_plus): Remove deprecated VALID_ARCHS iOS property

vbuberen:
## Description

~~Found out that some plugins have explicit exception to only allow x86_64 arches which didn't allow to run builds on ARM Macs natively.
Note, I currently don't have any ARM Mac to validate that this doesn't break anything, so would appreciate checking out if iOS builds correctly after this change.~~

Removed deprecated iOS build config property.

## Related Issues

#1467

## Breaking Change

### DIFF
--- a/packages/connectivity_plus/connectivity_plus/ios/connectivity_plus.podspec
+++ b/packages/connectivity_plus/connectivity_plus/ios/connectivity_plus.podspec
@@ -20,5 +20,5 @@ Downloaded by pub (not CocoaPods).
   s.dependency 'ReachabilitySwift'
   s.platform = :ios, '11.0'
   s.swift_version = '5.0'
-  s.pod_target_xcconfig = { 'DEFINES_MODULE' => 'YES', 'VALID_ARCHS[sdk=iphonesimulator*]' => 'x86_64' }
+  s.pod_target_xcconfig = { 'DEFINES_MODULE' => 'YES' }
 end


nohli:
Great! I can test later 👍🏽

vbuberen:
If everything is correct you should see in Activity Monitor the following picture (note the `Kind` field):
Screenshot 2023-07-07 at 10 41 11

nohli:
I ran the app on the iOS simulator and as a macOS app, and there is no difference between `main` and this branch.

Both run natively on Apple silicon.

Maybe the behavior was changed and the issue is obsolete? But in any way, imho it makes sense to merge this and the other PRs, what do you think?

Screenshot 2023-08-07 at 21 09 06
Screenshot 2023-08-07 at 21 05 31

vbuberen:
I was dealing with this topic on my job a few months ago trying to understand why our ARM build machines were running integrations tests via Rosetta. I have just revisited what I was doing there and it looks like I confused [`EXCLUDED_ARCHS` ](https://developer.apple.com/documentation/Xcode/build-settings-reference#Excluded-Architectures)property with `VALID_ARCHS` as in this case we have. 

Did some more research and it looks like `VALID_ARCHS` is [deprecated](https://developer.apple.com/documentation/technotes/tn3117-resolving-build-errors-for-apple-silicon/#Overview), so we can merge this and other related PRs anyway. I will adjust the title for every PR accordingly.

PR #2055, opened 2023-08-21
ci(all): Update Java to 17

vbuberen:
## Description

Bumping Java version for all example build jobs to support upgrade to Gradle 8, so there are no issues like: https://github.com/fluttercommunity/plus_plugins/actions/runs/5923071232/job/16058055161?pr=2010

## Breaking Change


nohli (approved):
Great!

Had to restart some flaky tests a few times.

PR #1989, opened 2023-07-28
fix(connectivity_plus): Downgrade js version to work with Flutter 3.3.10

vbuberen:
## Description

Made the plugin use older `js` version as there were issues with compatibility for older Flutter versions as reported in #1960

## Related Issues

Fixes #1960

## Checklist

- [x] I read the [Contributor Guide](https://github.com/fluttercommunity/plus_plugins/blob/main/CONTRIBUTING.md) and followed the process outlined there for submitting PRs.
- [x] I titled the PR using [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0).
- [x] I did not modify the `CHANGELOG.md` nor the plugin version in `pubspec.yaml` files.
- [x] All existing and new tests are passing.
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate that with a `!` in the title as explained in [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0)).
- [x] No, this is *not* a breaking change.

PR #2056, opened 2023-08-21
chore(all): Adjust dependabot settings

nohli:
## Description

As partly discussed in [this PR](https://github.com/fluttercommunity/plus_plugins/pull/1989#issuecomment-1655610449).

## Related Issues

Our dependabot update policy continuously tightens the constraints, updating every single patch release - making the latest versions of plus_plugins less compatible with the rest of the ecosystem.

I'll just open this PR and leave it to discussion whether any included changes make sense for `plug_plugins`.

Following the changes plus my understanding of the benefit:
1. Ignore dependabot pub updates for minor and patch releases:
If a plugin declares a dependency x of `1.0.0`, it's potentially compatible with version `1.0.0` up to `<2.0.0` and apps using the same dependency itself, or other packages using it (within the range `>=1.0.0 <2.0.0`). Updating to i.e. `1.2.4` tightens the constraints of all apps and packages using the plugin, making it less compatible with other packages, while bringing no benefit other than forcing users or other package maintainers to update their dependencies, which should not be our goal (unless there is an incompatibility with an older version, in which case the dependency should be manually updated to exclude lower versions).
2. Change the website dependency check from `weekly` to `monthly`. Idk if the overhead of almost weekly updating a lint dependency for the website brings a benefit.
3. Add dependabot gradle checks (please let me know if they were missing for a reason).

## Checklist

## Breaking Change

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -37,6 +37,22 @@ updates:
       - dependency-name: "*"
         update-types: ["version-update:semver-minor", "version-update:semver-patch"]
 
+  - package-ecosystem: "gradle"
+    directory: "/packages/android_alarm_manager_plus/android"
+    schedule:
+      interval: "monthly"
+    commit-message:
+      prefix: "chore"
+      include: "scope"
+
+  - package-ecosystem: "gradle"
+    directory: "/packages/android_alarm_manager_plus/example/android"
+    schedule:
+      interval: "monthly"
+    commit-message:
+      prefix: "chore"
+      include: "scope"
+
   - package-ecosystem: "pub"
     directory: "/packages/android_alarm_manager_plus/example"
     schedule:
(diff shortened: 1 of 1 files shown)


vbuberen:
I totally agree here with the suggested update. Same applies to Gradle plugin updates, like currently open bump to 8.1.0 - I don't think it makes sense to push that hard with only latest.

In general, except for what you described I would suggest to bump some dependency version also in case there is some important fix (security or critical bug) introduced - I believe such cases worth it to bump minor versions as well.

> Add dependabot gradle checks (please let me know if they were missing for a reason).

I believe that we just missed those parts. Namely, it was me, who worked on this Dependabot config and I could forgot a few items worth watching.

vbuberen (approved):

nohli:
Ok cool. I also agree about the importance of manual updates if it's critical or necessary.

How to proceed with the Gradle parts (so far unchanged in this PR)?
In my understanding, they don't behave the same as pub dependencies (you could have a higher version in a package than in your app). Your feedback is important, since you're the Android expert here 😁

vbuberen:
> How to proceed with the Gradle parts (so far unchanged in this PR)?

Didn't get what you mean here

vbuberen:
> The dependabot Gradle update policies remain unchanged in this PR.
> 
> Should we add the same as for pub dependencies?

Understood now. I would suggest to also make Gradle checks monthly as we really don't need to keep that on par with Android ecosystem in plugins. As to update types I would only drop patch updates as from experience minor releases in Android libraries are usually useful with fixes/new features. 

We can do it in another PR.

P.S. A bit off-topic, but I have an itch to try and replace Dependabot with Renovate for some time already as it seems to me that we will have a better experience with configs/updates in Renovate.

nohli:
Good idea trying out another tool for dependency updates 👍🏽

PR #2057, opened 2023-08-21
ci(all): Update Java version for integration tests, update Android APIs

vbuberen:
## Description

I missed the fact that Java version is set in 2 places in workflow files and didn't update Java for integration tests in #2055

Additionally I looked at a few other repos that use Android emulators Action that we use and found out that a few of such repos identify 21 and 32 as flaky emulators, which is something we often experience ourselves. As a result I changed min Android API to 22 and max to 31 in attempt to get lower number of failed CI tests due to flakiness of some emulators.

## Breaking Change

### DIFF
--- a/.github/workflows/android_alarm_manager_plus.yaml
+++ b/.github/workflows/android_alarm_manager_plus.yaml
@@ -43,7 +43,7 @@ jobs:
     timeout-minutes: 30
     strategy:
       matrix:
-        android-api-level: [21, 26, 32]
+        android-api-level: [22, 26, 31]
 
     steps:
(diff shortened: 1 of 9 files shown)


nohli (approved):
Nice!

How do you feel about also testing on 33 or 34?

vbuberen:
> How do you feel about also testing on 33 or 34?

I would like to replace 31 with 34 as soon as Google releases Android 14 officially to be our `latest` platform to test.

vbuberen:
Reverted `connectivity_plus` min Android API as we have tests specifically for API 21, so let's leave the same version there.

PR #2092, opened 2023-08-23
ci(android_intent_plus): Remove Java installation step

vbuberen:
## Description

According to Github documentation `macos-13` machine should already have Java 17 installed: https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md#java
Thus, I am curious if we can drop installation step in order to save a few seconds during workflow runs. In case this PR doesn't fail I will open similar ones for other plugins workflows.

## Breaking Change

### DIFF
--- a/.github/workflows/android_intent_plus.yaml
+++ b/.github/workflows/android_intent_plus.yaml
@@ -29,12 +29,6 @@ jobs:
       - name: "Install Tools"
         run: ./.github/workflows/scripts/install-tools.sh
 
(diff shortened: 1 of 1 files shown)


vbuberen:
Checked one of previous runs and it looks like removed step was doing nothing already, so we are not saving time here, but just dropping some redundant code.

Screenshot 2023-08-23 at 13 37 47

nohli (approved):

PR #2198, opened 2023-10-02
ci(all): Disable automatic rebase for Dependabot PRs

vbuberen:
## Description

When Dependabot opens a bunch of PRs it becomes a nightmare to merge them all as every merge causes all other PRs to rebase. It wastes a lot of time and energy for no reason as in most cases these Dependabot PRs are not related. Thus, I am disabling the default behavior of Dependabot to be more energy efficient and save our (maintainers) time as well.

Hope to not see such situation again:
Screenshot 2023-10-02 at 16 52 05

## Breaking Change

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,6 +9,7 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    rebase-strategy: "disabled"
     commit-message:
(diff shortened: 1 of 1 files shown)


nohli (approved):
Exactly! 🙈

Thanks for changing 🙏🏽

Issue #1280, opened 2022-10-21
[Request]: Adjusting sample rate for web with Sensors Plus plugin


bjornwinther1996:
### Plugin

Sensors Plus

### Use case

There has been a discussion about adjusting the sample rate for sensor data in the Sensors_plus flutter package (https://github.com/flutter/flutter/issues/25971). (Sensors_plus is the only sensor package for web).

Following the above thread, I managed to adjust the rate when building for Android, by changing SENSOR_DELAY_NORMAL to SENSOR_DELAY_GAME.

Now my question is: Is there a way to adjust the sample rate for sensordata when building for web?

### Proposal

Be able to adjust the sample rate / update rate for sensor data when building for web? For the sensors in the package (Gyroscope, Accelerometer etc.)

PR #2248, opened 2023-10-11
feat(sensors_plus): Configurable sample rate

### LINKED ISSUE #439: [plugin][sensors_plus] Need to specify sensor delay level.
## Use case
In my case, I need to specify sensor delay level.
```
sensorManager.registerListener(sensorEventListener, sensor, sensorManager.SENSOR_DELAY_NORMAL);
```

## Proposal
sensorManager.SENSOR_DELAY_GAME is also needed.

### LINKED ISSUE #1318: [Request]: Sample rate for accelerometer and gyroscope
### Plugin

sensors_plus

### Use case

Listening to accelerometer and gyroscope starts bombarding app with realtime data, there should be an option to limit sample rate per second.

### Proposal

I think it would be better if it's done on the native side, rather limiting event in dart side to get optimal performance.

### LINKED ISSUE #1381: [Request]: possibility to set SensorDelay on android
### Plugin

sensors_plus

### Use case

For some use cases it would be useful to be able to change the sensor delay. It's currently always set to value SENSOR_DELAY_NORMAL.

### Proposal

Not sure what the smartest way to do it is. Maybe add a configuration method for setting the default sensor delay.

tlserver:
## Description

* Support changing `samplingPeriod` for different sensors.
  ```dart
  void main() {
    final oldGetterStillWork = accelerometerEvents; // default to SensorInterval.normalInterval = 200ms
    final newEntryPoint = accelerometerEventStream(samplingPeriod: SensorInterval.gameInterval);
    final customInterval= accelerometerEventStream(samplingPeriod: Duration(seconds: 6));
  }
  ```
* Introduce `SensorInterval` class to provide 4 androids interval constants.
  ```dart
  normalInterval = Duration(milliseconds: 200);
  uiInterval = Duration(milliseconds: 66, microseconds: 667);
  gameInterval = Duration(milliseconds: 20);
  fastestInterval = Duration.zero;
  ```
* Updated example
  image

* Rewrited ios platform-specific code in swift

Attention needed:

* This new feature support both android and ios, but not web (need help).
* The document may need to update and tell user that specific permission is needed for very small sensor interval ( <20ms ? ).

Implementation details:
xxx`EventStream` always return the same stream. If the method is called with different `samplingPeriod` later, the stream is changed to the new value. All previous listeners are affected. Note that this is same behaviour as ios but different than android.

In android, each registered handler has its own sampling period, and the system use the fastest sampling period value. For example, handler A with samplingPeriod=250ms and handler B with samplePeriod = 100ms are registered, the system try to update sensor every 100ms. But if handler B is unregistered later, system try to update sensor every 250ms then.

Known issue:
In previous version, FPPMagnetometerStreamHandlerPlus use `deviceMotion` instead of `magnetometer` to allow iOS present calibration interaction. So, in iOS, userAccelerometerEventStream and magnetometerEventStream share the same sampling period. In android, they have their own sampling period. See also #2250.

## Related Issues

Resolve #439
Resolve #1318
Resolve #1381

#1280

## Checklist

## Breaking Change


vbuberen:
Thank you for this valuable contribution. I will get to check your PRs by the end of this week.

In the meantime, could you please fix formatting issues, so CI checks pass?

vbuberen:
Other issues related to this topic:

https://github.com/fluttercommunity/plus_plugins/issues/439
https://github.com/fluttercommunity/plus_plugins/issues/1280
https://github.com/fluttercommunity/plus_plugins/issues/1381

mengyanshou:
Hello, I am following this PR, but I have some questions and I'm not sure if it's due to my own understanding. Firstly, I'm testing on the Android platform. In the SensorManager, there are the following constants:
```java
    /** get sensor data as fast as possible */
    public static final int SENSOR_DELAY_FASTEST = 0;
    /** rate suitable for games */
    public static final int SENSOR_DELAY_GAME = 1;
    /** rate suitable for the user interface  */
    public static final int SENSOR_DELAY_UI = 2;
    /** rate (default) suitable for screen orientation changes */
    public static final int SENSOR_DELAY_NORMAL = 3;
```
However, in the code here: [code](https://github.com/tlserver/plus_plugins/blob/fe74162f0e1ce9d8b14365938f3a413329900ac3/packages/sensors_plus/sensors_plus/android/src/main/kotlin/dev/fluttercommunity/plus/sensors/StreamHandlerImpl.kt#L28C18-L28C18), I found that a value like var samplingPeriod = 200000 is used. In reality, I'm using a branch dependency in my YAML file:
```yaml
  sensors_plus:
    git:
      url: https://github.com/tlserver/plus_plugins/
      ref: main
      path: packages/sensors_plus/sensors_plus
  sensors_plus_platform_interface:
    git:
      url: https://github.com/tlserver/plus_plugins/
      ref: main
      path: packages/sensors_plus/sensors_plus_platform_interface
```
And I'm using the corresponding Dart code: `gyroscopeEventStream(samplingPeriod: SensorInterval.gameInterval).listen`. I've noticed that the sensitivity isn't very high. I'd like to ask if there's something wrong with this or if I've misunderstood. Thank you.

mengyanshou:
sorry i found this
```java
* @param samplingPeriodUs The desired delay between two consecutive events in microseconds.
*            This is only a hint to the system.  Events may be received faster or slower than
*            the specified rate.  Usually events are received faster.  Can be one of
*            {@link #SENSOR_DELAY_NORMAL}, {@link #SENSOR_DELAY_UI},
*            {@link #SENSOR_DELAY_GAME}, {@link #SENSOR_DELAY_FASTEST} or the delay in
*            microseconds.
```
thanks

packages/sensors_plus/sensors_plus/example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -468,8 +476,10 @@
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ENABLE_MODULES = YES;
 				CURRENT_PROJECT_VERSION = "$(FLUTTER_BUILD_NUMBER)";
+				DEVELOPMENT_TEAM = DDS7MYD673;
```
vbuberen, line 479: Let's drop signing info from this file to no confuse anyone.

packages/sensors_plus/sensors_plus/ios/sensors_plus.podspec
```diff
@@ -14,9 +15,10 @@ Flutter plugin to access the accelerometer, gyroscope, and magnetometer sensors.
...
   s.documentation_url = 'https://pub.dev/packages/sensors_plus'
   s.source_files = 'Classes/**/*'
-  s.public_header_files = 'Classes/**/*.h'
   s.dependency 'Flutter'
-
   s.platform = :ios, '11.0'
-  s.pod_target_xcconfig = { 'DEFINES_MODULE' => 'YES' }
+
+  # Flutter.framework does not contain a i386 slice.
+  s.pod_target_xcconfig = { 'DEFINES_MODULE' => 'YES', 'EXCLUDED_ARCHS[sdk=iphonesimulator*]' => 'i386' }
```
vbuberen, line 22: Why is this change needed? Previously, there were no need for this to build for simulator.

vbuberen (commented):
Looks mostly fine to me except a few minor moments.
Will do some additional testing tomorrow on Android as results on real devices seemed a bit odd during testing today. Afterwards would be good to merge and release.

P.S. Sorry for getting back that late.

packages/sensors_plus/sensors_plus/ios/sensors_plus.podspec
tlserver: I don't know. Since swift language was mirgated to, the whole ios project have been regenerated by the latest version of flutter SDK. And this is the default setting.

vbuberen: Understood. Thought it was added manually.

packages/sensors_plus/sensors_plus/example/ios/Runner.xcodeproj/project.pbxproj
vbuberen: As I understood the request is ignored. Ok, will do it myself.

vbuberen (approved):
Looks good.

As I mentioned in the previous comment on iOS everything works Ok, while on Android depending on version and device delays more than 200 ms don't always work and data is provided faster anyway.
However, it seems Ok based on the description of `samplingPeriodUs` parameter: https://developer.android.com/reference/android/hardware/SensorManager.html#registerListener(android.hardware.SensorEventListener,%20android.hardware.Sensor,%20int)

PR #2316, opened 2023-10-30
fix(package_info_plus): migrate to pkg:web from dart:html

kevmoo:
## Description

Change from using `dart:html` to `package:web`. This will allow this plugin to be used from Wasm.

## Checklist

## Breaking Change

### DIFF
--- a/packages/package_info_plus/package_info_plus/pubspec.yaml
+++ b/packages/package_info_plus/package_info_plus/pubspec.yaml
@@ -46,3 +46,4 @@ dev_dependencies:
 
 environment:
   sdk: ^3.2.0
+  flutter: '>=3.6.0'
--- a/packages/package_info_plus/package_info_plus/pubspec.yaml
+++ b/packages/package_info_plus/package_info_plus/pubspec.yaml
@@ -33,6 +33,7 @@ dependencies:
   meta: ^1.8.0
   path: ^1.8.2
   package_info_plus_platform_interface: ^2.0.1
+  web: '>=0.3.0 <0.5.0'
 
   # win32 is compatible across v4 and v5 for Win32 only (not COM)
   win32: ">=4.0.0 <6.0.0"
@@ -44,5 +45,4 @@ dev_dependencies:
   test: ^1.22.0
 
 environment:
-  sdk: ">=2.18.0 <4.0.0"
-  flutter: ">=3.3.0"
+  sdk: ^3.2.0
--- a/packages/package_info_plus/package_info_plus/lib/src/file_version_info.dart
+++ b/packages/package_info_plus/package_info_plus/lib/src/file_version_info.dart
@@ -12,7 +12,7 @@ import 'dart:io';
 import 'package:ffi/ffi.dart';
 import 'package:win32/win32.dart';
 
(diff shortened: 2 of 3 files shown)


kevmoo:
CC @vbuberen

vbuberen:
Thanks for your contribution.
Too bad that min Dart is that high now. I would postpone this PR a bit.

packages/package_info_plus/package_info_plus/pubspec.yaml
```diff
@@ -44,5 +45,4 @@ dev_dependencies:
   test: ^1.22.0
 
 environment:
-  sdk: ">=2.18.0 <4.0.0"
-  flutter: ">=3.3.0"
```
vbuberen, line 48: Why min Flutter is dropped?

kevmoo: I should add that back. Sorry. In general it's not needed. But it adds some clarity.

kevmoo:
> Thanks for your contribution. Too bad that min Dart is that high now. I would postpone this PR a bit.

This now aligns with the latest stable SDK released on Wednesday.

packages/package_info_plus/package_info_plus/pubspec.yaml
vbuberen: I understand that people with Dart 3.x for sure have more or less fresh version of Flutter as well, but better have some clarity for some users.

kevmoo:
> > This now aligns with the latest stable SDK released on Wednesday.
> 
> I have seen the release, but not all projects update that fast. I would like to not release a breaking version of the plugin now.

Understood.

But isn't this package pretty stable? Are you expecting a lot of changes?

This unblocks our wasm experiments because we have an app that depends on `package_info_plus`.

If folks using old SDKS don't get the LATEST for a while, does it cause problems?

vbuberen:
> But isn't this package pretty stable? Are you expecting a lot of changes?

I am expecting lots of users who can't/don't want to read changelogs starting opening issues in the repo and complaining. Really don't want to deal with it now, like it was recently where lots of people couldn't update their iOS deployment target to 12 as it was required by `connectivity_plus` release. It is a time waste for now, so I would prefer to merge and release it later when have more time and more codebases adopt Dart 3.x.

vbuberen:
To be clear about timelines - it is not months, but a few weeks.

kevmoo:
> To be clear about timelines - it is not months, but a few weeks.

🙏 Thanks!

vbuberen:
Due to another PR that was also a breaking change I would like to combine both of breaking changes and release only one major version bump this week. Thus, will release changes from this PR this week as well.

kevmoo:
Thank you!

vbuberen (approved):

amrgetment:
@kevmoo Do I need to migrate dart:html to package:web to allow using Wasm? or it will be automatic?

vbuberen:
I think this is not the right place for such discussion, but here is the place you should check for info on WASM support
https://docs.flutter.dev/platform-integration/web/wasm

amrgetment:
@vbuberen Thanks for clarification

PR #2481, opened 2023-12-29
fix(battery_plus): battery state always unknown (iOS only)

rostopira:
## Description

If battery percentage never read battery state is always unknown

## Related Issues

None

## Checklist

## Breaking Change

### DIFF
--- a/packages/battery_plus/battery_plus/ios/Classes/FPPBatteryPlusPlugin.m
+++ b/packages/battery_plus/battery_plus/ios/Classes/FPPBatteryPlusPlugin.m
@@ -70,7 +70,9 @@
 }
 
 - (NSString *)getBatteryState {
-  UIDeviceBatteryState state = [[UIDevice currentDevice] batteryState];
+  UIDevice *device = UIDevice.currentDevice;
+  device.batteryMonitoringEnabled = YES;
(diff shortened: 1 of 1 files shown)


vbuberen:
Thank you for your contribution.

Could you elaborate more on what is going on? Didn't understand from your description what is wrong there.

vbuberen:
@rostopira Could you reply to ☝🏻 ?

rostopira:
@vbuberen sorry, missed previous comment.

There is a steps to reproduce a bug
```dart
Future<void> main() async {
  final batteryState = await Battery().batteryState;
  print(batteryState);
}
```

If you run it on iOS without my patch you will get `BatteryState.unknown` instead of the real battery state.
It may be reproducible on macOS as well, but I didn't tried that.

vbuberen:
Thanks a lot for detailed explanations and reproducer. I was even able to reproduce the problem in the example app we have for the plugin. And indeed your patch works. Approving and will release this fix the upcoming week.

vbuberen (approved):
Thanks for your contribution.

vbuberen:
On MacOS couldn't get the same bug, so it seems applies only to iOS.

PR #2506, opened 2024-01-04
feat(sensors_plus)!: Add support for platform timestamp in event

Ortes:
## Description

Add timestamp to events from plateform instead of using DateTime.now() at event arrival that can suffer from jitter.
This can be useful to make fft or signal analysis.

I would start this PR as a discussion about the usefulness of this feature or how to implement it

I will provide the iOS implementation in the next few days

## Related Issues

- Fix https://github.com/flutter/flutter/issues/26078 event if it is not the same repo the problem is the same

## Checklist

## Breaking Change

- The API change but most sensor streams users won't need to change anything (if you don't create events like `UserAccelerometerEvent`)

### DIFF
--- a/packages/sensors_plus/sensors_plus/android/src/main/kotlin/dev/fluttercommunity/plus/sensors/StreamHandlerImpl.kt
+++ b/packages/sensors_plus/sensors_plus/android/src/main/kotlin/dev/fluttercommunity/plus/sensors/StreamHandlerImpl.kt
@@ -4,6 +4,7 @@ import android.hardware.Sensor
 import android.hardware.SensorEvent
 import android.hardware.SensorEventListener
 import android.hardware.SensorManager
+import android.os.SystemClock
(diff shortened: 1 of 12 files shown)


miquelbeltran:
I know this has been opened since January, I will see if I can get it merged.

Ortes:
I carefully rebased on main but not test it

miquelbeltran:
I've added the timestamp value to the Barometer events. As well, updated the example project to use the provided timestamp to calculate the refresh interval.

I did change the swift code, but I was on Linux so it is probably wrong, need to fix on my Mac.

Overall, the feature looks good to me. I still want to test on iOS when possible.

miquelbeltran:
Ok, so my iPhone 12 Mini doesn't have a barometer sensor, but compilation works, so.... 🤷 let's go!

miquelbeltran (approved):

PR #2518, opened 2024-01-10
fix(share_plus): Ensure subject is not null before calling putExtra(Intent.EXTRA_SUBJECT, subject)

jmsheff:
## Description

My friend who would share links via `share_plus` through Verizon's Message+ app, faced an issue where there would always be an additional null message alongside the shared link:

Screenshot 2024-01-10 at 3 14 34 PM

After doing a lot of digging and trying different things to try to fix this, I noticed that `Intent.EXTRA_SUBJECT` is always set to null when a subject is not specified. I suspected that this was the issue, given it was only happening for this one messaging platform, which may not filter out a null subject like most others do. Sure enough, after using my fork in place of the official `share_plus` package and having my friend send me a link again, this issue was fixed. I suspect that there may be other platforms affected by this as well.

## Related Issues

N/A

## Checklist

## Breaking Change


vbuberen (approved):
Thanks for your investigation and contribution. Looks good to me. 

Will release next week.

PR #2524, opened 2024-01-19
docs: Added ios simulators note for testing

reasje:
Since I didn't see that network info plus doesn't work in IOS simulators and I got null every time I made this PR to add that note to readme and avoid wasting time.

### DIFF
--- a/packages/network_info_plus/network_info_plus/README.md
+++ b/packages/network_info_plus/network_info_plus/README.md
@@ -61,6 +61,10 @@ To successfully get WiFi Name or Wi-Fi BSSID starting with Android 1O, ensure al
 >
 > This package does not provide the ACCESS_FINE_LOCATION nor the ACCESS_COARSE_LOCATION permission by default
 
+> **Note**
+>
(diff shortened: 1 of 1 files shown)


packages/network_info_plus/network_info_plus/README.md
```diff
@@ -61,6 +61,10 @@ To successfully get WiFi Name or Wi-Fi BSSID starting with Android 1O, ensure al
 >
 > This package does not provide the ACCESS_FINE_LOCATION nor the ACCESS_COARSE_LOCATION permission by default
 
+> **Note**
+>
+> Unlike Android, It is not possible to get wifi info in IOS simulators and It is going to be always `null`.
```
vbuberen, line 66: I would suggest to change it a bit for people who are not good with English.
```suggestion
> On iOS simulators wifi info will always return `null`.
> Android emulators have no such issue.
```

vbuberen (approved):
Thank you for your contribution. Indeed a valuable note.

PR #2578, opened 2024-02-06
chore(android_alarm_manager_plus): Partially switch CI runners to M1

vbuberen:
## Description

Switching to M1. Note, like the rest of plugins job to run Android integration tests doesn't migrate due to emulator Action not supporting ARMs at the moment.

### DIFF
--- a/.github/workflows/android_alarm_manager_plus.yaml
+++ b/.github/workflows/android_alarm_manager_plus.yaml
@@ -19,7 +19,7 @@ env:
 
 jobs:
   android_example_build:
-    runs-on: macos-13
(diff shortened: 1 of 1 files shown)


miquelbeltran:
I completely forgot, but there was a reason why we use MacOS images for Android tests? Rather than a Linux image.

miquelbeltran (approved):

vbuberen:
> but there was a reason why we use MacOS images for Android tests

They were faster at some point as far as I remember. And with M1 machines should also be faster than Ubuntu ones.

PR #2587, opened 2024-02-08
chore(share_plus): Workaround for failing integration tests

vbuberen:
## Description

This is the workaround that I mentioned during one of discussions. Found it here: https://github.com/flutter/flutter/issues/105913#issuecomment-1451585766

As build happens for integration tests anyway I think it makes no sense to build the app in a separate job, I believe, as ше ші just a waste of resources. Also, the drawback of this is that we won't be able to use M1 machine for this job till the issue with emulators which I mentioned in one of comments there is resolved.

In case it works I will open a similar PR for other plugins.

### DIFF
--- a/.github/workflows/share_plus.yaml
+++ b/.github/workflows/share_plus.yaml
@@ -22,19 +22,6 @@ env:
   PLUGIN_EXAMPLE_SCOPE: "*share_plus_example*"
 
 jobs:
-  android_build:
-    runs-on: macos-14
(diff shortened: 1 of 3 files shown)


miquelbeltran (approved):
All in favor of unifying the two CI jobs and improve stability

vbuberen:
No, this is wrong. But un fact I found an error already. I will apply some changes later.

luisredondo (approved):

Issue #2522, opened 2024-01-17
[Request]: README add note about VPN


AVaccar:
### Plugin

connectivity_plus

### Use case

I spent a lot of time debugging why the application on my phone wasn't detecting connectivity changes when i was enabling/disabling the wifi.  
In the end i had the duck duck go vpn enabled that was impeding network changes to be triggered.  
The moment i disabled it, the `Connectivity().onConnectivityChanged.listen` started working as expected.

### Proposal

Add a note in the readme about VPN next to the example of the stream subscription

vbuberen:
With [6.0.1](https://pub.dev/packages/connectivity_plus/changelog#601) release of the plugin it is not the case anymore as the result returned from the plugin now contains multiple connectivity types, so now your case should also work properly.

Closing as resolved.

Issue #2262, opened 2023-10-13
[Bug]: onConnectivityChanged not update . Your library doesn’t check the connection correctly when I’m using local network


duytient31:
### Platform

MACOS 13.3.1 (22E261)

### Plugin

connectivity_plus

### Version

4.0.2

### Flutter SDK

 3.10.5

### Steps to reproduce

Your library doesn’t check the connection correctly when I’m using local network

### Logs

```shell
no
```

### Flutter Doctor

```shell
[✓] Flutter (Channel stable, 3.10.5, on macOS 13.3.1 22E261 darwin-arm64, locale en-VN)
    • Flutter version 3.10.5 on channel stable at /Users/hodienthiet/fvm/versions/3.10.5
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 796c8ef792 (4 months ago), 2023-06-13 15:51:02 -0700
    • Engine revision 45f6e00911
    • Dart version 3.0.5
    • DevTools version 2.23.1

[!] Android toolchain - develop for Android devices (Android SDK version 30.0.3)
    • Android SDK at /Users/hodienthiet/Library/Android/sdk2
    ✗ cmdline-tools component is missing
      Run `path/to/sdkmanager --install "cmdline-tools;latest"`
      See https://developer.android.com/studio/command-line for more details.
    ✗ Android license status unknown.
      Run `flutter doctor --android-licenses` to accept the SDK licenses.
      See https://flutter.dev/docs/get-started/install/macos#android-setup for more details.

[✓] Xcode - develop for iOS and macOS (Xcode 14.3)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 14E222b
    • CocoaPods version 1.11.3

[✓] Android Studio (version 2020.3)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build 11.0.10+0-b96-7281165)

[✓] IntelliJ IDEA Ultimate Edition (version 2022.3)
    • IntelliJ at /Applications/IntelliJ IDEA.app
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
[...]

Issue #2142, opened 2023-09-08
[Bug]: Repeated Connectivity result - ConnectivityResult.mobile


raphire08:
### Platform

iOS 16

### Plugin

Connectivity Plus 

### Version

3.0.2

### Flutter SDK

3.7.12

### Steps to reproduce

Whenever connectivity result change we are logging it

```
Connectivity _connectivity = Connectivity();
 _connectivity.onConnectivityChanged
        .listen((ConnectivityResult result) {
     log('Connectivity subscription result: $result');
});
```

For many users we are getting repeated ConnectivityResult.mobile. Wanted to know what could cause this? Ideally it should be ConnectivityResult.mobile -> ConnectivityResult.none -> ConnectivityResult.mobile.

```
 
Info: 9/5/2023 2:05:01 PM - Skipping connectivity change : ConnectivityResult.mobile
 
Info: 9/5/2023 2:05:02 PM - Connectivity subscription result: ConnectivityResult.mobile
 
Info: 9/5/2023 2:05:02 PM - Skipping connectivity change : ConnectivityResult.mobile
 
Info: 9/5/2023 2:05:05 PM - Connectivity subscription result: ConnectivityResult.mobile
 
Info: 9/5/2023 2:05:05 PM - Skipping connectivity change : ConnectivityResult.mobile
 
Info: 9/5/2023 2:11:51 PM - Connectivity subscription result: ConnectivityResult.mobile
 
Info: 9/5/2023 2:11:51 PM - Skipping connectivity change : ConnectivityResult.mobile
 
Info: 9/5/2023 2:11:54 PM - Connectivity subscription result: ConnectivityResult.mobile
 
Info: 9/5/2023 2:11:54 PM - Skipping connectivity change : ConnectivityResult.mobile
 
Info: 9/5/2023 2:11:56 PM - Connectivity subscription result: ConnectivityResult.mobile
 
Info: 9/5/2023 2:11:56 PM - Skipping connectivity change : ConnectivityResult.mobile
 
Info: 9/5/2023 2:12:01 PM - Connectivity subscription result: ConnectivityResult.mobile
 
Info: 9/5/2023 2:12:01 PM - Skipping connectivity change : ConnectivityResult.mobile
 
Info: 9/5/2023 2:32:17 PM - Connectivity subscription result: ConnectivityResult.mobile
 
Info: 9/5/2023 2:32:17 PM - Skipping connectivity change : ConnectivityResult.mobile
 
[...]

Issue #2041, opened 2023-08-14
[Bug]: onConnectivityChanged - iOS not showing when WiFi is not available


Sun3:
### Platform

iOS 16.4 simulator

### Plugin

connectivity_plus

### Version

^4.0.2

### Flutter SDK

3.10.6

### Steps to reproduce

1. Run App with WiFi turned on
2. Turn off WiFi
3. ```onConnectivityChanged``` still shows 'wifi' instead of 'none' for internet connection

But if you do the following it works ok but will break again:

Run the App with WiFi turned off
Turn on the WiFi on Mac (it works)
Turn off the WifFi on Mac
onConnectivityChanged still shows 'wifi' instead of 'none' for internet connection

### Code Sample

```dart
connectionService.connectivity.onConnectivityChanged
    .listen((connectivityResult) {
  debugPrint('connectivityResult: ${connectivityResult.name}');
});
```
```


By the way, when you run it on an actual iPhone device it looks like it's working fine.



### Logs

```shell
Users/marco/StudioProjects/simpsons_viewer/build/ios/iphonesimulator/Runner.app
[ +255 ms] executing: /usr/bin/plutil -convert xml1 -o -
/Users/marco/StudioProjects/simpsons_viewer/build/ios/iphonesimulator/Runner.app/Info.plist
[   +4 ms] Exit code 0 from: /usr/bin/plutil -convert xml1 -o -
/Users/marco/StudioProjects/simpsons_viewer/build/ios/iphonesimulator/Runner.app/Info.plist
[        ] <?xml version="1.0" encoding="UTF-8"?>
           <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN"
           "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
           <plist version="1.0">
           <dict>
                <key>BuildMachineOSBuild</key>
                <string>22G74</string>
                <key>CADisableMinimumFrameDurationOnPhone</key>
                <true/>
                <key>CFBundleDevelopmentRegion</key>
                <string>en</string>
                <key>CFBundleDisplayName</key>
                <string>Simpsons Viewer</string>
                <key>CFBundleExecutable</key>
                <string>Runner</string>
                <key>CFBundleIcons</key>
                <dict>
                        <key>CFBundlePrimaryIcon</key>
                        <dict>
                                <key>CFBundleIconFiles</key>
                                <array>
                                        <string>AppIcon60x60</string>
                                </array>
[...]

PR #2599, opened 2024-02-15
feat(connectivity)!: support multiple ConnectivityResult values at the same time

suquant:
## Description

This PR updates the **connectivity_plus** plugin to handle multiple connectivity types simultaneously. Previously, **checkConnectivity** and **onConnectivityChanged** methods returned a single **ConnectivityResult** type, which could be either _wifi_, _cellular_, _ethernet_, _bluetooth_, _vpn_, or _none_. Due to this, on iOS/macOS platforms, the types _other_ & _vpn_ were never returned because the presence of _wifi_ or _cellular_ took precedence. This update changes the return type of these methods to a list of **ConnectivityResult**, allowing for a more accurate representation of the device's connectivity status, especially in scenarios where multiple types of connections are available or in use simultaneously.

## Related Issues

- Related #2522 
- Related #2262
- Related #2142
- Related #2041

## Checklist

## Breaking Change

Users of the connectivity_plus plugin will need to update their apps to handle the new list of **ConnectivityResult** types returned by **checkConnectivity** and **onConnectivityChanged** methods. This change allows for a more comprehensive understanding of the device's connectivity status, particularly in environments where multiple connectivity options are available or in use.


miquelbeltran (commented):
Thanks a lot for the contribution, I think this is a good start for a change that will fix several issues.

First, I'd like to ask you to properly fill the PR template, including the different checkboxes: https://github.com/fluttercommunity/plus_plugins/blob/main/.github/PULL_REQUEST_TEMPLATE.md 

We can see CI issues (analytics and tests not building) and those should be fixed as well.

Regarding the code, and before I go and do a more deep review, is there any reason why is preferred to handle the list of connectivity items as a coma-separated string rather than a List of Strings? e.g. in Connectivity.java, the MethodChannel, etc.

suquant:
@miquelbeltran 
> Regarding the code, and before I go and do a more deep review, is there any reason why is preferred to handle the list of connectivity items as a coma-separated string rather than a List of Strings? e.g. in Connectivity.java, the MethodChannel, etc.

Fair note, agreed!

packages/connectivity_plus/connectivity_plus/android/src/main/java/dev/fluttercommunity/plus/connectivity/Connectivity.java
```diff
@@ -23,58 +25,75 @@ public Connectivity(ConnectivityManager connectivityManager) {
...
+        break;
       case ConnectivityManager.TYPE_MOBILE:
       case ConnectivityManager.TYPE_MOBILE_DUN:
       case ConnectivityManager.TYPE_MOBILE_HIPRI:
-        return CONNECTIVITY_MOBILE;
+        types.add(CONNECTIVITY_MOBILE);
+        break;
       default:
-        return CONNECTIVITY_NONE;
+        types.add(CONNECTIVITY_NONE);
```
miquelbeltran, line 94: Would it be possible to simplify this by returning a `List.of(type)`? For what I understand, the array will always have just one item.

I haven't done Java in a long time, so I don't remember if that was possible :)

packages/connectivity_plus/connectivity_plus/android/src/main/java/dev/fluttercommunity/plus/connectivity/ConnectivityBroadcastReceiver.java
```diff
@@ -78,12 +78,12 @@ public void onCancel(Object arguments) {
   @Override
   public void onReceive(Context context, Intent intent) {
     if (events != null) {
-      events.success(connectivity.getNetworkType());
+      events.success(String.join(",", connectivity.getNetworkTypes()));
```
miquelbeltran, line 81: `success()` accepts an `Object`, so I think it would be possible to pass an array here, instead of a String. I'd prefer if we can keep the data as it is, rather than serializing/deserializing, if not strictly necessary. Would that be possible?

packages/connectivity_plus/connectivity_plus/ios/Classes/ReachabilityConnectivityProvider.swift
```diff
@@ -4,15 +4,18 @@ import Reachability
...
     switch reachability.connection {
     case .wifi:
-      return .wifi
+      return [.wifi]
     case .cellular:
-      return .cellular
+      return [.cellular]
     default:
-      return .none
+      return [.none]
```
miquelbeltran, line 18: I know this is unrelated to this PR, but, could be that the other connection types are missing in this switch statement?

packages/connectivity_plus/connectivity_plus/ios/Classes/ReachabilityConnectivityProvider.swift
```diff
@@ -4,15 +4,18 @@ import Reachability
 public class ReachabilityConnectivityProvider: NSObject, ConnectivityProvider {
   private var _reachability: Reachability?
 
-  public var currentConnectivityType: ConnectivityType {
-    let reachability = ensureReachability()
```
miquelbeltran, line 8: What is the reason here to change from using `ensureReachability()` to use the `_reachability` property?

miquelbeltran (commented):
Thanks again for the changes!

And btw, thanks for working on the Windows code as well, very few people want to touch that part of the project :)

I have left some comments, I need to also test this so it will take some time until we do the final approval, but so far it looks good.

I'd like to ask you too, that instead of force pushing changes, you can push new commits, so that I way I can track the PR changes since the last time I reviewed.

Thanks!

packages/connectivity_plus/connectivity_plus/android/src/main/java/dev/fluttercommunity/plus/connectivity/ConnectivityBroadcastReceiver.java
suquant: Agreed. I refactored the code to support the exchange list type between flutter and a host system in all integrations.

packages/connectivity_plus/connectivity_plus/ios/Classes/ReachabilityConnectivityProvider.swift
suquant: I leave the comment above the code, there are only 3 types in reachability module (wifi, cellular and none).

packages/connectivity_plus/connectivity_plus/ios/Classes/ReachabilityConnectivityProvider.swift
suquant: Returned it back, it my personal preferences, cause sometimes is confusing where you use guard construction in the code.

packages/connectivity_plus/connectivity_plus/android/src/main/java/dev/fluttercommunity/plus/connectivity/Connectivity.java
suquant: I thinks it's not so necessary cause return executed straight after switch statement.

suquant:
@miquelbeltran [android_integration_test (31)](https://github.com/fluttercommunity/plus_plugins/actions/runs/8003277978/job/21858253083?pr=2599#logs) need to be re-run, looks like there is some problem due to installation APK to emulator... Could you please rerun it, cause I don't have such permission?

packages/connectivity_plus/connectivity_plus/android/src/main/java/dev/fluttercommunity/plus/connectivity/Connectivity.java
miquelbeltran: what I mean is: 
- the code is creating an empty array on line 74
- then adding a single item in the switch statement
- then returning the array at the end after the switch.

And I'd prefer if the code could just call return inside the switch case with a `List.of(item)`. It is less code, and less prone to returning an empty array in case of a future code refactor.

miquelbeltran:
I'll try to run some tests this week and do a deeper code review. Thanks!

miquelbeltran:
Tested Android (also with VPN), iOS, MacOS, Linux and Chrome. I could not test Windows. Seems to work as expected, except in Linux and Web it did not report VPN correctly, but that's another issue.

packages/connectivity_plus/connectivity_plus/android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
+#Tue Feb 20 19:54:18 GST 2024
 distributionBase=GRADLE_USER_HOME
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.6.1-bin.zip
 distributionPath=wrapper/dists
-zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.2-bin.zip
 zipStoreBase=GRADLE_USER_HOME
+zipStorePath=wrapper/dists
```
miquelbeltran, line 6: This should be OK, but I am not sure if we want to upgrade gradle in this PR, or would be better to do it on a separate PR. Any thoughts @vbuberen

packages/connectivity_plus/connectivity_plus/darwin/Classes/ConnectivityPlugin.m
```diff
@@ -8,8 +8,8 @@
 #import "connectivity_plus-Swift.h"
 #endif
 
-@implementation ConnectivityPlusPlugin
+@implementation ConnectivityPlugin
 + (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar> *)registrar {
-  [SwiftConnectivityPlusPlugin registerWithRegistrar:registrar];
+  [SwiftConnectivityPlugin registerWithRegistrar:registrar];
 }
```
miquelbeltran, line 14: I think that this can cause troubles with people using the original connectivity plugin, if I remember correctly. I'd suggest not making this refactor now, and instead keeping the original file name.

packages/connectivity_plus/connectivity_plus/darwin/Classes/SwiftConnectivityPlugin.swift
```diff
@@ -2,9 +2,14 @@
...
 
+#if os(iOS)
 import Flutter
+#elseif os(macOS)
+import Cocoa
+import FlutterMacOS
+#endif
 
-public class SwiftConnectivityPlusPlugin: NSObject, FlutterPlugin, FlutterStreamHandler {
+public class SwiftConnectivityPlugin: NSObject, FlutterPlugin, FlutterStreamHandler {
```
miquelbeltran, line 12: Same about renaming this class, I'd avoid doing that in this PR

packages/connectivity_plus/connectivity_plus/pubspec.yaml
```diff
@@ -7,7 +7,9 @@ issue_tracker: https://github.com/fluttercommunity/plus_plugins/labels/connectiv
 
 environment:
   sdk: ">=2.18.0 <4.0.0"
-  flutter: ">=3.3.0"
+  # Flutter versions prior to 3.7 did not support the
+  # sharedDarwinSource option.
+  flutter: ">=3.7.0"
```
miquelbeltran, line 12: @vbuberen What do you think about increasing the min flutter version in order to share the macos and ios codebases? personally I am ok with it, but want to know your thoughts

packages/connectivity_plus/connectivity_plus_platform_interface/lib/src/utils.dart
```diff
@@ -1,22 +1,23 @@
...
-      return ConnectivityResult.mobile;
-    case 'vpn':
-      return ConnectivityResult.vpn;
-    case 'other':
-      return ConnectivityResult.other;
-    case 'none':
-    default:
-      return ConnectivityResult.none;
-  }
+/// Convert a comma-separated String to a list of ConnectivityResult values.
```
miquelbeltran, line 3: Small thing, but this comment seems incorrect, should be something like "Convert list of String connectivity states to list of ConnectivityResult values"

miquelbeltran (commented):
Good refactor unifying the iOS and MacOS codebases into a single one! We need to check if we can afford increasing the min-flutter sdk version.

I have added some other comments, otherwise it seems to be good.

Please, when you do updates, do not force-push, because if not I lose all the "viewed" history and need to review the whole PR again.

packages/connectivity_plus/connectivity_plus/android/gradle/wrapper/gradle-wrapper.properties
vbuberen: Yes, agree here. Even though we plan to move to 8.x in next release I would like it to not be mixed with the topic of this PR.

packages/connectivity_plus/connectivity_plus/darwin/Classes/ConnectivityPlugin.m
vbuberen: Yes, naming has to be reverted. Don't get why files were renamed here in the first place. Plugins are called `Plus Plugins`, so, please, don't touch naming for things like these ones.

packages/connectivity_plus/connectivity_plus/darwin/Classes/PathMonitorConnectivityProvider.swift
```diff
@@ -1,30 +1,34 @@
 import Foundation
 import Network
 
-@available(iOS 12, *)
+@available(iOS 12, macOS 10.14, *)
```
vbuberen, line 4: This makes no sense as with recent changes plugins already have MacOS 10.14 as min deployment target. Same applies to iOS 12.

packages/connectivity_plus/connectivity_plus/pubspec.yaml
vbuberen: I am Ok with that here, even though I don't like this merging of iOS and MacOS into one.

vbuberen (changes requested):
Thanks for the contribution, but I would ask to revert changes not related to the topic of the PR:
1. Revert Gradle wrapper version.
2. Revert naming changes for iOS/MacOS part. 

Also, please address the review feedback comments provided by Miguel.

packages/connectivity_plus/connectivity_plus/android/src/main/java/dev/fluttercommunity/plus/connectivity/Connectivity.java
suquant: @miquelbeltran Thanks for the suggestion! I aimed to optimize by avoiding extra array creation, but I'm open to your perspective. If you think changing it is best, I'm on board.

packages/connectivity_plus/connectivity_plus/android/gradle/wrapper/gradle-wrapper.properties
suquant: Agreed. I reverted it back.

packages/connectivity_plus/connectivity_plus/darwin/Classes/ConnectivityPlugin.m
suquant: Agreed. Reverted the name to the original one. Since we've merged ios & macos codebases into one then the macos plugin was renamed from `ConnectivityPlugn` to `ConnectivityPlusPlugin`. I hope it going to be ok.

packages/connectivity_plus/connectivity_plus/pubspec.yaml
suquant: I opted to maintain a single codebase for both platforms, aligning with best practices for Apple's products.

packages/connectivity_plus/connectivity_plus/darwin/Classes/PathMonitorConnectivityProvider.swift
suquant: I've removed that part to reflect this, especially since installation of this plugin is indeed confined to using a pod.

packages/connectivity_plus/connectivity_plus/darwin/connectivity_plus.podspec
```diff
@@ -14,12 +14,22 @@ Downloaded by pub (not CocoaPods).
...
+  s.ios.dependency 'Flutter'
+  s.osx.dependency 'FlutterMacOS'
   s.dependency 'ReachabilitySwift'
-  s.platform = :ios, '12.0'
+  s.platform = {
+    :ios => '12.0',
+    :osx => '10.14'
+  }
+  s.ios.deployment_target = '12.0'
+  s.osx.deployment_target = '10.14'
```
miquelbeltran, line 31: Do you know if it is necessary to define both `platform` and `deployment_target`? The docs only mention `deployment_target`: https://docs.flutter.dev/packages-and-plugins/developing-packages#shared-ios-and-macos-implementations

packages/connectivity_plus/connectivity_plus/pubspec.yaml
miquelbeltran: I am in favor of having a single code folder for both platform when possible, although it may not be the case for all plugins.

miquelbeltran (approved):
LGTM, waiting for Volodymyr to comment

vbuberen:
I see that the latest podspec comment is addressed as well.
So with that change all looks good to me. Let's merge it. 

Thanks a lot for this contribution, George.

We will need to do some updates to our docs, I believe, but let's do not in this PR.

suquant:
If all is good, I'm gonna squash all commits into a single one for better tracking and well organized. Thanks for review guys @miquelbeltran @vbuberen

vbuberen:
> If all is good, I'm gonna squash all commits into a single one for better tracking and well organized. Thanks for review guys @miquelbeltran @vbuberen

Don't do it. We squash it ourselves.

vbuberen (approved):

pamafe1976:
Hi, I just tested this update, and the behavior does not seem to be as intended. 
I always receive a List with 1 ConnectivityResult. 

If there is neither wifi or mobile I get **[ConnectivityResult.none]**
Then I turn on mobile and I get **[ConnectivityResult.mobile]**  =>OnConnectionChanged called 3 times
Then I turn on wifi and I get **[ConnectivityResult.wifi]** =>OnConnectionChanged called 4 times
If at this point I turn off mobile, onConnectionChanged is not called.
Then I turn on wifi and I get **[ConnectivityResult.none]** =>OnConnectionChanged called 1 time

What I expected was

If there is neither wifi or mobile I get **[ConnectivityResult.none]**
Then I turn on mobile and I get **[ConnectivityResult.mobile]**
Then I turn on wifi and I get **[ConnectivityResult.wifi, ConnectivityResult.mobile]**
Then turn off mobile and get **[ConnectivityResult.wifi]**
Then turn off wifiand get **[ConnectivityResult.none]**

Of course the callback should be called only 1 time on each change

vbuberen:
You are on Android, right? 

If so, check the API documentation as it is documented that on Android having both wifi and mobile returns only wifi (or, to be correct, the one that is used for network activity). For example, if you do the same test with VPN turned on you would see 2 results instead.

As to number of times you get the event - on Android we listen to onCapabilitiesChanged callback from NetworkManager as well to catch all events from network state changes. So it is how this callback works. It is up to you to distinct changes in the stream.

With that being said feel free to open a new issue instead of discussing here, but from what you described it is expected behavior from the plugin and part about currently active states on Android is described in the API documentation as well.

janniklind:
Thanks for pointing to the documentation :)
May I ask what is the reasoning behind never returning an empty list?

miquelbeltran:
Since the plugin returned a `ConnectivityResult.none` originally, we kept that.

We could add a convenience extension method that returns a single value and wraps the response with a `.first`, and/or one that just tells if there's connectivity or not.

PR #2612, opened 2024-02-23
feat(battery_plus): Added power source type of how the device is powered.

devaronius:
## Description

This PR updates the battery_plus plugin to handle power source type fetching on mobile. 
When you request the PowerSource type from the plugin it will try to get the source type from the intent battery changed extras, to indicate if the power source is AC, USB, Wireless or by battery.

### Motivation
We work with mobile scanner devices that run on Android and are socketed in Cradles for their charging and release to the users. 
There is no communication between the cradle and the scanner to detect if the scanner is in the cradle except for its power state. With the current version of the battery_plus package, we can detect whether the system is charging. But for development purposes, as an example, it is possible to also have a USB cable attached to the device.
This will break the charging logic, and we require a second parameter to determine if the power source is our USB cable, the cradle, or the battery. 

## Checklist

## Breaking Change

### DIFF
--- a/packages/battery_plus/battery_plus/android/src/main/kotlin/dev/fluttercommunity/plus/battery/BatteryPlusPlugin.kt
+++ b/packages/battery_plus/battery_plus/android/src/main/kotlin/dev/fluttercommunity/plus/battery/BatteryPlusPlugin.kt
@@ -58,6 +58,9 @@ class BatteryPlusPlugin : MethodCallHandler, EventChannel.StreamHandler, Flutter
                     result.error("UNAVAILABLE", "Battery level not available.", null)
                 }
             }
+            "getPowerSourceType" -> {
+                result.success(getPowerSourceType())
+            }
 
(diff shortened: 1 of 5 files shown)


devaronius:
> Thanks for your contribution. Will you also provide an implementation of this feature for iOS? If this feature is planned to cover only Android we won't be able to accept it as stated in our Contributor guide.

Hi @vbuberen, I planned to implement the iOS feature this weekend.
For Web and Windows, I do need to-do some research on how to accomplice this feature.

If macOS uses the same implementation as iOS, then this feature update will be planned for next weekend.

vbuberen:
Thanks for info. No rush from our side at all. Next release for plugins is planned to be somewhere in March, depending on when other desired features/fixes are ready. So take you time.
From my side I just wanted to be sure that not only Android will get this feature.

vbuberen:
@devaronius Any updates on this?

Issue #2601, opened 2024-02-15
[Request]: (device_info_plus): migrate to pkg:web from dart:html


dxvid-pts:
### Plugin

device_info_plus

### Use case

Flutter web apps are required to migrate from `dart:html` to `package:web` to use WASM.

### Proposal

This issue tracks the upgrade process for `device_info_plus`.

See https://docs.flutter.dev/platform-integration/web/wasm and https://pub.dev/packages/web for more information

amorenew:
any updates?

miquelbeltran:
> any updates?

https://github.com/fluttercommunity/plus_plugins/issues/2600#issuecomment-1952001128

amorenew:
@miquelbeltran wow, it closed same day I asked 😁😂

amorenew:
@miquelbeltran I made a PR with a different account to fix this issue

https://github.com/fluttercommunity/plus_plugins/pull/2622

miquelbeltran:
> @miquelbeltran I made a PR with a different account to fix this issue
> 
> #2622

Looks like the PR has nothing to do with this issue. Maybe you are mixing packages? The PR is modifying package_info_plus code and this ticket is about device_info_plus.

amorenew:
@miquelbeltran oh no I made a mistake, I will open an issue for package_info_plus
it is night here, I need to focus more 😭😱

miquelbeltran:
`package_info_plus` is already using web

miquelbeltran:
On `device_info_plus`: `Navigator.deviceMemory` seems to be missing in the current `web` implementation.

<img width="423" alt="image" src="https://github.com/fluttercommunity/plus_plugins/assets/2494376/bd42553b-c0f2-4e4d-b74a-abc69509fe7a">

References: 
- https://developer.mozilla.org/en-US/docs/Web/API/Navigator/deviceMemory
- https://api.dart.dev/stable/3.3.0/dart-html/Navigator/deviceMemory.html

PR #2622, opened 2024-02-28
feat(package_info_plus)!: Migrate to js_interop for conditional export

amrgetment:
## Description

I removed html for conditional import and used js_interop to support WASM
## Related Issues

-  Fix https://github.com/fluttercommunity/plus_plugins/issues/2623

## Checklist

## Breaking Change

### DIFF
--- a/packages/package_info_plus/package_info_plus/lib/package_info_plus.dart
+++ b/packages/package_info_plus/package_info_plus/lib/package_info_plus.dart
@@ -9,7 +9,7 @@ import 'package:package_info_plus_platform_interface/package_info_platform_inter
 
 export 'src/package_info_plus_linux.dart';
 export 'src/package_info_plus_windows.dart'
-    if (dart.library.html) 'src/package_info_plus_web.dart';
+    if (dart.library.js_interop) 'src/package_info_plus_web.dart';
 
 /// Application metadata. Provides application bundle information on iOS and
 /// application package information on Android.
--- a/packages/package_info_plus/package_info_plus/lib/src/package_info_plus_web.dart
+++ b/packages/package_info_plus/package_info_plus/lib/src/package_info_plus_web.dart
@@ -4,7 +4,7 @@ import 'package:flutter_web_plugins/flutter_web_plugins.dart';
 import 'package:http/http.dart';
 import 'package:package_info_plus_platform_interface/package_info_data.dart';
 import 'package:package_info_plus_platform_interface/package_info_platform_interface.dart';
-import 'package:web/helpers.dart' as web;
+import 'package:web/web.dart' as web;
 
 /// The web implementation of [PackageInfoPlatform].
 ///


miquelbeltran:
Why is the description saying "Fix https://github.com/fluttercommunity/plus_plugins/issues/2601" but the code is `package info plus`?

Can you provide more information in the PR description regarding this change as well?

amrgetment:
@miquelbeltran it is a mistake, actually I use package info plus not device info plus 
updated the description

packages/package_info_plus/package_info_plus/lib/src/package_info_plus_web.dart
```diff
@@ -4,7 +4,7 @@ import 'package:flutter_web_plugins/flutter_web_plugins.dart';
 import 'package:http/http.dart';
 import 'package:package_info_plus_platform_interface/package_info_data.dart';
 import 'package:package_info_plus_platform_interface/package_info_platform_interface.dart';
-import 'package:web/helpers.dart' as web;
```
miquelbeltran, line 7: I think you need to leave helpers.dart, because we want to support web 0.3

packages/package_info_plus/package_info_plus/lib/package_info_plus.dart
```diff
@@ -9,7 +9,7 @@ import 'package:package_info_plus_platform_interface/package_info_platform_inter
 
 export 'src/package_info_plus_linux.dart';
 export 'src/package_info_plus_windows.dart'
-    if (dart.library.html) 'src/package_info_plus_web.dart';
```
miquelbeltran, line 12: Can you explain this change?

miquelbeltran (commented):
This is a bit confusing.

The package already uses `web`, so the ticket you created seems to be wrong. Can you elaborate on that?

vbuberen:
`package_info_plus` is already migrated to `web` package and we need to support versions since `0.3.0`. 
So I don't see the reason to have this PR. Closing it as we are not going to accept it anyway.

vbuberen:
The migration happened a few months ago already https://github.com/fluttercommunity/plus_plugins/pull/2316

amrgetment:
@vbuberen according to this document you need to use js_interop for conditional import
https://dart.dev/interop/js-interop/package-web

image

so are you sure it is compatible with WASM with dart.library.html?

packages/package_info_plus/package_info_plus/lib/package_info_plus.dart
amrgetment: according to this document you need to use js_interop for conditional import to support WASM,
am I wrong?
https://dart.dev/interop/js-interop/package-web

packages/package_info_plus/package_info_plus/lib/src/package_info_plus_web.dart
amrgetment: oh, I will revert it back if the PR open again

amrgetment: reviewer made an alternative PR already

vbuberen:
I am sure, because if you would check the link right above your comment you would see that I created another PR with the same change, but took a few minutes to provide a proper description with links instead of what was provided in this PR. 
Instead, you are replying to something is already irrelevant, not to mention that tag somebody for no reason, considering that people who leave comments in issue/PR will get notifications anyway.

Screenshot 2024-02-29 at 12 44 30

amrgetment:
@vbuberen I replied to this 
> So I don't see the reason to have this PR
we have a different time zones so I replied late as I made this PR night and I checked it noontime

vbuberen:
I have just explained to you why you shouldn't tag people and you still continue 🤦🏻 

> we have a different time zones so I replied late as I made this PR night and I checked it noontime

It doesn't matter. In case PR had a proper description with links (like the one that I created instead) and didn't reference an issue `migrate to package:web` which is already resolved for this package, nobody would have any questions and the timezone wouldn't matter as reviewers would understand what is the reason for change.

vbuberen:
Next time please provide a clear and full description as people might not be in the context of what you have in your head. With that being said locking this discussion as there is nothing to discuss more.

PR #2624, opened 2024-02-29
feat(device_info_plus)!: migrate to package:web

### LINKED ISSUE #2601: [Request]: (device_info_plus): migrate to pkg:web from dart:html

miquelbeltran:
## Description

- Replace calls to `dart:html` by the equivalent with the `package:web`.
- Tested on Chrome + macOS
- Set web version to `>=0.5.0 <=0.6.0` to be compatible with any upcoming `web` release.
- Set min Dart sdk version to `3.3.0` in package, as it is the min version of `web` `0.5`. This is required since `Navigator` is only available from version `0.5`, so it is not possible to be compatible with older versions of `web`.
- Note: The property `deviceMemory` seems to be missing in the current web implementation. The corresponding ticket has been created in the `package:web` repo: https://github.com/dart-lang/web/issues/192 This property is nullable (`int?`) so it is not a blocker, however, users relying on this property will be affected. A solution is possible by creating our own `deviceMemory` extension.

## Related Issues

- Closes #2601 

## Checklist

## Breaking Change


vbuberen (approved):
Looks good to me.

Based on recent discussions we had, I think we should go moving to Dart 3.3 as min for all plugins and align on `web 0.5.0` as min version for consistency. Agree?

miquelbeltran:
Sounds good to me! I will merge this one now

PR #2652, opened 2024-03-02
feat(device_info_plus): LinuxDeviceInfo toString method

defuncart:
## Description

Adds `toString` for `LinuxDeviceInfo`. This is useful when debugging apps on Linux distros.

## Related Issues

N/A

## Checklist

- [X] I did not modify the `CHANGELOG.md` nor the plugin version in `pubspec.yaml` files.

## Breaking Change

### DIFF
--- a/packages/device_info_plus/device_info_plus/lib/src/model/linux_device_info.dart
+++ b/packages/device_info_plus/device_info_plus/lib/src/model/linux_device_info.dart
@@ -160,4 +160,9 @@ class LinuxDeviceInfo implements BaseDeviceInfo {
       'machineId': machineId,
     };
   }
+
+  @override
+  String toString() {
+    return 'LinuxDeviceInfo(name: $name, version: $version, id: $id, idLike: $idLike, versionCodename: $versionCodename, versionId: $versionId, prettyName: $prettyName, buildId: $buildId, variant: $variant, variantId: $variantId, machineId: $machineId)';
+  }
 }
--- a/packages/device_info_plus/device_info_plus/test/model/linux_device_info_test.dart
+++ b/packages/device_info_plus/device_info_plus/test/model/linux_device_info_test.dart
@@ -34,5 +34,26 @@ void main() {
         'machineId': 'machineId',
       });
     });
+
+    test('toString should return string representation', () {
+      final linuxDeviceInfo = LinuxDeviceInfo(
+        name: 'name',
+        version: 'version',
+        id: 'id',
+        idLike: ['idLike'],
+        versionCodename: 'versionCodename',
+        versionId: 'versionId',
+        prettyName: 'prettyName',
+        buildId: 'buildId',
+        variant: 'variant',
+        variantId: 'variantId',
+        machineId: 'machineId',
+      );
+
+      expect(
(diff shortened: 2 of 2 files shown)


packages/device_info_plus/device_info_plus/lib/src/model/linux_device_info.dart
```diff
@@ -160,4 +160,9 @@ class LinuxDeviceInfo implements BaseDeviceInfo {
       'machineId': machineId,
     };
   }
+
+  @override
+  String toString() {
+    return 'LinuxDeviceInfo(name: $name, version: $version, id: $id, idLike: $idLike, versionCodename: $versionCodename, versionId: $versionId, prettyName: $prettyName, buildId: $buildId, variant: $variant, variantId: $variantId, machineId: $machineId)';
```
miquelbeltran, line 166: Can you split this into a multi-line string?

miquelbeltran (commented):
Thanks for the contribution, and including a test as well!

I'd prefer if the long string could be split into several lines or a multiline string, as it will be easier to extend and review changes in the future.

Can you take a look at the Flutter analyze issues as well?

defuncart:
> I'd prefer if the long string could be split into several lines or a multiline string, as it will be easier to extend and review changes in the future.

The method was generated by [dart-data-class-generator](https://marketplace.visualstudio.com/items?itemName=dotup.dart-data-class-generator) and uses a standard format. How would you like it to be multi-line? Code of the form

```dart
    return '''LinuxDeviceInfo(
name: $name,
version: $version,
id: $id, 
idLike: $idLike, 
versionCodename: $versionCodename, 
versionId: $versionId, 
prettyName: $prettyName, 
buildId: $buildId, 
variant: $variant, 
variantId: $variantId, 
machineId: $machineId
)''';
```

would return a multi-line string, not a single line as expected in `toString`. To improve extensibility, `StringBuffer` could be considered, or alternatively  `data` could be used as follows

```dart
String toString() => 'LinuxDeviceInfo(${data.toString().replaceAll('}', '').replaceAll('{', '')})';
```

> Can you take a look at the Flutter analyze issues as well?

Done.

miquelbeltran:
That's right, I was thinking on a different approach, but since we already do similar `toString` methods in other parts of the projects, I guess that would be fine as well:
https://github.com/fluttercommunity/plus_plugins/blob/c9435836e8c7b354d5ca5383029e0172efe301d1/packages/package_info_plus/package_info_plus/lib/package_info_plus.dart#L119

packages/device_info_plus/device_info_plus/lib/src/model/linux_device_info.dart
miquelbeltran: As discussed, we leave this as is.

miquelbeltran (approved):

defuncart:
@miquelbeltran Seems the pipeline needs to be re-run, android integration test is failing & I cannot manually re-run. Probably a flaky issue as PR does not touch any android code.

miquelbeltran:
That's fine, no worries, the `android_integration_test (22)` is indeed flaky

PR #2668, opened 2024-03-10
fix(connectivity_plus): Return valid connection type when only one available

vbuberen:
## Description

I was working on updates for example app to show newly added multiple [connection types feature](#2599) in a more nice way and found a bug that when only one connection type is available there is an exception in the platform interface as returned result isn't a list, but a String:
Screenshot 2024-03-10 at 13 48 22

This PR fixes mentioned bug and also updates the example UI to look better:
<details>
<summary>Before</summary>

photo_2024-03-10 14 18 31
</details>

<details>
<summary>After</summary>

photo_2024-03-10 14 18 29
</details>

## Breaking Change

### DIFF
--- a/packages/connectivity_plus/connectivity_plus/example/lib/main.dart
+++ b/packages/connectivity_plus/connectivity_plus/example/lib/main.dart
@@ -92,11 +92,32 @@ class _MyHomePageState extends State<MyHomePage> {
   Widget build(BuildContext context) {
     return Scaffold(
       appBar: AppBar(
-        title: const Text('Connectivity example app'),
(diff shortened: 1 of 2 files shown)


miquelbeltran:
On which platform did you get this error? I would like to look it up and ensure that the stream always returns an array

miquelbeltran (approved):

vbuberen:
On Android

asonni:
I hope this issue will be fixed soon.

vbuberen:
> I hope this issue will be fixed soon.

Do you experience this issue? It never made it to the stable release of the plugin.

PR #2674, opened 2024-03-11
ci(all): Update Google Java Format tool

vbuberen:
## Description

Just found out that we had version 1.3 of Google Java Format tool, which was released in 2017. Updated to the latest one.
Probably, some new issues would appear with formatting as I expect rules changes during all these years.

## Breaking Change

### DIFF
--- a/.github/workflows/all_plugins.yaml
+++ b/.github/workflows/all_plugins.yaml
@@ -93,7 +93,7 @@ jobs:
       - name: "Install Tools"
         run: |
           ./.github/workflows/scripts/install-tools.sh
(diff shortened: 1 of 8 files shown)


miquelbeltran (approved):

vbuberen:
Fixed formatting issues which appeared from this update.

Also, had to bump Gradle version in `example` app of Android Alarm Manager because of using Android Gradle Plugin 8.3.0. It aligns with version compatibility table here: https://developer.android.com/build/releases/gradle-plugin#updating-gradle but something that I will try to not forget to mention in changelogs when it comes to release.
Screenshot 2024-03-12 at 14 18 49

@miquelbeltran FYI ☝🏻

PR #2709, opened 2024-03-16
feat(share_plus)!: Migrate to package:web

### LINKED ISSUE #2603: [Request]: (share_plus) Migrate to `package:web` to support WASM
### Plugin

share_plus

### Use case

Compiling to WASM

### Proposal

The short version:

Migrate usage of dart:html to package:web
Migrate usage of dart:js, dart:js_util, and package:js to dart:js_interop
This will allow compiling to WASM engine instead of just JavaScript on web.

Full migration guide: https://dart.dev/interop/js-interop/package-web

### LINKED ISSUE #2663: [Request]: In `share_plus` for web, use `navigator.canShare` method to check if files are shareable
### Plugin

share_plus

### Use case

See discussion in https://github.com/fluttercommunity/plus_plugins/issues/1643

Several users complain that the `shareFiles` and `shareXFiles` method fail in web with obscure errors.

### Proposal

The Navigator API provides a `canShare` method, which can be called with the list of files and will return true if the files are shareable: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/canShare

> The Navigator.canShare() method of the [Web Share API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Share_API) returns true if the equivalent call to [navigator.share()](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share) would succeed.

The idea would be that use this method to verify if the files are shareable, and if not throw an exception with an error similar to: "Files cannot be share, canShare returned false, check https://developer.mozilla.org/en-US/docs/Web/API/Navigator/canShare for more information"

koji-1009:
## Description

Migration to the `web` package and addition of override methods.

---

https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share#browser_compatibility

Chrome desktop doesn't support Share API on macOS.If someone want to run on macOS, please use Safari.
First, `flutter run` and check the service's url (ex. `http://localhost:55926/`). Then, open the url by Safari. 

---

https://koji-1009.github.io/plus_plugins/

I uploaded example app. The branch is [share_plus_web_demo](https://github.com/koji-1009/plus_plugins/tree/feat/share_plus_web_demo).
When I try to build on GitHub Actions, it fails at [odd points](https://github.com/koji-1009/plus_plugins/actions/runs/8304801875/job/22730899994). For this reason, we push the build on the local machine to the gh-pages branch.

## Related Issues

- Closes #2603
- Closes #2663

## Checklist

## Breaking Change

### DIFF
--- a/packages/share_plus/share_plus/example/pubspec.yaml
+++ b/packages/share_plus/share_plus/example/pubspec.yaml
@@ -24,4 +24,5 @@ flutter:
     - assets/flutter_logo.png
 
 environment:
-  sdk: '>=2.18.0 <4.0.0'
+  sdk: '>=3.3.0 <4.0.0'
+  flutter: '>=3.19.0'
--- a/packages/share_plus/share_plus/lib/share_plus.dart
+++ b/packages/share_plus/share_plus/lib/share_plus.dart
@@ -12,7 +12,7 @@ export 'package:share_plus_platform_interface/share_plus_platform_interface.dart
 
 export 'src/share_plus_linux.dart';
 export 'src/share_plus_windows.dart'
-    if (dart.library.html) 'src/share_plus_web.dart';
+    if (dart.library.js_interop) 'src/share_plus_web.dart';
 
 /// Plugin for summoning a platform share sheet.
 class Share {
--- a/packages/share_plus/share_plus/lib/src/share_plus_web.dart
+++ b/packages/share_plus/share_plus/lib/src/share_plus_web.dart
@@ -1,12 +1,16 @@
-import 'dart:html' as html;
+import 'dart:developer' as developer;
+import 'dart:js_interop';
 import 'dart:typed_data';
+import 'dart:ui';
 
-import 'package:flutter/widgets.dart';
 import 'package:flutter_web_plugins/flutter_web_plugins.dart';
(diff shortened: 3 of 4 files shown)


miquelbeltran:
Looking forward to this! I tried to work on it but couldn't figure out many things. Let me know if you need any assistance testing.

koji-1009:
Hi! Implementation is almost complete!

I was having trouble with an unexpected exception in chrome on macOS, but it may be a bug in dart.

https://github.com/dart-lang/web/issues/211

If we try-catch the `Object` instead of `DOMException`, it seems to work as expected in all browsers. However, this is an obvious bug and I would like to know what the dart team thinks.

koji-1009:
The Issue was due to my lack of confirmation.

If there are API differences between browsers, they should be checked with `NoSuchMethodError`; if the API exists, error handling can be done with `DOMException`.

If we try to handle API differences with `DOMException`, it succeeds in the debug build, but fails in the release build. Probably due to the minify process.

---

I will also check sensors_plus again with the releae build.

miquelbeltran (approved):
Tested on Chrome + Windows. Thanks for the effort and for providing a hosted working example!

PR #2711, opened 2024-03-16
build(android_intent_plus): Update to target and compile SDK 34 on Android

vbuberen:
## Description

Same as #2701 but for `android_intent_plus`. 

Additionally, fixed the example app as it didn't even build locally.

## Breaking Change

### DIFF
--- a/packages/android_intent_plus/android/build.gradle
+++ b/packages/android_intent_plus/android/build.gradle
@@ -46,14 +46,6 @@ android {
 }
 
 dependencies {
-    constraints {
-        implementation("org.jetbrains.kotlin:kotlin-stdlib-jdk7:1.8.22") {
-            because("kotlin-stdlib-jdk7 is now a part of kotlin-stdlib")
-        }
-        implementation("org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.8.22") {
-            because("kotlin-stdlib-jdk8 is now a part of kotlin-stdlib")
-        }
-    }
     implementation 'androidx.annotation:annotation:1.7.1'
     testImplementation 'junit:junit:4.13.2'
     testImplementation 'org.mockito:mockito-core:5.11.0'
--- a/packages/android_intent_plus/android/build.gradle
+++ b/packages/android_intent_plus/android/build.gradle
@@ -22,12 +22,12 @@ rootProject.allprojects {
 apply plugin: 'com.android.library'
 
 android {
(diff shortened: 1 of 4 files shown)


packages/android_intent_plus/android/build.gradle
```diff
@@ -46,7 +46,8 @@ android {
 }
 
 dependencies {
-    implementation 'androidx.annotation:annotation:1.7.0'
+    implementation(platform("org.jetbrains.kotlin:kotlin-bom:1.8.10"))
```
vbuberen, line 49: This is one of ways to resolve build errors due to different AndroidX packages having different Kotlin versions.

See https://youtrack.jetbrains.com/issue/KT-55297/kotlin-stdlib-should-declare-constraints-on-kotlin-stdlib-jdk8-and-kotlin-stdlib-jdk7

packages/android_intent_plus/example/android/app/src/main/java/io/flutter/plugins/androidintentexample/MainActivity.java
```diff
@@ -1,17 +1,22 @@
...
+import androidx.core.content.ContextCompat;
+
 import io.flutter.embedding.android.FlutterActivity;
 
 public class MainActivity extends FlutterActivity {
 
-  @Override
-  protected void onCreate(Bundle savedInstanceState) {
-    super.onCreate(savedInstanceState);
+    @SuppressLint("WrongConstant")
```
vbuberen, line 13: Here is the same thing: https://github.com/fluttercommunity/plus_plugins/pull/1811/files#r1193861724

Looks like Google folks don't really care about fixing this bug.

packages/android_intent_plus/android/build.gradle
vbuberen: That fix from above didn't work on CI, so replaced with another one, which I use in other projects: https://youtrack.jetbrains.com/issue/KT-54136/Duplicated-classes-cause-build-failure-if-a-dependency-to-kotlin-stdlib-specified-in-an-android-project#focus=Comments-27-6583109.0-0

vbuberen:
It is so strange that on CI build fails https://github.com/fluttercommunity/plus_plugins/actions/runs/8308234561/job/22738260821?pr=2711

While locally for me both options on declaring constraints for Kotlin version work fine.

miquelbeltran (approved):

PR #2712, opened 2024-03-16
build(share_plus): Update to target and compile SDK 34 on Android

vbuberen:
## Description

Same as #2701 but with a few additional fixes for the example app.

## Breaking Change

### DIFF
--- a/packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/Share.kt
+++ b/packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/Share.kt
@@ -65,21 +65,22 @@ internal class Share(
             }
         }
         // If we dont want the result we use the old 'createChooser'
-        val chooserIntent = if (withResult) {
-            // Build chooserIntent with broadcast to ShareSuccessManager on success
-            Intent.createChooser(
-                shareIntent,
-                null, // dialog title optional
-                PendingIntent.getBroadcast(
-                    context,
-                    0,
-                    Intent(context, SharePlusPendingIntent::class.java),
-                    PendingIntent.FLAG_UPDATE_CURRENT or immutabilityIntentFlags
-                ).intentSender
-            )
-        } else {
-            Intent.createChooser(shareIntent, null /* dialog title optional */)
-        }
+        val chooserIntent =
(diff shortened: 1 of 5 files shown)


packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/Share.kt
```diff
@@ -65,21 +65,22 @@ internal class Share(
...
-                    0,
-                    Intent(context, SharePlusPendingIntent::class.java),
-                    PendingIntent.FLAG_UPDATE_CURRENT or immutabilityIntentFlags
-                ).intentSender
-            )
-        } else {
-            Intent.createChooser(shareIntent, null /* dialog title optional */)
-        }
+        val chooserIntent =
+            if (withResult && Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP_MR1) {
```
vbuberen, line 69: Add version check to not see lint errors when working with Android part. 
This change influences nothing, because we have the same check in the method call handler class.

packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/Share.kt
```diff
@@ -123,21 +126,22 @@ internal class Share(
...
```
vbuberen, line 130: Added version check to not see lint errors when working with Android part. 
This change influences nothing, because we have the same check in the method call handler class.

packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/SharePlusPendingIntent.kt
```diff
@@ -28,6 +28,7 @@ internal class SharePlusPendingIntent: BroadcastReceiver() {
             intent.getParcelableExtra(Intent.EXTRA_CHOSEN_COMPONENT, ComponentName::class.java)
         } else {
             // Deprecated in API level 33
+            @Suppress("DEPRECATION")
```
vbuberen, line 31: Just to not see lint warnings

miquelbeltran (approved):

PR #2714, opened 2024-03-16
build(android_alarm_manager_plus): Update to target and compile SDK 34 on Android

vbuberen:
## Description

Moving the plugin to target and compile against SDK 34 (Android 14).

Decided to also rework the example app a bit to let users request SCHEDULE_EXACT_ALARM permission. When app has no such permission schedule alarm button is disabled and request permission buttons is enabled. When user provides the permission schedule button becomes enabled while request permission button disables.
Additionally improved texts to improve UX by explaining what is currently happening.

## Breaking Change

### DIFF
--- a/packages/android_alarm_manager_plus/example/android/app/build.gradle
+++ b/packages/android_alarm_manager_plus/example/android/app/build.gradle
@@ -26,13 +26,17 @@ apply plugin: 'kotlin-android'
 apply from: "$flutterRoot/packages/flutter_tools/gradle/flutter.gradle"
 
 android {
-    compileSdkVersion 33
+    compileSdk 34
 
     namespace 'com.example.example'
 
     compileOptions {
         sourceCompatibility JavaVersion.VERSION_17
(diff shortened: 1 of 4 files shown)


packages/android_alarm_manager_plus/example/android/app/build.gradle
```diff
@@ -35,6 +35,10 @@ android {
         targetCompatibility JavaVersion.VERSION_17
     }
 
+    kotlinOptions {
+        jvmTarget = 17
```
vbuberen, line 39: Had to do, so Android Studio doesn't pick 21 which is not supported yet but used Kotlin versions.
In fact, moving to Java 17 is something I want to do to the rest of plugins to be up to date. But such change will happen in another set of PRs.

vbuberen:
One more thing I would like to address in follow-up PRs is to update the documentation and README with Android 14 related behavior changes information

miquelbeltran (approved):

PR #2715, opened 2024-03-17
refactor(package_info_plus): Migrate Android example to use the new plugins declaration

vbuberen:
## Description

Considering that fact that some of Plug Plugins will require Flutter 3.19 as min version in their next release I decided to also address the deprecation in how we declare Android related plugins in example apps. Currently users would see warnings during example app builds like this one: https://github.com/fluttercommunity/plus_plugins/actions/runs/8308735389/job/22743183649#step:5:96

There is also an official guidance from the Flutter team on this topic: https://docs.flutter.dev/release/breaking-changes/flutter-gradle-plugin-apply

With this change our example apps will be more or less up to date and show a good example on how to apply the new approach to apps.

## Breaking Change

### DIFF
--- a/packages/package_info_plus/package_info_plus/example/android/build.gradle
+++ b/packages/package_info_plus/package_info_plus/example/android/build.gradle
@@ -5,12 +5,12 @@ allprojects {
     }
 }
 
-rootProject.layout.buildDirectory = '../build'
+rootProject.buildDir = '../build'
 subprojects {
-    project.layout.buildDirectory = "${rootProject.layout.buildDirectory}/${project.name}"
+    project.buildDir = "${rootProject.buildDir}/${project.name}"
     project.evaluationDependsOn(':app')
 }
 
 tasks.register("clean", Delete) {
-    delete rootProject.layout.buildDirectory
+    delete buildDir
 }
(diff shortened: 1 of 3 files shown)


packages/package_info_plus/package_info_plus/example/android/build.gradle
```diff
@@ -1,29 +1,16 @@
...
-
 allprojects {
     repositories {
         google()
         mavenCentral()
     }
 }
 
-rootProject.buildDir = '../build'
+rootProject.layout.buildDirectory = '../build'
```
vbuberen, line 8: This is not related to the guidance that I shared, but something that was deprecated in Gradle 8.2.
https://docs.gradle.org/current/userguide/upgrading_version_8.html#project_builddir

Considering that we require 8.4 now due to Android Gradle Plugin 8.3.0, I thought that it won't hurt to address this deprecation as well.

vbuberen:
Hmm, it is strange that such error happens as it works fine locally 🤔

packages/package_info_plus/package_info_plus/example/android/build.gradle
vbuberen: Reverted this change to stick with only changes from the guidance I shared. Hope it works now.

vbuberen: Looks like after reverting this change all works as expected.

miquelbeltran (approved):
Nice cleanup!

Issue #2653, opened 2024-03-02
[Announcement]: 📢 package web migration and releases


miquelbeltran:
This ticket contains up-to-date information regarding the state of the migration from `dart:html` to package `web`: https://pub.dev/packages/web

This is a live ticket and information posted here will be changing.

As always, we would like to remind you that this is a project run by volunteers in their free time, and ask you to be mindful when asking for updates. Repeated requests and general unprofessional behavior is considered a breach of our [Code of Conduct](https://github.com/fluttercommunity/plus_plugins/blob/main/CODE_OF_CONDUCT.md).

### When will be packages migrated?

**Update 2024-03-20**: All packages that have a web implementation have been migrated.

#### Packages migrated

- `package_info_plus`
- `device_info_plus`
- `connectivity_plus`
- `sensors_plus`
- `share_plus`
- `battery_plus`

#### Packages that don't require migration to `web`

- `android_alarm_manager_plus`
- `android_intent_plus`
- `network_info_plus`

All these packages are not supported on web.

#### Tickets

You can search in the repo for a specific package to see if it has been migrated, or if it still in progress:

- https://github.com/fluttercommunity/plus_plugins/issues/2600
- https://github.com/fluttercommunity/plus_plugins/issues/2601
- https://github.com/fluttercommunity/plus_plugins/issues/2602
- https://github.com/fluttercommunity/plus_plugins/issues/2603
- https://github.com/fluttercommunity/plus_plugins/pull/2316
- https://github.com/fluttercommunity/plus_plugins/pull/2625
- https://github.com/fluttercommunity/plus_plugins/pull/2605
- https://github.com/fluttercommunity/plus_plugins/pull/2624
- https://github.com/fluttercommunity/plus_plugins/pull/2697
- https://github.com/fluttercommunity/plus_plugins/pull/2720

### When will be packages released?

**Update 2024-03-20**: Releases should be going out this week, if everything is correct.

#### Dependency Overrides

If you have `web` package conflicts, you can try to override the version using "dependency override" https://dart.dev/tools/pub/dependencies#dependency-overrides, but there is no warranty that versions are compatible. **Do this under your own risk!**

```
dependency_overrides:
  web: ^0.5.0
```

#### Git Ref

You can also use a [git ref](https://dart.dev/tools/pub/dependencies#git-packages) to reference a specific commit in your pubspec.yml, so you don't have to wait for a package release. **Do this under your own risk!**

```
dependencies:
  package_info_plus:
[...]

PR #2720, opened 2024-03-17
feat(battery_plus)!: Migrate to package:web

koji-1009:
## Description

Migration to the `web` package and add missing API.

https://developer.mozilla.org/en-US/docs/Web/API/BatteryManager#browser_compatibility

BatteryManager API is only supported Chrome or Edge.
Safari and Firefox are not supported, so status is expected to be unknown.

## Related Issues

- Related #2653

## Checklist

## Breaking Change

### DIFF
--- a/packages/battery_plus/battery_plus/example/lib/main.dart
+++ b/packages/battery_plus/battery_plus/example/lib/main.dart
@@ -42,9 +42,24 @@ class _MyHomePageState extends State<MyHomePage> {
   @override
   void initState() {
     super.initState();
-    _battery.batteryState.then(_updateBatteryState);
-    _batteryStateSubscription =
-        _battery.onBatteryStateChanged.listen(_updateBatteryState);
+    try {
+      _battery.batteryState.then(
+        _updateBatteryState,
+        onError: (e) {
+          _showError('onError: batteryState: $e');
+          _updateBatteryState(BatteryState.unknown);
+        },
+      );
+      _batteryStateSubscription = _battery.onBatteryStateChanged.listen(
+        _updateBatteryState,
+        onError: (e) {
+          _showError('onError: onBatteryStateChanged: $e');
+          _updateBatteryState(BatteryState.unknown);
+        },
+      );
+    } on Error catch (e) {
+      _showError('catch: batteryState: $e');
+    }
   }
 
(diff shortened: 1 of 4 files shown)


JgomesAT:
when will be available a new version in pub.dev?

miquelbeltran:
> when will be available a new version in pub.dev?

Please check the pinned issue: https://github.com/fluttercommunity/plus_plugins/issues/2653

koji-1009:
I am working on a strange problem.

In the case of `flutter run` or `flutter run --release`, we can try-catch the `NoSuchMethodError` that occurs on `window.navigator.getWindow()` with `battery_plus_web.dart`. 

```dart
  bool get isSupported {
    try {
      web.window.navigator.getBattery();
      return true;
    } on NoSuchMethodError catch (_) {
      // flutter run mode, catch the exception
      return false;
    }
  }
```

However, if we have a `flutter build`, we cannot try-catch it in `battery_plus_web.dart`, but have to try-catch it in `main.dart` (in example).

```dart
  @override
  void initState() {
    super.initState();
    try {
      _battery.batteryState.then(_updateBatteryState);
      _batteryStateSubscription =
          _battery.onBatteryStateChanged.listen(_updateBatteryState);
    } catch (_) {
      // flutter build mode, catch the exception
    }
  }
```

koji-1009:
Tested on local server, no problems. Therefore, I believe the problem is caused by hosting on GitHub Pages.

https://stackoverflow.com/questions/65536213/how-to-open-compiled-flutter-web-index-html-in-browser-locally

I'll update the code base and update the demo. And after adding comments (just in case), I will open the PR!

koji-1009:
demo: https://koji-1009.github.io/plus_plugins/
pr: https://github.com/koji-1009/plus_plugins/pull/5

koji-1009:
Another option is to add a method that determines the browser and returns true only for chrome or edge.
This would be different from other plugin implementations, but it seems like it would avoid Exception, so it may be worth considering.

miquelbeltran:
Thank you Koji! This is already a big leap forward. We are going to be merging it for now, and our plan is to release all packages by Thursday, we are just waiting for another PR to land.

miquelbeltran (approved):

PR #2723, opened 2024-03-18
build(alarm_manager_plus)!: Target Java 17 on Android

vbuberen:
## Description

Targeting Java 17 to be up to date with ecosystem.
Considering that we are using Gradle 8.4 now no compatibility problems expected: https://docs.gradle.org/current/userguide/compatibility.html

Also, with this change users should not see such deprecation warnings anymore:
Screenshot 2024-03-16 at 15 05 19

Marking as a breaking change to highlight it in release notes.

## Breaking Change

### DIFF
--- a/packages/android_alarm_manager_plus/android/build.gradle
+++ b/packages/android_alarm_manager_plus/android/build.gradle
@@ -30,12 +30,12 @@ android {
     namespace 'dev.fluttercommunity.plus.androidalarmmanager'
 
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_1_8
-        targetCompatibility JavaVersion.VERSION_1_8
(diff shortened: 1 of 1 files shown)


miquelbeltran (approved):

mwk24:
Quick note here: this is the only plugin I've seen targeting 17 (usually it's 1.8), and it did break my project. Not sure whether really this is a flutter thing they should fix, but fyi.

lauroboeni:
> ly plugin I've seen targeting 17 (usually it's 1.8), and it did break my project. Not sure whether really this is a flutter thing they sho

having the same issue. Is there a work-around or do I need to reference an older version of share_plus?

vbuberen:
It is not an issue. Update your project the similar way as plugin updated in this PR and everything will work fine.

PR #2730, opened 2024-03-18
build(share_plus)!: Target Java 17 on Android

vbuberen:
## Description

Same as #2723 

## Breaking Change

### DIFF
--- a/packages/share_plus/share_plus/android/build.gradle
+++ b/packages/share_plus/share_plus/android/build.gradle
@@ -30,12 +30,12 @@ android {
     namespace 'dev.fluttercommunity.plus.share'
 
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_1_8
-        targetCompatibility JavaVersion.VERSION_1_8
+        sourceCompatibility JavaVersion.VERSION_17
+        targetCompatibility JavaVersion.VERSION_17
     }
 
     kotlinOptions {
-        jvmTarget = '1.8'
+        jvmTarget = 17
     }
 
     defaultConfig {
--- a/packages/share_plus/share_plus/example/android/app/build.gradle
+++ b/packages/share_plus/share_plus/example/android/app/build.gradle
@@ -30,8 +30,8 @@ android {
     namespace 'io.flutter.plugins.shareexample'
 
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_1_8
-        targetCompatibility JavaVersion.VERSION_1_8
+        sourceCompatibility JavaVersion.VERSION_17
+        targetCompatibility JavaVersion.VERSION_17
     }
 
     lintOptions {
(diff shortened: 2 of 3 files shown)


miquelbeltran (approved):

orkun1675:
Since this PR I'm getting the following error when trying to build for Android:

```
Execution failed for task ':share_plus:compileReleaseKotlin'.
> Inconsistent JVM-target compatibility detected for tasks 'compileReleaseJavaWithJavac' (17) and 'compileReleaseKotlin' (1.8).
```

Could you explain what kind of migration is neccesary?

vbuberen:
> Could you explain what kind of migration is neccesary?

You need to change your `build.gradle` like it was changed in this PR to specify Java 17 to use during compilation of Kotlin code. From your error it looks like you have `1.8` specified in `kotlinOptions` section of your `build.gradle` file.

> Do we need to upgrade to Gradle 8.4 even though it's in [beta](https://mvnrepository.com/artifact/com.android.application/com.android.application.gradle.plugin)?

You are confusing Android Gradle Plugin and Gradle itself. While Android Gradle Plugin 8.4 is indeed in beta, Gradle 8.4 itslef isn't in beta. We require Gradle wrapper version to be 8.4. Gradle wrapper is specified in files like this inside any project: https://github.com/fluttercommunity/plus_plugins/blob/main/packages/share_plus/share_plus/android/gradle/wrapper/gradle-wrapper.properties

orkun1675:
Thanks Volodymyr.

I have updated to Java 17 and Gradle Wrapper 8.4. I have kept Gradle Plugin at 8.2.2 due to a [bug with 8.3](https://issuetracker.google.com/issues/327696048).

Unfortunately this only replaced ther error with:

```
Execution failed for task ':games_services:compileReleaseKotlin'.
> Inconsistent JVM-target compatibility detected for tasks 'compileReleaseJavaWithJavac' (1.8) and 'compileReleaseKotlin' (17).
```

I'm not sure why, since the game_services plugin does not specify a Java or Kotlin version? [link](games_services/android/build.gradle)

As a final result, based on this [Stackoverflow post](https://stackoverflow.com/a/77861041/3531663), I added `kotlin.jvm.target.validation.mode = IGNORE` to my `gradle.properties` file. This doesn't sound great. But did fix the build issue.

vbuberen:
And what about Java compile options? Did you also change it to 17 like it is done in this PR? 
Both Java compile and Kotlin compile tasks should have the same Java target version.

These 2 lines are the ones responsible for Java version for Java compile: https://github.com/fluttercommunity/plus_plugins/pull/2730/files#diff-2607f8958e631dcb7f1ce17b341b8520cde253d7cde36a0bafc64d98676f4270R33

vbuberen:
What is `game_services` plugin? 🤔 Not sure I understood that line

orkun1675:
Yes, I did replaced the compile options with Java 17 as well. Here is the full `app/build.gradle` file: https://pastebin.com/ZQMH5NLf

And the settings.gradle file is: https://pastebin.com/WeXLHTab

Sorry for the broken link for the game_services plugin. I've updated it above.

Issue #2699, opened 2024-03-14
[Bug]: Null Package Version Web


EricPHassey:
### Platform

Web

### Plugin

package_info_plus

### Version

5.0.1

### Flutter SDK

3.19.3

### Steps to reproduce

Not sure where the issue lies.

When I build it locally, it works fine. When I upload it to Firebase Hosting, I get package is null. Appears that it can't find the version.json file? It's in my build/web folder.

Happy to give additional information if the issue lies with me or with Firebase Hosting.

### Logs

```shell
No logs due to running on Firebase Hosting
```

### Flutter Doctor

```shell
[✓] Flutter (Channel stable, 3.19.3, on macOS 14.2 23C64 darwin-x64, locale en-US)
    • Flutter version 3.19.3 on channel stable at /Users/preventamedical/Software/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision ba39319843 (7 days ago), 2024-03-07 15:22:21 -0600
    • Engine revision 2e4ba9c6fb
    • Dart version 3.3.1
    • DevTools version 2.31.1

[✗] Android toolchain - develop for Android devices
    ✗ Unable to locate Android SDK.
      Install Android Studio from: https://developer.android.com/studio/index.html
      On first launch it will assist you in installing the Android SDK components.
      (or visit https://flutter.dev/docs/get-started/install/macos#android-setup for detailed instructions).
      If the Android SDK has been installed to a custom location, please use
      `flutter config --android-sdk` to update to that location.


[✓] Xcode - develop for iOS and macOS (Xcode 15.2)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 15C500b
    • CocoaPods version 1.15.2

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[!] Android Studio (not installed)
    • Android Studio not found; download from https://developer.android.com/studio/index.html
      (or visit https://flutter.dev/docs/get-started/install/macos#android-setup for detailed instructions).

[✓] IntelliJ IDEA Ultimate Edition (version 2023.3.3)
    • IntelliJ at /Users/preventamedical/Applications/IntelliJ IDEA Ultimate.app
    • Flutter plugin version 77.3.1
    • Dart plugin version 233.13763.5

[✓] IntelliJ IDEA Ultimate Edition (version 2023.3)
[...]

Issue #2675, opened 2024-03-12
[Bug]: [package_info_plus] Web: Version info returns next version in server


rpaat:
### Platform

Web

### Plugin

package_info_plus

### Version

5.0.1

### Flutter SDK

3.19.2

### Steps to reproduce

1. Deploy app to hosting
2. Open app in browser
3. Bump app version
4. Deploy new version to hosting
5. Call PackageInfo.fromPlatform() in already running app and check app version in app
6. New upcoming version info from hosting is shown instead of the currently running app version

### Logs

```shell
NA
```

### Flutter Doctor

```shell
[√] Flutter (Channel stable, 3.19.2, on Microsoft Windows [Version 10.0.22631.3155], locale et-EE)
    • Flutter version 3.19.2 on channel stable at C:\flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 7482962148 (13 days ago), 2024-02-27 16:51:22 -0500
    • Engine revision 04817c99c9
    • Dart version 3.3.0
    • DevTools version 2.31.1                                                                                                                                                  09:17

[√] Windows Version (Installed version of Windows is version 10 or higher)

[√] Android toolchain - develop for Android devices (Android SDK version 29.0.3)
    • Android SDK at C:\Android\SDK
    • Platform android-29, build-tools 29.0.3
    • ANDROID_HOME = C:\Android\SDK
    • Java binary at: C:\Program Files\Android\Android Studio\jbr\bin\java
    • Java version OpenJDK Runtime Environment (build 17.0.6+0-b2043.56-9586694)
    • All Android licenses accepted.

[√] Chrome - develop for the web
    • Chrome at C:\Program Files (x86)\Google\Chrome\Application\chrome.exe

[√] Visual Studio - develop Windows apps (Visual Studio Community 2019 16.11.13)
    • Visual Studio at C:\Program Files (x86)\Microsoft Visual Studio\2019\Community
    • Visual Studio Community 2019 version 16.11.32413.511
    • Windows 10 SDK version 10.0.19041.0

[√] Android Studio (version 2022.2)
    • Android Studio at C:\Program Files\Android\Android Studio
    • Flutter plugin can be installed from:
       https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
       https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build 17.0.6+0-b2043.56-9586694)

[...]

PR #2733, opened 2024-03-19
feat(package_info_plus)!: Support multiple version.json locations

### LINKED ISSUE #2732: [Bug]: package_info_plus is not using the right URL to load the version.json file in web
### Platform

Any web browser

### Plugin

package_info_plus

### Version

5.0.1

### Flutter SDK

3.7.0 and above

### Introduction

As of Flutter 3.7.0, Flutter allow us to [pass an assetBase value](https://docs.flutter.dev/platform-integration/web/initialization#initializing-the-engine) when the Flutter web engine is initialized:

```js
_flutter.loader.loadEntrypoint({
    entrypointUrl: "https://www.my-app-domain.com/app/main.dart.js"
    onEntrypointLoaded: async function(engineInitializer) {
        let appRunner = await engineInitializer.initializeEngine({
            assetBase: "https://www.my-app-domain.com/app/"
        });
        appRunner.runApp();
    }
});
```

This allow us to load the application from a domain **my-business-domain.com** when the application is actually deployed in other one **my-app-domain.com**

### Bug

Having the next domains:

- **my-app-domain.com**: Serving the app files domain
- **my-business-domain.com**: Executing the app files domain

**package_info_plus** is failing because it is trying to load the file **version.json** using the second domain instead the first one

Other users are having the same issue: https://github.com/fluttercommunity/plus_plugins/issues/1409#issuecomment-1566603862

The problem of this issue is here: https://github.com/fluttercommunity/plus_plugins/blob/a3cd72f86ba47f43c507f8b83f89aac7519404de/packages/package_info_plus/package_info_plus/lib/src/package_info_plus_web.dart#L54

Instead to get the base URI using some Flutter API, it is being fetched using a web specific API, which returns the URL used in that moment by the browser, what as we could see previously, that URL could not be where the Flutter app is actually deployed.

### Code Sample

_No response_

### Logs

```shell
NA
```

### Flutter Doctor

```shell
NA
```

- [X] I did a `flutter clean`

AriasBros:
## Description

Currently, the `version.json` file is fetched using a web specific API to get the base URI. Because the base URI obtained with this API is always the URL from the browser, this could lead that some setups where the Flutter app is deployed and served using different domains and paths could fail.

This PR changes that replacing the web specific API with a Flutter API to get the right URL used to load the assets of the application.

## Related Issues

- Fix #2732

## Checklist

## Breaking Change


miquelbeltran:
Thanks for the pr! I was experimenting with the same yesterday, it's kind of weird that the framework doesn't provide a direct way to obtain the base URLs. As pointed out, reading the base URL from the browser window is not exactly right in all cases.

My concern is if this would be wrong, given the case a user setups a different assets' URL than the entry point where the app is installed.

e.g.

```
_flutter.loader.loadEntrypoint({
    entrypointUrl: "https://www.app-domain-1.com/app/main.dart.js"
    onEntrypointLoaded: async function(engineInitializer) {
        let appRunner = await engineInitializer.initializeEngine({
            assetBase: "https://www.app-domain-2.com/myappcontent/"
        });
        appRunner.runApp();
    }
});
```

Where would we expect the `version.json` to be found? in domain 1 or 2?

So, we can have three URLs:

- The browser URL
- The entry point URL (where the Dart JS code is)
- The Assets URL

Another option is to use it as fallback, say, check if the version.json exists on assets base URL, if not, check on browser URL. This could be done together with also providing a custom param.

Allowing to do:

1. If a custom version.json path exists, use that.
2. If not, check in the assets' base URL.
3. If not, use browser URL.

miquelbeltran:
Hi @AriasBros I am thinking on taking your commit https://github.com/fluttercommunity/plus_plugins/pull/2733/commits/860b8ba4b3f78fa7db21fd0bb3635de0afaf47c9 and creatting a new PR that would support custom version.json url + fallback to browser baseUrl.

Do you mind if I get your commit and to that?

AriasBros:
Hi @miquelbeltran,

Sorry the delay to answer you. I agree with your proposed solution, but these days my company changed my priorities and this one was down a little in the list, but I think I will reach this task today in a few hours.

I would like to send an update to this PR today or the monday.

If there is hurry to fix this issue just let me know and in that case of course you can use my commit, no problem with that.

miquelbeltran:
Let's do it in this PR then, I will push some changes so you can review them, still WIP as the documentation should be updated as well.

miquelbeltran:
Actually, I will just restore back to the original state of your PR, I will experiment on my own branch.

packages/package_info_plus/package_info_plus/lib/package_info_plus.dart
```diff
@@ -33,12 +33,54 @@ class PackageInfo {
...
+  ///         }
+  ///     });
+  ///     ```
+  ///
+  ///     For more information about the Flutter Web Engine initialization see here:
+  ///     https://docs.flutter.dev/platform-integration/web/initialization#initializing-the-engine
+  ///
+  ///   * Finally, if none of the previous locations return the `version.json` file,
+  ///     the package will use the browser window base URL to resolve its location.
+  static Future<PackageInfo> fromPlatform({String? baseUrl}) async {
```
miquelbeltran, line 75: Would it be possible to pass here the full `version.json` path instead of the `baseUrl`, or said differently, is there a reason why the `baseUrl` is preferred?

I know that we can reuse the `versionJsonUrl` and it also adds the "cache buster", but I am thinking of giving full flexibility, so users can provide the `Uri` directly as it will give them more control.

miquelbeltran (commented):
So far the implementation is looking good.

I've left a question regarding the custom baseUrl.

I'll do some tests when I got the time.

We also should add integration tests. Testing the custom version.json uri should be easy (like done here: https://github.com/fluttercommunity/plus_plugins/blob/095499d2925dfaa162eacdd66c51df106fdd295b/packages/package_info_plus/package_info_plus/example/integration_test/package_info_plus_web_test.dart#L83-L100). Testing the assets' manager may not be as easy without allowing to mock it from the outside, but it is something we can do at the end.

AriasBros:
> We also should add integration tests. Testing the custom version.json uri should be easy. Testing the assets' manager may not be as easy without allowing to mock it from the outside, but it is something we can do at the end.

Yes, it is the next thing I want to work on.

packages/package_info_plus/package_info_plus/lib/package_info_plus.dart
AriasBros: Well, I feel like the right way is to allow only to indicate the base URL where the `version.json` is located. It is like less implementation details dependent and it is consistent with the other 2 ways the package gets the version file.

On the other hand, the `baseUrl` is giving you the same flexibility as the full `Uri`, excepting the case you want to rename the file, which it is something that I don't feel we should support. `version.json` is a private artifact as far as I know.

miquelbeltran: Yes, I see both points. My idea on providing the version.json URI directly was mostly because we had complaints regarding the cache-busting mechanism, so I thought that with the full path they can decide if they want to apply any cache busting query params or not.

Anyway, let's go with your implementation and see how the feedback goes!

miquelbeltran:
I'll mark the PR as breaking, just because the platform interface signature also changed, and I want to avoid version conflicts when pulling dependencies. Technically, this could pass as a non-breaking for package users, but I want to be on the safe side.

miquelbeltran (approved):
Thank you, David! I think this is good to go for now. Let me know if you still want to do some more changes.

packages/package_info_plus/package_info_plus/lib/package_info_plus.dart
AriasBros: Yes, I saw that issue https://github.com/fluttercommunity/plus_plugins/issues/2675

Anyway, the source of that problem I think is the nature of the web implementation, that is dependent of an external resource that needs to be loaded with a fetch request (GET version.json) and as all web resources, it has its own cache.

You can remove the buster cache, but then you will depend of the browser cache invalidation. Your `main.dart.js` and your `version.json` can be invalidated in different moments, so you will be in the same situation.

And this only thinking in the browser cache invalidation, but it could be other cache layers, like for example an AWS CloudFront cache invalidation layer, or others.

It is a complicated situation, to be honest.

We could add a new parameter to opt out the cache buster. Maybe that could mitigate the issue, but in my view, it will be there yet, just happening sometimes, instead always.

Or, we could implement a solution no dependent of the `version.json` file. For example, in the web build process of the app, the package could inject the same information that we can find in the `version.json` file in a dart file. In that way, you can know exactly which version of the app you are seeing in your browser, because the information it is in the same `main.dart.js` of your app.

Of course, this is just a theoretical solution, I don't know if it could be implemented even.

I will put this comment in the original issue, so we keep the discussion there if you agree.

miquelbeltran: Yep, I am happy to explore alternatives to the `version.json` file. We can keep the discussion on that ticket.

AriasBros:
> Thank you, David! I think this is good to go for now. Let me know if you still want to do some more changes.

Thanks @miquelbeltran 

From my side I don't see nothing else to do. Maybe update online documentation, but I am not sure if I should do it in this same PR or that follows other procedures. Just let me know about it, I am available for any change that this PR could need.

AriasBros:
> I'll mark the PR as breaking, just because the platform interface signature also changed, and I want to avoid version conflicts when pulling dependencies. Technically, this could pass as a non-breaking for package users, but I want to be on the safe side.

If it is necessary, I could split this PR in 2:
 * One adding only the implementation about to use the AssetManager base URL, which it should not be a breaking change.
 * And then other with the changes related to add the custom base URL parameter.

Anyway, in this last commit that I sent, I added a new parameter to the plugin constructor to allow testing the AssetManager. I understand that that constructor is a private API of the package, but well, maybe it could be considered a breaking change too.

miquelbeltran:
Thanks for including a test with the `AssetManager` as well!

Having a breaking change is not an issue, it will simply go into a new major version when we generate a new release, which I prefer anyway. It may take a while until we release the package since we did a big release very recently, but this will be available to anyone who wants to try using a git ref dependency.

I'm specially interested if people in https://github.com/fluttercommunity/plus_plugins/issues/2699 can give it a try.

Thanks again for the work! I will merge now, any further improvements we can do in a new PR.

PR #2734, opened 2024-03-20

vbuberen:
## Description

Release notes for all packages. I did some manual adjustments instead of using just Melos generated part to provide more clarity on WASM availability and requirements that plugins now have.

Note, for some plugins I had to do versions like `4.0.1` instead of `4.0.0` because last year there was a problem with one of releases after which I had to retract those releases. But since registry already contains `4.0.0` I can't re-release `4.0.0` again. Thus, had to release a few plugins as `x.0.1` version. 

All plugins versions successfully validated with `melos publish`

### DIFF
--- a/packages/sensors_plus/sensors_plus/CHANGELOG.md
+++ b/packages/sensors_plus/sensors_plus/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 > Note: This release has breaking changes.
 
-In this release plugin migrated migrated to package:web, meaning that it now supports WASM!
+In this release plugin migrated migrated from dart:html to js_interop, meaning that it now supports WASM!
 
 Plugin now requires the following:
 - Flutter >=3.19.0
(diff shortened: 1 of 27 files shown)


packages/sensors_plus/sensors_plus/CHANGELOG.md
```diff
@@ -1,3 +1,23 @@
+## 5.0.0
+
+> Note: This release has breaking changes.
+
+In this release plugin migrated migrated to package:web, meaning that it now supports WASM!
```
miquelbeltran, line 5: I think the plugin doesn't depend on package:web at all, instead it uses dart:js_interop directly. It's a small difference, doesn't matter much, just in case you want to mention it.

miquelbeltran (approved):
Left a couple of comments on the changelogs, just in case you want to fix them before proceeding. Otherwise, the whole release looks good!

miquelbeltran (approved):

PR #2742, opened 2024-03-21
refactor(share_plus): Migrate Android example to use the new plugins declaration

vbuberen:
## Description

Updating example app according to https://docs.flutter.dev/release/breaking-changes/flutter-gradle-plugin-apply
Similar to #2715

## Breaking Change

### DIFF
--- a/packages/package_info_plus/package_info_plus/example/android/settings.gradle
+++ b/packages/package_info_plus/package_info_plus/example/android/settings.gradle
@@ -23,5 +23,3 @@ plugins {
 }
 
 include ":app"
-
-rootProject.name = 'package_info_plus'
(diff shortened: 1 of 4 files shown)


packages/package_info_plus/package_info_plus/example/android/settings.gradle
```diff
@@ -23,5 +23,3 @@ plugins {
 }
 
 include ":app"
-
-rootProject.name = 'package_info_plus'
```
vbuberen, line 27: Typo during work on #2715 
This line doesn't have to be there.

miquelbeltran (approved):

PR #2743, opened 2024-03-21
refactor(sensors_plus): Migrate Android example to use the new plugins declaration

vbuberen:
## Description

Similar to #2715 and #2742 

## Breaking Change

### DIFF
--- a/packages/sensors_plus/sensors_plus/example/android/app/build.gradle
+++ b/packages/sensors_plus/sensors_plus/example/android/app/build.gradle
@@ -1,14 +1,14 @@
+plugins {
+    id "com.android.application"
+    id "dev.flutter.flutter-gradle-plugin"
+}
+
 def localProperties = new Properties()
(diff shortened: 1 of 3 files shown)


packages/sensors_plus/sensors_plus/example/android/app/build.gradle
```diff
@@ -73,5 +65,4 @@ dependencies {
     testImplementation 'junit:junit:4.13.2'
     androidTestImplementation 'androidx.test:runner:1.5.2'
     androidTestImplementation 'androidx.test.espresso:espresso-core:3.5.1'
-    implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
```
vbuberen, line 76: It seems we don't really need to have Kotlin and its plugin in the example app for now, so removed it.

miquelbeltran (approved):

PR #2746, opened 2024-03-21
docs(connectivity_plus): Improve documentation onConnectivityChanged method

miquelbeltran:
## Description

Following the discussion here: https://github.com/fluttercommunity/plus_plugins/pull/2599#issuecomment-2013129686

The goal of the plus plugins are to be as truth to the underlying platform as possible, while providing an easy way to access for all Flutter compatible platforms.

This PR adds a bit of context to the `onConnectivutyChanged` method, explaining why the `Stream` may emit multiple values.

Not that people ever reads documentation, but maybe someone will find it useful :)

Any comments are welcome.

## Related Issues

- Related: #2599

## Checklist

## Breaking Change

### DIFF
--- a/packages/connectivity_plus/connectivity_plus/lib/connectivity_plus.dart
+++ b/packages/connectivity_plus/connectivity_plus/lib/connectivity_plus.dart
@@ -34,11 +34,16 @@ class Connectivity {
     return ConnectivityPlatform.instance;
   }
 
-  /// Fires whenever the connectivity state changes.
+  /// Exposes connectivity update events from the platform.
   ///
   /// On iOS, the connectivity status might not update when WiFi
   /// status changes, this is a known issue that only affects simulators.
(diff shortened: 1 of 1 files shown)


vbuberen:
Thanks for taking care of this.

I had some thoughts on it and feel that maybe we could consider doing filtering with some distinct method? I don't think it should lead to some missed connection types changes being missed. I am in sort of doubtful state if it is worth it to send all events, especially duplicated ones.

vbuberen (approved):

miquelbeltran:
Yes, I cannot think a case where a dev would want to get all the events and not just distinct.

We should also change from a List to a Set, to ensure the order of the data doesn't matter when calculating the distinct.

Kind of related, I'd like to add an assert as well to ensure that the returned list/set always contains at least one item and cannot be empty, since I've found myself having to check "if contains None or isEmpty" every time.

PR #2770, opened 2024-03-30
docs(connectivity_plus): Update README for more clarity

vbuberen:
## Description

Revisited the README after looking through https://github.com/fluttercommunity/plus_plugins/issues/2041 issue. Re-phrased a few statements to bring some more clarity and did minor typo/formatting fixes.

## Breaking Change

### DIFF
--- a/packages/connectivity_plus/connectivity_plus/README.md
+++ b/packages/connectivity_plus/connectivity_plus/README.md
@@ -14,7 +14,9 @@ This plugin allows Flutter apps to discover network connectivity types that can 
 
 > **Note**
 >
-> On Android, this does not guarantee connection to Internet. For instance, the app might have wifi access but it might be a VPN or a hotel WiFi > with no access.
+> You should not rely on the current connectivity status to decide whether you can reliably make a network request. Always guard your app code against timeouts and errors that might come from the network layer.
+> Connection type availability does not guarantee that there is an Internet access. For example, the plugin might return Wi-Fi connection type, but it might be a connection with no Internet access due to network requirements (like on hotel Wi-Fi networks where user often needs to go through a captive portal to authorize first).
+>
 
 ## Platform Support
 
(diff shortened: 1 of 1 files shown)


packages/connectivity_plus/connectivity_plus/README.md
```diff
@@ -55,10 +57,6 @@ if (connectivityResult.contains(ConnectivityResult.mobile)) {
 }
 ```
 
-> **Note**
->
-> You should not be using the current network status for deciding whether you can reliably make a network connection. Always guard your app code against timeouts and errors that might come from the network layer.
-
```
vbuberen, line 61: I don't think that we need 2 notes, so merged them together and put on top of the file.

packages/connectivity_plus/connectivity_plus/README.md
```diff
@@ -84,16 +82,13 @@ dispose() {
...
+Connectivity changes are no longer communicated to Android apps in the background starting with Android O (8.0). You should always check for connectivity status when your app is resumed. The broadcast is only useful when your application is in the foreground.
 
 ### iOS
 
-On iOS, the connectivity status might not update when WiFi status changes, this is a known issue that affects mainly simulators.
-
-Starting on iOS 12, the implementation uses `NWPathMonitor` to obtain the enabled connectivity types.
-
-We noticed that this observer can give multiple or unreliable results. For example, reporting connectivity "none" followed by connectivity "wifi" right after reconnecting.
+On iOS simulators, the connectivity types stream might not update when Wi-Fi status changes. This is a known issue.
```
vbuberen, line 89: Changed to simulators only because I believe that this note was created before `NWPathMonitor` usage, which brings results better, even though with some issues noted below. 

Considering that iOS 12 is the minimum version now rephrased to only mention simulators as can only reproduce the issue on simulators only with plugin version 6.0.1.

miquelbeltran (approved):

PR #2813, opened 2024-04-04
refactor(connectivity_plus): remove ReachabilityConnectivityProvider and ReachabilitySwift dependency

### LINKED ISSUE #2779: [Request]: Remove Reachability implementation from connectivity_plus ios & macos (darwin)
### Plugin

connectivity_plus

### Use case

The plugin code contains two implementations, one using `NWPathManager` and one using a 3rd party library `Reachability`.

The plugin will always use `NWPathManager` on iOS 12 and macOS 10.14.

At the same time, the minimum supported iOS version is 12, and the minimum supported macOS version is 10.14. So the `Reachability` implementation would never be used.

### Proposal

Remove the `Reachability` implementation, cleanup the dependency too and simplify the code to always use the `NWPathManager`.

miquelbeltran:
## Description

- The min OS versions used are iOS: 12 and macOS: 10.14.
- At the same time, the min OS versions to use the `NWPathProvider` implementation are also 12 and 10.14.
- The old implementation that users `ReachibilitySwift` is not being used because the app will always use the `NWPathProvider` implementation.

## Related Issues

- Closes #2779 

## Checklist

## Breaking Change


vbuberen (approved):
Nice catch. I remember how this thing caused issues to us some time ago. Glad to see it going.

PR #2954, opened 2024-05-22
 Android system cannot obtain AndroidID

kang-breakfly:
## Description
Aiming at the problem that some phones of the Android system cannot obtain AndroidID or the user reinstalls AndroidID after uninstalling, the plugin makes SPUtil local storage, which will first take the local storage value after uninstalling and reinstalling 
For the IOS system, users will get a different UUID each time they uninstall and reinstall. The solution to this problem is to store the UUID obtained for the first time through the keychain.

### DIFF
--- a/packages/android_intent_plus/android/build.gradle
+++ b/packages/android_intent_plus/android/build.gradle
@@ -50,5 +50,5 @@ dependencies {
     testImplementation 'junit:junit:4.13.2'
     testImplementation 'org.mockito:mockito-core:5.11.0'
     testImplementation 'androidx.test:core:1.5.0'
-    testImplementation 'org.robolectric:robolectric:4.12.1'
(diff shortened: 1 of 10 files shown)


kang-breakfly:
Aiming at the problem that some phones of the Android system cannot obtain AndroidID or the user reinstalls AndroidID after uninstalling, the plugin makes SPUtil local storage, which will first take the local storage value after uninstalling and reinstalling
For the IOS system, users will get a different UUID each time they uninstall and reinstall. The solution to this problem is to store the UUID obtained for the first time through the keychain.

miquelbeltran:
Android ID was removed on purpose: https://github.com/fluttercommunity/plus_plugins/pull/918 please check that PR to learn why.

Besides, your PR contains unrelated code. I did not review it, but I want to assume good faith. In any case, I am closing this PR.

PR #2975, opened 2024-05-31
fix(package_info_plus): Address changed fields nullability on Android 15 (API 35)

ChristianEdwardPadilla:
## Description

This plugin fails to compile with the latest nullability signatures for Android 35

The compilation error messages are:

```
PackageInfoPlugin.kt:45:56: error: [UNSAFE_CALL] Only safe (?.) or non-null asserted (!!.) calls are allowed on a nullable receiver of type ApplicationInfo?
                    put("appName", info.applicationInfo.loadLabel(packageManager).toString())
                                                       ^
PackageInfoPlugin.kt:47:36: error: [TYPE_MISMATCH] Type mismatch: inferred type is String? but String was expected
                    put("version", info.versionName)
                                   ^^^^^^^^^^^^^^^^
PackageInfoPlugin.kt:108:51: error: [SMARTCAST_IMPOSSIBLE] Smart cast to 'Array<(out) Signature!>' is impossible, because 'packageInfo.signatures' is a mutable property that could have been changed by this time
                if (signatures.isNullOrEmpty() || packageInfo.signatures.first() == null) {
```

## Checklist

## Breaking Change

### DIFF
--- a/packages/package_info_plus/package_info_plus/android/src/main/kotlin/dev/fluttercommunity/plus/packageinfo/PackageInfoPlugin.kt
+++ b/packages/package_info_plus/package_info_plus/android/src/main/kotlin/dev/fluttercommunity/plus/packageinfo/PackageInfoPlugin.kt
@@ -44,7 +44,7 @@ class PackageInfoPlugin : MethodCallHandler, FlutterPlugin {
                 infoMap.apply {
                     put("appName", info.applicationInfo?.loadLabel(packageManager)?.toString() ?: "")
                     put("packageName", applicationContext!!.packageName)
-                    put("version", info?.versionName ?: "")
+                    put("version", info.versionName ?: "")
                     put("buildNumber", getLongVersionCode(info).toString())
                     if (buildSignature != null) put("buildSignature", buildSignature)
                     if (installerPackage != null) put("installerStore", installerPackage)
--- a/packages/package_info_plus/package_info_plus/android/src/main/kotlin/dev/fluttercommunity/plus/packageinfo/PackageInfoPlugin.kt
+++ b/packages/package_info_plus/package_info_plus/android/src/main/kotlin/dev/fluttercommunity/plus/packageinfo/PackageInfoPlugin.kt
@@ -42,9 +42,9 @@ class PackageInfoPlugin : MethodCallHandler, FlutterPlugin {
 
                 val infoMap = HashMap<String, String>()
                 infoMap.apply {
-                    put("appName", info.applicationInfo.loadLabel(packageManager).toString())
+                    put("appName", info.applicationInfo?.loadLabel(packageManager)?.toString() ?: "")
                     put("packageName", applicationContext!!.packageName)
-                    put("version", info.versionName)
+                    put("version", info?.versionName ?: "")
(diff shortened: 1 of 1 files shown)


vbuberen:
Could you share how you were able to reproduce the compilation failure? In which setup? Because the plugin switched to `compileSDK 34` quite some time ago and there were no issues with compilation.

I assume it is something happening with some different set of Android dependencies than the ones that we have in the example app, because I can't see any of these warnings and can run it successfully.

packages/package_info_plus/package_info_plus/android/src/main/kotlin/dev/fluttercommunity/plus/packageinfo/PackageInfoPlugin.kt
```diff
@@ -42,9 +42,9 @@ class PackageInfoPlugin : MethodCallHandler, FlutterPlugin {
 
                 val infoMap = HashMap<String, String>()
                 infoMap.apply {
-                    put("appName", info.applicationInfo.loadLabel(packageManager).toString())
+                    put("appName", info.applicationInfo?.loadLabel(packageManager)?.toString() ?: "")
                     put("packageName", applicationContext!!.packageName)
-                    put("version", info.versionName)
+                    put("version", info?.versionName ?: "")
```
miquelbeltran, line 47:
```suggestion
                    put("version", info.versionName ?: "")
```

The `?` in `info` shouldn't be necessary.

miquelbeltran (commented):
I haven't tested, but I believe this `?` is not necessary.

miquelbeltran:
It is weird indeed that we haven't seen this compilation issues, but checking the documentation, these two methods `versionName` and `applicationInfo` can indeed return null, so adding the null checks seems correct to me.

- https://developer.android.com/reference/android/content/pm/PackageInfo#applicationInfo
- https://developer.android.com/reference/android/content/pm/PackageInfo#versionName

ChristianEdwardPadilla:
Apologies, the initial description was wrong. The new signatures are in API level 35: https://developer.android.com/about/versions/15

packages/package_info_plus/package_info_plus/android/src/main/kotlin/dev/fluttercommunity/plus/packageinfo/PackageInfoPlugin.kt
```diff
@@ -42,9 +42,9 @@ class PackageInfoPlugin : MethodCallHandler, FlutterPlugin {
 
                 val infoMap = HashMap<String, String>()
                 infoMap.apply {
-                    put("appName", info.applicationInfo.loadLabel(packageManager).toString())
+                    put("appName", info.applicationInfo?.loadLabel(packageManager)?.toString() ?: "")
```
vbuberen, line 45:
```suggestion
                    put("appName", info.applicationInfo?.loadLabel(packageManager).toString() ?: "")
```

vbuberen (approved):

vbuberen:
There is a failing integration test on Android API 34 which is not related to timeouts we usually have and not related to changes in this PR.
Will be addressed in another PR.

PR #2898, opened 2024-05-02
chore(release): prepare for release

vbuberen:
## Description

Update for `android_alarm_manager_plus` based on #2864

PR #3017, opened 2024-06-02
ci(android_alarm_manager_plus): fix example package name

vbuberen:
## Description

Fix similar to #2973 as we had the same problem for the `android_alarm_manager_plus` (for example, issue is mentioned here https://github.com/fluttercommunity/plus_plugins/pull/2898#issuecomment-2089880624) plugin as well with bootstrapping and building the example app on CI.


miquelbeltran (approved):
thank you!

PR #3018, opened 2024-06-02
refactor(all): Remove website files, configs, mentions

vbuberen:
## Description

Removing everything connected with Plus Plugins website. Maybe I missed something and we would need an additional PR, but this one should do the majority of work.

I have just disabled the website on Firebase Hosting as well, so it is not working anymore.

P.S. Maybe it is worth to create a set of commits into every plugin to make Melos add a line to release notes mentioning that the website is removed.


miquelbeltran (approved):
thanks!

PR #812, opened 2022-03-31
fix(sensors_plus): iOS calibrated magnetometer

Zabadam:
## Description

iOS: Corrects magnetometer implementation, returning calibrated values from `DeviceMotion` sensor rather than raw sensor samples.

As the **user acceleration** implementation already employs this *calibrated* `DeviceMotion` sensor, it seems like an appropriate solution to acquiring compensated magnetometer data as well.

`melos run format` made changes to a huge number of files across all packages. I discarded all changes outside of `packages\sensors_plus\`.

The focus of this PR is the magnetometer implementation at the end of `sensors_plus\sensors_plus\ios\Classes\FLTSensorsPlusPlugin.m`.

I kindly request verification of this proposed solution, specifically as I have no means to test it myself without an iOS device, and also because I am uncertain of my Objective-C code.

## Related Issues

#781 ~~Magnetometer values are off~~

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`).
This will ensure a smooth and quick review process.

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [x] All existing and new tests are passing.
- [x] I updated the version in `pubspec.yaml` and `CHANGELOG.md`.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR [for `sensors_plus`].
- [x] I read and followed the [Flutter Style Guide].
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [x] No, this is *not* a breaking change, considering the previous output was technically not the expected output. Though if a developer *worked around* this gap previously in their code, this change may require some refactoring: 1.3.0 -> 1.4.0.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/fluttercommunity/plus_plugins/blob/main/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning

Issue #781, opened 2022-03-11
Magnetometer values are off


filippkowalski:
## System info

Platform the Issue occurs on: iOS
Plugin name: sensors_plus
Plugin version: 1.3.0

## Steps to Reproduce

It seems that Flutter's implementation of the Magnetometer is off when compared to iOS devices. 
I don't know yet if it's due to some additional flags that are set on the iOS side to calibrate the magnetometer ( https://stackoverflow.com/questions/28365121/how-to-get-magnetometer-data-using-swift ), but results for Flutter are way higher than for iOS devices. And actually sometimes when the Flutter app detects metals the values go down instead of going up.

You can easily test it out by downloading those 3 apps below and seeing the results by yourself.

iOS implementation
https://apps.apple.com/us/app/stud-finder-wall-detector/id1530653768
https://apps.apple.com/us/app/teslameter/id976926346

Flutter implementation
https://apps.apple.com/us/app/stud-finder/id1613688190

iOS
<p align="center">
  <img width="320" src="https://user-images.githubusercontent.com/10477861/157858540-561a7f00-a25e-4a78-bac3-8b6e2a39f341.PNG">
  <img width="320" src="https://user-images.githubusercontent.com/10477861/157858553-bdfc93f1-7286-4130-b1a4-19523bc0760e.PNG">
</p>

Flutter
<p align="center">
  <img width="320" src="https://user-images.githubusercontent.com/10477861/157858551-aa4f2a8c-6e9b-4949-9b8a-d466499358d3.PNG">
</p>

You can see on the screenshots that iOS values are lower than those on Flutter. All of those screenshots were made in the same phone position, with the phone lying flat.

One note here: Look at the X, Y, Z values here, because the Flutter app has the main displayed wrong.

cc: @Zabadam because I've seen that you were implementing this feature

Zabadam:
I have an extremely limited knowledge of iOS and as such the testing for that platform was less than thorough. 

If I take another look, it may be as simple as calling values from sensor manager incorrectly.

It seems Android has moved things around over the years, and iOS may be similar; whereby I could have been referencing irrelevant docs.

filippkowalski:
Possibly, I'll also take a look at what could be wrong here

Zabadam:
I have finally sat down to look into this.

[...]

PR #2250, opened 2023-10-11
fix(sensors_plus)!: Use magnetometer instead of deviceMotion

tlserver:
## Description

As #2248 said:
> In previous version, FPPMagnetometerStreamHandlerPlus use deviceMotion instead of magnetometer to allow iOS present calibration interaction. So, in iOS, userAccelerometerEventStream and magnetometerEventStream share the same sampling period. In android, they have their own sampling period.

So this PR use magnetometer instead of deviceMotion. but this is a BREAKING CHANGE!!

## Related Issues

Require #2248

## Checklist

- [x] I read the [Contributor Guide](https://github.com/fluttercommunity/plus_plugins/blob/main/CONTRIBUTING.md) and followed the process outlined there for submitting PRs.
- [x] I titled the PR using [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0).
- [x] I did not modify the `CHANGELOG.md` nor the plugin version in `pubspec.yaml` files.
- [x] All existing and new tests are passing.
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [x] Yes, this is a breaking change (please indicate that with a `!` in the title as explained in [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0)).
- [ ] No, this is *not* a breaking change.

PR #3019, opened 2024-06-03
fix(sensors_plus): corrected iOS magnetometer source

### LINKED ISSUE #2655: [Question]: sensors_plus reports different magnetic values in Android and iOS platforms
### What is your question?

I listen magnetometerEventStream to get magnetic values of the device in x, y, z directions. However, when I executed in Android and iOS devices which are oriented in almost same attitude, they reported very different values of MagnetometerEvent as shown below.

Android:
I/flutter (30017): [MagnetometerEvent (x: 7.65749979019165, y: 17.76624870300293, z: 17.66499900817871)]

iOS:
flutter: [MagnetometerEvent (x: 192.7864227294922, y: 8.718887329101562, z: -91.60765075683594)]

My code snippet is:

  void listenMagnetometerEvent() {
    _magnetometerEventStream ??= magnetometerEventStream().listen(
      (MagnetometerEvent event) {
        _magnetometerEventNotifier.value = event;
        print(event);
      },
      onError: (error) {
        _magnetometerEventStream = null;
        throw Exception('Listen magnetometer event error: $error');
      },
      cancelOnError: true,
    );
  }

### Checklist before submitting a question

- [X] I am using the latest version of the plugin
- [X] I did a `flutter clean`

Zabadam:
Obtain magnetometer events from DeviceMotion rather than raw startMagnetometerUpdates(). This also makes the magnetometer onCancel() method now correct.<sup>1</sup> Added set `showsDeviceMovementDisplay` to true for magnetometer.

<sup>1</sup> While working on the changes, I noticed the current implementation in `onCancel()` calls `_motionManager.stopDeviceMotionUpdates()`. This is interesting, because the correct course *is to obtain* events from MotionManager, but the current implementation is not. Because it currently calls `startMagnetometerUpdates()`, the `onCancel()` method ought to call `stopMagnetometerUpdates()`. Current users of sensors_plus, after obtaining magnetometer readings, could never have the sensor correctly canceled, [to my understanding](https://developer.apple.com/documentation/coremotion/cmmotionmanager/1615968-startmagnetometerupdatestoqueue#discussion).

At any rate, the onCancel gets to stay as it is while the source of the events is corrected.

## Description

When this package was converted to Swift, a [previous fix that corrected magnetometer](https://github.com/fluttercommunity/plus_plugins/pull/812) source to `DeviceMotion` was reverted to using `startMagnetometerUpdates()`.

This pull aims to restore the calibrated magnetometer readings that users are expecting.

As before, I have no way to test this code (no iOS device), and would appreciate someone's assitance in verifying the functionality.

## Related Issues

- *Fix #2655*
- *Refix #781*

## Checklist

## Breaking Change

  - Fixes incorrect behavior with no API changes.

### DIFF
--- a/packages/sensors_plus/sensors_plus/ios/Classes/FPPStreamHandlerPlus.swift
+++ b/packages/sensors_plus/sensors_plus/ios/Classes/FPPStreamHandlerPlus.swift
@@ -192,7 +192,10 @@ class FPPMagnetometerStreamHandlerPlus: NSObject, MotionStreamHandler {
             eventSink sink: @escaping FlutterEventSink
     ) -> FlutterError? {
         _initMotionManager()
-        _motionManager.startMagnetometerUpdates(to: OperationQueue()) { data, error in
+        _motionManager.showsDeviceMovementDisplay = true
+        _motionManager.startDeviceMotionUpdates(
+            using: CMAttitudeReferenceFrame.xArbitraryCorrectedZVertical, to: OperationQueue()
+        ) { data, error in
             if _isCleanUp {
                 return
             }
@@ -204,7 +207,7 @@ class FPPMagnetometerStreamHandlerPlus: NSObject, MotionStreamHandler {
                 ))
                 return
             }
-            let magneticField = data!.magneticField
+            let magneticField = data!.magneticField.field
             sendTriplet(x: magneticField.x, y: magneticField.y, z: magneticField.z, sink: sink)
         }
         return nil


Zabadam:
I dug into that test file at `sensors_plus\example\integration_test\sensors_plus_test.dart` and saw this:

```dart
void main() {
  IntegrationTestWidgetsFlutterBinding.ensureInitialized();

  // NOTE: The accelerometer events are not returned on iOS simulators.
  testWidgets('Can subscribe to accelerometerEvents and get non-null events',
      (WidgetTester tester) async {
    final completer = Completer<AccelerometerEvent>();
    late StreamSubscription<AccelerometerEvent> subscription;
    subscription =
        accelerometerEventStream().listen((AccelerometerEvent event) {
      completer.complete(event);
      subscription.cancel();
    });
    expect(await completer.future, isNotNull);
  }, skip: !Platform.isAndroid);
}
```

So perhaps this test was expected to fail regardless of my changes. (My changes are not tested by `subscription = accelerometerEventStream().listen(...)` and the comment describes the simulator's downfall.)

Thus it would still be beneficial for someone to manually test this change.

vbuberen:
Thanks a lot for your contribution. I will try to test the change by the end of the week.

Zabadam:
It seems that @Silverviql [has provided confirmation](https://github.com/fluttercommunity/plus_plugins/issues/2655#issuecomment-2149290477) of corrected functionality. Happy to hear this.

vbuberen:
Sorry for not getting back in the promised timeline.
Will do some additional testing myself today for sure.

vbuberen:
Finally got to this PR to test it and do some research.

> When this package was converted to Swift, a https://github.com/fluttercommunity/plus_plugins/pull/812 source to DeviceMotion was reverted to using startMagnetometerUpdates().

This is not the case. In fact, this change was done on purpose. See https://github.com/fluttercommunity/plus_plugins/pull/2248 and https://github.com/fluttercommunity/plus_plugins/pull/2250

However, I see that the calibrated values provided in your branch are actually the ones users might want to get. Also compared the output with a few apps obtaining magnetometer data and see that with changes from this PR getting same output.

Unfortunately, I see that the problem mentioned in #2250 with `userAccelerometer` and `magnetometer` sharing sampling rate is back with these changes, so there we need to find some solution to avoid this issue before releasing changes with this PR.

I will do some additional research in the nearest days to see what we can do here.

Zabadam:
> > When this package was converted to Swift, a https://github.com/fluttercommunity/plus_plugins/pull/812 source to DeviceMotion was reverted to using startMagnetometerUpdates().
> 
> This is not the case. In fact, this change was done on purpose. See https://github.com/fluttercommunity/plus_plugins/pull/2248 and https://github.com/fluttercommunity/plus_plugins/pull/2250

Better phrasing:  "At some point after this package was converted to using Swift . . ." I was gone a while and, without the full research on newer PRs, was trying to reason about a sort of deja-vu feeling. "Did I or didn't I get this corrected last time?" 😅 I never did receive full confirmation of fix. So, thank you for the full timeline.

> However, I see that the calibrated values provided in your branch are actually the ones users might want to get. Also compared the output with a few apps obtaining magnetometer data and see that with changes from this PR getting same output.

It seems to be the case overwhelmingly, yes, and it's excellent to hear the results match expectations after this PR.

> Unfortunately, I see that the problem mentioned in #2250 with `userAccelerometer` and `magnetometer` sharing sampling rate is back with these changes, so there we need to find some solution to avoid this issue before releasing changes with this PR.

I need to familiarize myself with the sampling rate code before I could be of further assistance here.

mshzhb (approved):
Tested and the values are correct on IOS now.

vbuberen:
> Magnetometer values are wrong on 6.1.0 IOS

This is not the place for a bug report. Open an issue or provide info in the existing one

Dchharlow4171717 (approved):

Issue #1548, opened 2023-02-16
[Request]: Use the `XFile.name` as the shared file name


josephelliot-wk:
### Plugin

share_plus

### Use case

Our app lets users open files (usually PDFs), and some of these files have special characters that don't play nicely with temp-file paths.

For example, `My First / Last File.pdf` causes an error when writing to temp storage.

To combat this, you need to sanitize the file name to play nicely with the native file system. But, when `share_plus` shares this file (via [Share.shareXFiles](https://github.com/fluttercommunity/plus_plugins/blob/main/packages/share_plus/share_plus_platform_interface/lib/method_channel/method_channel_share.dart#L142)), it uses the file name from the path instead of the `XFile`'s `name` property.

According to the [XFile documentation](https://github.com/flutter/packages/blob/main/packages/cross_file/lib/src/types/interface.dart#L21), `name` is intended to be used for this exact scenario.

### Proposal

I propose that the function `Share.shareXFiles` utilizes `XFile.name` for the file name, if available. If not available, then default to the path file name.

lukasnevosad:
I second to this. I am creating the file with `XFile.fromData()` and this is the only option for me to have a user friendly file name.

intraector:
Same here. Getting image from web without any file extension causes the share with .BIN

Boucaa:
Is there at least any workaround for this currently?

wildsurfer:
same issue

DieGlueckswurst:
Any updates on this ?

IshanBaliyan:
Same here, seems like the bug still isnt fixed

intraector:
Please fix this

adimshev:
any updates?

erighet:
Same request from me, now Share.shareXFiles causes files to have uuid like file name

bytepoets-jhe:
Same problem here

nidalshehab:
it  would be better to use file instead of XFile.fromData, this way you can name the file and delete it after sharing to keep the temporary directory clean, here is a work around:

```
Future<File> writeFile(List<int>? excel, String fileName) async {
    final directory = await getApplicationDocumentsDirectory();
    final path = directory.path;
    final file = File('$path/$fileName');
    return file.writeAsBytes(excel!);
  }
```
  if it's an exported excel file:
 ```
 List<int>? excel =  exportToExcel(); 
  String fileName = 'example.xlsx'; // Desired file name
[...]

PR #3077, opened 2024-07-05
feat(share_plus)!: Introduce optional parameter `nameOverride` to `shareXFiles`.

### LINKED ISSUE #3032: [Request]:  Improve documentation for filename used in shareXFiles
### Plugin

share_plus

### Use case

When sharing a file that is created from data, like so
```dart
      var data = utf8.encode('Hello World');
      var fileName = 'myFile.txt';
      var mimeType = 'text/plain';
      var xFile = XFile.fromData(data, name: fileName, mimeType: mimeType);
      Share.shareXFiles([xFile]);
```

expected behavior would be that a file called "myFile.txt" would be shared. Instead "{randomUUID}.txt" gets shared.

This is caused by `cross_file` ignoring the `name` parameter, when creating an xFile via the fromData constructor see [cross_file/lib/src/types/io.dart:43](https://github.com/flutter/packages/blob/9c6f691a8c891ff61d32eec4382c1850df599c67/packages/cross_file/lib/src/types/io.dart#L43)

This then causes `share_plus` to assign a name based on a random UUIDv6 and the mimetype.

### Proposal

Ideally a convenient way (for example an optional parameter) would be introduced that allows us to set/ override the filename of a given file to share.

But if that is not wanted for some reason at least please add to the documentation that
* A random filename is assigned to files that have no filename set
* A note that `XFile.fromData` ignores the the filename
  * I am aware that cross_file is a separate project, but at least when navigating the API docs it is not possible to find this information.
  * For example https://pub.dev/documentation/share_plus/latest/share_plus/XFile/XFile.fromData.html lists no such note.

frieder-audriga:
## Description

* Adds parameter `nameOverride` to funciton `shareXFiles`
* If set the values in  `nameOverride` will be used as the file names for the files share
* Also added new button in example that demonstrates this behavior.
* This is needed because `XFile.fromData` ignores `name` parameter on platforms other than web.

## Related Issues
- *Fix #3032*
- *Related #1548*

## Checklist

## Breaking Change

### DIFF
--- a/packages/share_plus/share_plus/README.md
+++ b/packages/share_plus/share_plus/README.md
@@ -103,7 +103,7 @@ Share.shareXFiles([XFile('assets/hello.txt')], text: 'Great picture');
 
 #### Share Data
 
-You can also share files that you dynamically generated from its data using [`XFile.fromData`](https://pub.dev/documentation/share_plus/latest/share_plus/XFile/XFile.fromData.html).
+You can also share files that you dynamically generate from its data using [`XFile.fromData`](https://pub.dev/documentation/share_plus/latest/share_plus/XFile/XFile.fromData.html).
 
 To set the name of these files, use the `fileNameOverrides` parameter, otherwise the file name will be a random UUID string.
 
--- a/packages/share_plus/share_plus/README.md
+++ b/packages/share_plus/share_plus/README.md
@@ -90,7 +90,7 @@ if (result.status == ShareResultStatus.dismissed) {
 }
 ```
 
-On web, you can use `SharePlus.shareXFiles()`. This uses the [Web Share API](https://web.dev/web-share/)
+On web, this uses the [Web Share API](https://web.dev/web-share/)
 if it's available. Otherwise it falls back to downloading the shared files.
 See [Can I Use - Web Share API](https://caniuse.com/web-share) to understand
 which browsers are supported. This builds on the [`cross_file`](https://pub.dev/packages/cross_file)
@@ -101,13 +101,18 @@ package.
 Share.shareXFiles([XFile('assets/hello.txt')], text: 'Great picture');
 ```
 
-#### Share data as Files
+#### Share Data
 
 You can also share files that you dynamically generated from its data using [`XFile.fromData`](https://pub.dev/documentation/share_plus/latest/share_plus/XFile/XFile.fromData.html).
+
 To set the name of these files, use the `fileNameOverrides` parameter, otherwise the file name will be a random UUID string.
+
 ```dart
 Share.shareXFiles([XFile.fromData(utf8.encode(text), mimeType: 'text/plain')], fileNameOverrides: ['myfile.txt']);
 ```
+
+> [!CAUTION]
+> The `name` parameter in the `XFile.fromData` method is ignored in most platforms. Use `fileNameOverrides` instead.
 
 ### Share URI
 
--- a/packages/share_plus/share_plus/README.md
+++ b/packages/share_plus/share_plus/README.md
@@ -101,6 +101,14 @@ package.
 Share.shareXFiles([XFile('assets/hello.txt')], text: 'Great picture');
 ```
 
+#### Share data as Files
+
+You can also share files that you dynamically generated from its data using [`XFile.fromData`](https://pub.dev/documentation/share_plus/latest/share_plus/XFile/XFile.fromData.html).
+To set the name of these files, use the `fileNameOverrides` parameter, otherwise the file name will be a random UUID string.
+```dart
+Share.shareXFiles([XFile.fromData(utf8.encode(text), mimeType: 'text/plain')], fileNameOverrides: ['myfile.txt']);
+```
+
 ### Share URI
 
 iOS supports fetching metadata from a URI when shared using `UIActivityViewController`.
--- a/packages/share_plus/share_plus/README.md
+++ b/packages/share_plus/share_plus/README.md
@@ -105,7 +105,7 @@ Share.shareXFiles([XFile('assets/hello.txt')], text: 'Great picture');
 
 You can also share files that you dynamically generate from its data using [`XFile.fromData`](https://pub.dev/documentation/share_plus/latest/share_plus/XFile/XFile.fromData.html).
 
-To set the name of these files, use the `fileNameOverrides` parameter, otherwise the file name will be a random UUID string.
+To set the name of such files, use the `fileNameOverrides` parameter, otherwise the file name will be a random UUID string.
 
 ```dart
 Share.shareXFiles([XFile.fromData(utf8.encode(text), mimeType: 'text/plain')], fileNameOverrides: ['myfile.txt']);
--- a/packages/share_plus/share_plus/example/lib/main.dart
+++ b/packages/share_plus/share_plus/example/lib/main.dart
@@ -256,6 +256,7 @@ class DemoAppState extends State<DemoApp> {
 
     scaffoldMessenger.showSnackBar(getResultSnackBar(shareResult));
   }
+
   void _onShareTextAsXFile(BuildContext context) async {
     final box = context.findRenderObject() as RenderBox?;
     final scaffoldMessenger = ScaffoldMessenger.of(context);
--- a/packages/share_plus/share_plus/example/lib/main.dart
+++ b/packages/share_plus/share_plus/example/lib/main.dart
@@ -4,6 +4,7 @@
 
 // ignore_for_file: public_member_api_docs
 
+import 'dart:convert';
 import 'dart:io';
 
 import 'package:file_selector/file_selector.dart'
@@ -32,6 +33,7 @@ class DemoAppState extends State<DemoApp> {
   String text = '';
   String subject = '';
   String uri = '';
+  String fileName = '';
   List<String> imageNames = [];
   List<String> imagePaths = [];
 
@@ -87,6 +89,18 @@ class DemoAppState extends State<DemoApp> {
                 maxLines: null,
                 onChanged: (String value) {
                   setState(() => uri = value);
+                },
+              ),
+              const SizedBox(height: 16),
+              TextField(
+                decoration: const InputDecoration(
+                  border: OutlineInputBorder(),
+                  labelText: 'Share Filename',
+                  hintText: 'Enter the filename you want to share your text as',
+                ),
+                maxLines: null,
+                onChanged: (String value) {
+                  setState(() => fileName = value);
                 },
               ),
               const SizedBox(height: 16),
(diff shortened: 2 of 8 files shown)


miquelbeltran:
Thanks @frieder-audriga 

I'd have much preferred that the `cross_file` package would actually use the `name` param, but since that's not the case this is the best second option ;)

packages/share_plus/share_plus/example/lib/main.dart
```diff
@@ -90,6 +92,18 @@ class DemoAppState extends State<DemoApp> {
                 },
               ),
               const SizedBox(height: 16),
+              TextField(
+                decoration: const InputDecoration(
+                  border: OutlineInputBorder(),
+                  labelText: 'Share Filename',
```
miquelbeltran, line 98:
```suggestion
                  labelText: 'Share Text as File',
```

packages/share_plus/share_plus/lib/src/share_plus_web.dart
```diff
@@ -156,10 +156,20 @@ class SharePlusWebPlugin extends SharePlatform {
...
-      webFiles.add(await _fromXFile(xFile));
+    if (fileNameOverrides != null) {
+      for (final entry in files.asMap().entries) {
+        webFiles.add(await _fromXFile(entry.value,
+            nameOverride: fileNameOverrides[entry.key]));
+      }
+    } else {
+      for (final xFile in files) {
+        webFiles.add(await _fromXFile(xFile));
+      }
```
miquelbeltran, line 172: There is probably a simpler way to write this with just a for loop with an index

Something like this (haven't checked if correct)

```dart
for (var index = 0; index < files.length; index++) {
  xFile = files[index];
  filename = fileNameOverrides?.elementAt(index);
   webFiles.add(await _fromXFile(xFile, filename));
}
```

packages/share_plus/share_plus_platform_interface/lib/method_channel/method_channel_share.dart
```diff
@@ -117,7 +119,8 @@ class MethodChannelShare extends SharePlatform {
   /// then make new file in TemporaryDirectory and return with path
   /// the system will automatically delete files in this
   /// TemporaryDirectory as disk space is needed elsewhere on the device
-  Future<XFile> _getFile(XFile file, {String? tempRoot}) async {
+  Future<XFile> _getFile(XFile file,
+      {String? tempRoot, String? nameOverride}) async {
```
miquelbeltran, line 123:
```suggestion
  Future<XFile> _getFile(
    XFile file, {
    String? tempRoot, 
    String? nameOverride,
  }) async {
```

Adding a comma and formatting the method

packages/share_plus/share_plus_platform_interface/lib/method_channel/method_channel_share.dart
```diff
@@ -137,13 +140,15 @@ class MethodChannelShare extends SharePlatform {
...
-                  null //If the filename has a valid extension
-          ? file.name
-          : "${const Uuid().v1().substring(10)}.$extension";
+      final filename = nameOverride ??
+          (file.name.isNotEmpty // If filename exists
+                  ||
+                  lookupMimeType(file.name) !=
+                      null //If the filename has a valid extension
+              ? file.name
+              : "${const Uuid().v1().substring(10)}.$extension");
```
miquelbeltran, line 151: I know this was not your code, but since we are at it, could this be simplified a bit?

packages/share_plus/share_plus_platform_interface/lib/method_channel/method_channel_share.dart
```diff
@@ -155,8 +160,13 @@ class MethodChannelShare extends SharePlatform {
   }
 
   /// A wrapper of [MethodChannelShare._getFile] for multiple files.
-  Future<List<XFile>> _getFiles(List<XFile> files) async =>
-      await Future.wait(files.map((entry) => _getFile(entry)));
+  Future<List<XFile>> _getFiles(
+          List<XFile> files, List<String>? fileNameOverrides) async =>
```
miquelbeltran, line 164: Add a trailing coma to the last param so it formats nicely

```suggestion
  Future<List<XFile>> _getFiles(
    List<XFile> files, 
    List<String>? fileNameOverrides,
  ) async =>
```

packages/share_plus/share_plus_platform_interface/lib/method_channel/method_channel_share.dart
```diff
@@ -155,8 +160,13 @@ class MethodChannelShare extends SharePlatform {
...
+      (fileNameOverrides == null)
+          ? await Future.wait(files.map((entry) => _getFile(entry)))
+          : await Future.wait(files.asMap().entries.map((entry) => _getFile(
+              entry.value,
+              nameOverride: fileNameOverrides[entry.key])));
```
miquelbeltran, line 169: Need to look if this can be simplified as well.

frieder-audriga: We could convert the function from an expression to a block body and then use a for loop, like so
```dart
  /// A wrapper of [MethodChannelShare._getFile] for multiple files.
  Future<List<XFile>> _getFiles(
          List<XFile> files, List<String>? fileNameOverrides) async {
    final List<XFile> retrievedFiles = [];
    for (var index = 0; index < files.length; index++) {
      final String? fileName = fileNameOverrides?.elementAt(index);
      retrievedFiles.add(await _getFile(files[index], nameOverride:  fileName));
    }
    return retrievedFiles;
  }
```
I'm not completely sure if this would be simpler though.

packages/share_plus/share_plus_platform_interface/lib/method_channel/method_channel_share.dart
frieder-audriga: I have factored out the `file.name.isNotEmpty    ||   lookupMimeType(file.name) !=  null` boolean in a separate variable to simplify the expression.

However I'm not sure if this check makes too much sense in the first place.
In a scenario where the filename is not empty, the left half is true and the right doesn't get evaluated.
In a scenario there the filename *is* empty I'm pretty sure `lookupMimeType('')` will also return null.

packages/share_plus/share_plus/lib/src/share_plus_web.dart
frieder-audriga: You're right, I have refactored that part to use a single for loop, like you suggested

frieder-audriga:
Thank you for the comments and suggestions. I added a short subsection to the README and addressed the other comments.

packages/share_plus/share_plus_platform_interface/lib/method_channel/method_channel_share.dart
miquelbeltran: Yes, I think so, I much prefer to use indexes as well.

miquelbeltran: I pushed a refactor to your fork, I hope you don't mind!

packages/share_plus/share_plus_platform_interface/lib/platform_interface/share_plus_platform.dart
```diff
@@ -61,12 +61,14 @@ class SharePlatform extends PlatformInterface {
     String? subject,
     String? text,
     Rect? sharePositionOrigin,
+    List<String>? fileNameOverrides,
```
miquelbeltran, line 64: Because we are changing the platform interface, this requires a major version, otherwise this will cause issues with clients where the platform interface and the main package are not in sync.

packages/share_plus/share_plus_platform_interface/lib/method_channel/method_channel_share.dart
miquelbeltran: Had a second look and I think it is fine now, thanks!

miquelbeltran (approved):
Tested on Android. LGTM! I did some changes and pushed them to @frieder-audriga fork.

This feature has to be marked as a breaking change since it modifies the platform interface API. fyi @vbuberen 

I'll wait a bit and then it can be merged.

packages/share_plus/share_plus/README.md
```diff
@@ -101,6 +101,19 @@ package.
 Share.shareXFiles([XFile('assets/hello.txt')], text: 'Great picture');
 ```
 
+#### Share Data
+
+You can also share files that you dynamically generated from its data using [`XFile.fromData`](https://pub.dev/documentation/share_plus/latest/share_plus/XFile/XFile.fromData.html).
```
vbuberen, line 106:
```suggestion
You can also share files that you dynamically generate from its data using [`XFile.fromData`](https://pub.dev/documentation/share_plus/latest/share_plus/XFile/XFile.fromData.html).
```

packages/share_plus/share_plus/README.md
```diff
@@ -101,6 +101,19 @@ package.
...
+
+To set the name of these files, use the `fileNameOverrides` parameter, otherwise the file name will be a random UUID string.
```
vbuberen, line 108:
```suggestion
To set the name of such files, use the `fileNameOverrides` parameter, otherwise the file name will be a random UUID string.
```

packages/share_plus/share_plus/lib/share_plus.dart
```diff
@@ -111,6 +111,12 @@ class Share {
   /// origin rect for the share sheet to popover from on iPads and Macs. It has no effect
   /// on other devices.
   ///
+  /// The optional parameter [fileNameOverrides] can be used to override the file names
+  /// used for the shared files.
```
vbuberen, line 115:
```suggestion
  /// The optional parameter [fileNameOverrides] can be used to override the names of shared files
```

packages/share_plus/share_plus/lib/share_plus.dart
```diff
@@ -111,6 +111,12 @@ class Share {
...
+  /// If set, the length must match the length of [files].
```
vbuberen, line 116:
```suggestion
  /// When set, the list length must match the number of [files] to share.
```

packages/share_plus/share_plus_platform_interface/lib/method_channel/method_channel_share.dart
```diff
@@ -76,10 +76,14 @@ class MethodChannelShare extends SharePlatform {
...
     String? text,
     Rect? sharePositionOrigin,
+    List<String>? fileNameOverrides,
   }) async {
     assert(files.isNotEmpty);
-
-    final filesWithPath = await _getFiles(files);
+    assert(
+      fileNameOverrides == null || files.length == fileNameOverrides.length,
+      "fileNameOverrides list must be equal size as file list.",
```
vbuberen, line 84:
```suggestion
      "fileNameOverrides list must have the same length as files list.",
```

packages/share_plus/share_plus_platform_interface/lib/method_channel/method_channel_share.dart
```diff
@@ -137,13 +145,16 @@ class MethodChannelShare extends SharePlatform {
       final tempSubfolderPath = "$tempRoot/${const Uuid().v4()}";
       await Directory(tempSubfolderPath).create(recursive: true);
 
+      // True if filename exists, or the filename has a valid extension
```
vbuberen, line 148:
```suggestion
      // True if filename exists or the filename has a valid extension
```

vbuberen (approved):
Looks good to me, but have a few suggestions to improve comments.

miquelbeltran:
Merging!

PR #3082, opened 2024-07-09
test(package_info_plus): Fix integration test for Android

vbuberen:
## Description

Fixing tests for `installerStore` value on Android as it seems like since Android 14 OS returns `com.android.shell` for apps installed via package manager, adb command (I suppose so) instead of `null`. Don't see any mentions of it in the documentation. Earlier we didn't see this test failing because we had no API 34 in our build matrix.
Checked with emulators and real phone with Android 14 - on Android 13 I see `null`, while on 14 for both virtual and real device see `com.android.shell` when installing example via Android Studio. I believe this fix for the integration test should be correct.

Screenshot 2024-07-09 at 14 32 20

### DIFF
--- a/packages/package_info_plus/package_info_plus/example/integration_test/package_info_plus_test.dart
+++ b/packages/package_info_plus/package_info_plus/example/integration_test/package_info_plus_test.dart
@@ -85,13 +85,22 @@ void main() {
       expect(find.text('not available'), findsOneWidget);
     } else {
       if (Platform.isAndroid) {
+        final androidVersionInfo = await DeviceInfoPlugin().androidInfo;
+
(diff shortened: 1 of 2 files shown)


miquelbeltran (approved):
Smart to use `device_info_plus` to determine the OS version in the unit tests ;)

miquelbeltran:
Title should be `package_info_plus`!

vbuberen:
> Smart to use device_info_plus to determine the OS version in the unit tests ;)

The first thing that came to my mind. Also, I don't know other ways to do this check.

> Title should be package_info_plus!

Thanks for catching. Doing multiple things at once never ends well :(

PR #3083, opened 2024-07-09
feat(connectivity_plus): Return multiple active types on MacOS/iOS when available

### LINKED ISSUE #3073: [Bug]: Connectivity plus returns only ConnectivityResult.ethernet even when device is connected via LAN and WiFi
### Platform

MacOS 14.4.1

### Plugin

connectivity_plus

### Version

6.0.3

### Flutter SDK

Flutter 3.22.0 • channel stable

### Steps to reproduce

Run https://github.com/fluttercommunity/plus_plugins/blob/main/packages/connectivity_plus/connectivity_plus/example/lib/main.dart

Even while my Mac is connected to ethernet and Wifi, only the ethernet connection is shown. As a list of ConnectivityResult is returned, 

### Code Sample

```dart
Example app from repo.
https://github.com/fluttercommunity/plus_plugins/blob/main/packages/connectivity_plus/connectivity_plus/example/lib/main.dart
```

### Logs

```shell
flutter: Connectivity changed: [ConnectivityResult.ethernet]
flutter: Connectivity changed: [ConnectivityResult.none]
flutter: Connectivity changed: [ConnectivityResult.ethernet]

after disconnecting from lan network
flutter: Connectivity changed: [ConnectivityResult.none]
flutter: Connectivity changed: [ConnectivityResult.wifi]

plugging in to lan again
flutter: Connectivity changed: [ConnectivityResult.ethernet]
```

### Flutter Doctor

```shell
[✓] Flutter (Channel stable, 3.22.0, on macOS 14.4.1 23E224 darwin-arm64, locale de-DE)
    • Flutter version 3.22.0 on channel stable at /Users/vsz/Development/flutter
    • Framework revision 5dcb86f68f (8 weeks ago), 2024-05-09 07:39:20 -0500
    • Engine revision f6344b75dc
    • Dart version 3.4.0
    • DevTools version 2.34.3

[✓] Android toolchain - develop for Android devices (Android SDK version 34.0.0)
    • Android SDK at /Users/vsz/Library/Android/sdk
    • Platform android-34, build-tools 34.0.0
    • Java binary at: /Applications/Android Studio.app/Contents/jbr/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 17.0.7+0-17.0.7b1000.6-10550314)

[✓] Xcode - develop for iOS and macOS (Xcode 15.4)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 15F31d
    • CocoaPods version 1.15.2

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 2023.1)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build 17.0.7+0-17.0.7b1000.6-10550314)

[✓] VS Code (version 1.90.2)
    • VS Code at /Applications/Visual Studio Code.app/Contents
    • Flutter extension version 3.90.0

[✓] Connected device (3 available)
    • macOS (desktop)                 • macos                 • darwin-arm64   • macOS 14.4.1 23E224 darwin-arm64
    • Mac Designed for iPad (desktop) • mac-designed-for-ipad • darwin         • macOS 14.4.1 23E224 darwin-arm64
    • Chrome (web)                    • chrome                • web-javascript • Google Chrome 126.0.6478.127
    ! Error: Browsing on the local area network for iPad von Vera. Ensure the device is unlocked and attached with a cable or
      associated with the same local area network as this Mac.
      The device must be opted into Developer Mode to connect wirelessly. (code -27)

[✓] Network resources
    • All expected network resources are available.

• No issues found!
```

- [X] I did a `flutter clean`

vbuberen:
## Description

Switching from check for active used connection on iOS/MacOS to available connections to return multiple connection types when such are available. Tested with a dongle for MacBook which has Ethernet port and looks like getting results as expected:
Screenshot 2024-07-09 at 18 08 41
Here is a screenshot from a few more test cases where I tried to connect/disconnect both Wi-Fi and Ethernet
Screenshot 2024-07-09 at 18 13 41

Additionally did a few changes to the API documentation comments.

## Related Issues

Closes #3073

## Breaking Change

### DIFF
--- a/packages/connectivity_plus/connectivity_plus/darwin/Classes/PathMonitorConnectivityProvider.swift
+++ b/packages/connectivity_plus/connectivity_plus/darwin/Classes/PathMonitorConnectivityProvider.swift
@@ -11,24 +11,25 @@ public class PathMonitorConnectivityProvider: NSObject, ConnectivityProvider {
 
   private func connectivityFrom(path: NWPath) -> [ConnectivityType] {
     var types: [ConnectivityType] = []
-    
-    // Check for connectivity and append to types array as necessary
-    if path.status == .satisfied {
-      if path.usesInterfaceType(.wifi) {
-        types.append(.wifi)
-      }
-      if path.usesInterfaceType(.cellular) {
-        types.append(.cellular)
-      }
-      if path.usesInterfaceType(.wiredEthernet) {
-        types.append(.wiredEthernet)
-      }
-      if path.usesInterfaceType(.other) {
-        types.append(.other)
-      }
-    }
-    
-    return types.isEmpty ? [.none] : types
+
+     for interface in path.availableInterfaces {
+            switch interface.type {
+            case .wifi:
+                types.append(.wifi)
+            case .wiredEthernet:
+                types.append(.wiredEthernet)
+            case .cellular:
+                types.append(.cellular)
+            case .loopback:
+                types.append(.other)
+            case .other:
+                types.append(.other)
+            @unknown default:
+                types.append(.other)
+            }
+        }
+
+        return types.isEmpty ? [.none] : types
   }
 
   public var currentConnectivityTypes: [ConnectivityType] {
--- a/packages/connectivity_plus/connectivity_plus/example/macos/Runner.xcodeproj/project.pbxproj
+++ b/packages/connectivity_plus/connectivity_plus/example/macos/Runner.xcodeproj/project.pbxproj
@@ -259,7 +259,7 @@
 			isa = PBXProject;
 			attributes = {
 				LastSwiftUpdateCheck = 0920;
-				LastUpgradeCheck = 1300;
+				LastUpgradeCheck = 1510;
 				ORGANIZATIONNAME = "";
 				TargetAttributes = {
 					331C80D4294CF70F00263BE5 = {
--- a/packages/connectivity_plus/connectivity_plus/example/macos/Runner.xcodeproj/xcshareddata/xcschemes/Runner.xcscheme
+++ b/packages/connectivity_plus/connectivity_plus/example/macos/Runner.xcodeproj/xcshareddata/xcschemes/Runner.xcscheme
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Scheme
-   LastUpgradeVersion = "1300"
+   LastUpgradeVersion = "1510"
    version = "1.3">
    <BuildAction
       parallelizeBuildables = "YES"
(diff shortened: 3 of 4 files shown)


vbuberen:
Would appreciate an additional check on how it works on iOS with mobile and wi-fi active as don't have iPhone with a SIM card

miquelbeltran:
> Would appreciate an additional check on how it works on iOS with mobile and wi-fi active as don't have iPhone with a SIM card

I should be able to do that.

Btw, I think we can change the Android implementation to also report the available connections instead of just the used ones, but I don't remember exactly now what was the api.

miquelbeltran:
Maybe we should implement this change in a different stream listener?

In other words:

- `checkConnectivity` and `onConnectivityChanged` continue to report the "active" used connection
- A new property `availableConnectivity` and `onAvailableConnectivityChanged` report on all available interfaces, even when not actively used. 

I know it is more work, but I think there is still a valid use case for the previous implementation.

While we implement the rest of the platforms, we can mirror `availableConnectivity` to `checkConnectivity` if no specific implementation for `availableConnectivity` exists.

vbuberen:
> Btw, I think we can change the Android implementation to also report the available connections instead of just the used ones, but I don't remember exactly now what was the api.

For some reason I thought that on Android we already do it this way and return multiple types after introducing this feature with multiple types returned. But seems like I confused something as tested and see that no, we return only the one currently active/used ones.

> In other words:
checkConnectivity and onConnectivityChanged continue to report the "active" used connection
A new property availableConnectivity and onAvailableConnectivityChanged report on all available interfaces, even when not actively used.

I like this idea. It would make sense for users, who rely on `onConnectivityChanged` as a way to determine if user has internet (even though the plugin doesn't guarantee that, but still most people use the plugin for this exact purpose). 
Will need to do some additional research on APIs to achieve this.

miquelbeltran:
If I remember correctly, the `hasTransport()` method says if the interface exists, and `hasCapability` that it is being used. Both are on the `NetworkCapabilities` class, so the implementation shouldn't be too different.

vbuberen:
> If I remember correctly, the hasTransport() method says if the interface exists, and hasCapability() if it is being used. Both are on the NetworkCapabilities class, so the implementation shouldn't be too different.

Yeah, that is for Android. I meant more Windows/Linux platforms as a subject for research as not really familiar with them.

vbuberen:
I don't see any activity in the related issue, so will just close the PR, considering having no real value in expanding API that way.

PR #3097, opened 2024-07-20
chore(all): Add pub.dev topics

### LINKED ISSUE #2941: [Request]: Consider adding `topics` to `pubspec.yaml` for better discoverability
### Plugin

all

### Use case

Related packages can be hard to find.

### Proposal

Topics helps users on pub.dev discover packages, and related packages.

You can see a list of [topics here](https://pub.dev/topics).
You can also make your own topics.

Topics are added to `pubspec.yaml` as follows:

```yaml
topics:
 - my-topic
 - some-other-topic
```

See also: https://dart.dev/tools/pub/pubspec#topics

vbuberen:
## Description

Picked some of available topics from https://dart.dev/tools/pub/pubspec#topics and added for every plugin.
Feel free to add some if I missed something or there are some better topics.

## Related Issues

Closes #2941

### DIFF
--- a/packages/android_alarm_manager_plus/pubspec.yaml
+++ b/packages/android_alarm_manager_plus/pubspec.yaml
@@ -8,6 +8,7 @@ issue_tracker: https://github.com/fluttercommunity/plus_plugins/labels/android_a
 topics:
   - android
   - alarm
+  - utils
 
 environment:
   sdk: ">=2.18.0 <4.0.0"
--- a/packages/android_alarm_manager_plus/pubspec.yaml
+++ b/packages/android_alarm_manager_plus/pubspec.yaml
@@ -5,6 +5,9 @@ version: 4.0.3
 homepage: https://plus.fluttercommunity.dev/
 repository: https://github.com/fluttercommunity/plus_plugins/tree/main/packages/android_alarm_manager_plus
 issue_tracker: https://github.com/fluttercommunity/plus_plugins/labels/android_alarm_manager_plus
+topics:
+  - android
+  - alarm
 
 environment:
   sdk: ">=2.18.0 <4.0.0"
--- a/packages/android_intent_plus/pubspec.yaml
+++ b/packages/android_intent_plus/pubspec.yaml
@@ -4,6 +4,10 @@ version: 5.0.2
 homepage: https://plus.fluttercommunity.dev/
 repository: https://github.com/fluttercommunity/plus_plugins/tree/main/packages/android_intent_plus
 issue_tracker: https://github.com/fluttercommunity/plus_plugins/labels/android_intent_plus
+topics:
+  - android
+  - intent
+  - utils
 
 environment:
   sdk: ">=2.18.0 <4.0.0"
--- a/packages/device_info_plus/device_info_plus/pubspec.yaml
+++ b/packages/device_info_plus/device_info_plus/pubspec.yaml
@@ -5,6 +5,9 @@ version: 10.1.0
 homepage: https://plus.fluttercommunity.dev/
 repository: https://github.com/fluttercommunity/plus_plugins/tree/main/packages/device_info_plus/device_info_plus
 issue_tracker: https://github.com/fluttercommunity/plus_plugins/labels/device_info_plus
+topics:
+  - device
+  - information
 
 flutter:
   plugin:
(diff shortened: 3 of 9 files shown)


packages/android_intent_plus/pubspec.yaml
```diff
@@ -4,6 +4,10 @@ version: 5.0.2
 homepage: https://plus.fluttercommunity.dev/
 repository: https://github.com/fluttercommunity/plus_plugins/tree/main/packages/android_intent_plus
 issue_tracker: https://github.com/fluttercommunity/plus_plugins/labels/android_intent_plus
+topics:
+  - android
+  - intent
```
aanas-sayed, line 9: Topic for `intents` already exists which could be used as `intent` does not exist.
https://pub.dev/topics

packages/android_alarm_manager_plus/pubspec.yaml
```diff
@@ -5,6 +5,9 @@ version: 4.0.3
 homepage: https://plus.fluttercommunity.dev/
 repository: https://github.com/fluttercommunity/plus_plugins/tree/main/packages/android_alarm_manager_plus
 issue_tracker: https://github.com/fluttercommunity/plus_plugins/labels/android_alarm_manager_plus
+topics:
+  - android
```
aanas-sayed, line 9: 1. I noticed for some of these have `android` topic added, but then missing the platform information from others. Do you think it worth adding topics for every platform? There is the 5 topic limitation so I was thinking maybe either the specific platform like in this case `android` or in other cases `cross-platform`. What do you think?
2. Perhaps `native` could be added to this and most of the other packages as they include native code for each platform? Its the only other topic I can see being related to this project from existing topics.

packages/device_info_plus/device_info_plus/pubspec.yaml
```diff
@@ -5,6 +5,9 @@ version: 10.1.0
 homepage: https://plus.fluttercommunity.dev/
 repository: https://github.com/fluttercommunity/plus_plugins/tree/main/packages/device_info_plus/device_info_plus
 issue_tracker: https://github.com/fluttercommunity/plus_plugins/labels/device_info_plus
+topics:
+  - device
+  - information
```
aanas-sayed, line 10: All other packages have `utils` topic but not this (and android alarm manager plus). Should this be added?

packages/android_alarm_manager_plus/pubspec.yaml
```diff
@@ -5,6 +5,9 @@ version: 4.0.3
...
+  - alarm
```
aanas-sayed, line 10: All other packages have `utils` topic but not this (and device info plus). Should this be added?

aanas-sayed (commented):
Thought I'd have a look and had a few points I wanted to raise. Thank you!

packages/android_alarm_manager_plus/pubspec.yaml
miquelbeltran: I think just `android` is fine in this case, since it is an android-only package. Same with `android_intent_plus`.

miquelbeltran (approved):
LGTM. I :+1: some of the suggestions by @aanas-sayed as well.

packages/android_intent_plus/pubspec.yaml
vbuberen: It is done on purpose. Android has `Intent` [mechanism](https://developer.android.com/reference/android/content/Intent), so I want to be specific that it is about it and not some generic intents.

packages/android_alarm_manager_plus/pubspec.yaml
vbuberen: > I noticed for some of these have android topic added, but then missing the platform information from others. Do you think it worth adding topics for every platform? There is the 5 topic limitation so I was thinking maybe either the specific platform like in this case android or in other cases cross-platform. What do you think?

No, it makes no sense and is an equivalent to tons of hashtags in socials. There is already info on pub.dev about supported platforms. `android` here is to just highlight that 2 packages are specifically about some Android functionality.

> Perhaps native could be added to this and most of the other packages as they include native code for each platform? Its the only other topic I can see being related to this project from existing topics.

Same as previously - I don't want to turn this section into a set of hashtags just because.

packages/device_info_plus/device_info_plus/pubspec.yaml
vbuberen: Yes, missed that one. Thanks.

packages/android_alarm_manager_plus/pubspec.yaml
vbuberen: Yes, missed it. Thanks.

Issue #1067, opened 2022-09-13
[Bug]: Share plus dialog is not being closed.


mohit-varma:
### Platform

iPad 15.6.1

### Plugin

share_plus

### Version

4.1.0

### Flutter SDK

2.10.1

### Steps to reproduce

Hi, The share plus dialog is not being closed in case one. Suppose I opened the share plus dialog and sent a push notification on my mobile, after clicking on the notification share plus dialog should dismiss. it is not being closed.

### Logs

```shell
there is no such exceptions related share plus.
```

### Flutter Doctor

```shell
[✓] Flutter (Channel stable, 2.10.1, on macOS 12.5.1 21G83 darwin-arm, locale en-IN)
    • Flutter version 2.10.1 at /Users/mayank/Downloads/DownloadedSoftware/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision db747aa133 (7 months ago), 2022-02-09 13:57:35 -0600
    • Engine revision ab46186b24
    • Dart version 2.16.1
    • DevTools version 2.9.2

[✓] Android toolchain - develop for Android devices (Android SDK version 31.0.0)
    • Android SDK at /Users/mayank/Library/Android/sdk
    • Platform android-31, build-tools 31.0.0
    • Java binary at: /Users/mayank/Downloads/Android Studio.app/Contents/jre/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 11.0.10+0-b96-7249189)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 13.4.1)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • CocoaPods version 1.11.2

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 2020.3)
    • Android Studio at /Users/mayank/Downloads/Android Studio.app/Contents
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build 11.0.10+0-b96-7249189)

[✓] IntelliJ IDEA Community Edition (version 2021.3)
    • IntelliJ at /Applications/IntelliJ IDEA CE.app
    • Flutter plugin can be installed from:
[...]

PR #3109, opened 2024-07-27
feat!: #1067 close share plus sheet

Pierre-Monier:
## Description

*Add a `close` method to the `share_plus` plugin. Currently, it only works on iOS (this was my need for a project). This support only one platform, but after discussing in [this issue](https://github.com/fluttercommunity/plus_plugins/issues/1067) it appears to be ok to open a PR.*

## Related Issues

- *Related #1067*

## Checklist

## Breaking Change

### DIFF
--- a/packages/share_plus/share_plus/example/pubspec.yaml
+++ b/packages/share_plus/share_plus/example/pubspec.yaml
@@ -4,7 +4,11 @@ description: Demonstrates how to use the share_plus plugin.
 dependencies:
   flutter:
     sdk: flutter
-  share_plus: ^9.0.0
+  share_plus:
+    git:
+      url: git@github.com:Pierre-Monier/plus_plugins.git
+      ref: feat/close-share-plus-sheet
+      path: packages/share_plus
   image_picker: ^1.0.0
   file_selector: ^1.0.0
 
--- a/packages/share_plus/share_plus/example/pubspec.yaml
+++ b/packages/share_plus/share_plus/example/pubspec.yaml
@@ -5,10 +5,7 @@ dependencies:
   flutter:
     sdk: flutter
   share_plus:
-    git:
-      url: git@github.com:Pierre-Monier/plus_plugins.git
-      ref: feat/close-share-plus-sheet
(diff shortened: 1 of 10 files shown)


packages/share_plus/share_plus/example/pubspec.yaml
```diff
@@ -4,7 +4,8 @@ description: Demonstrates how to use the share_plus plugin.
 dependencies:
   flutter:
     sdk: flutter
-  share_plus: ^9.0.0
+  share_plus:
+    path: ../
```
miquelbeltran, line 8: This should be reverted. Please take a look at the contribution guide in the repo as it has instructions on how to bootstrap the repository so the examples use the local packages (melos bootstrap)

packages/share_plus/share_plus_platform_interface/lib/method_channel/method_channel_share.dart
```diff
@@ -115,6 +114,17 @@ class MethodChannelShare extends SharePlatform {
     return ShareResult(result, _statusFromResult(result));
   }
 
+  /// Closes the share sheet. Currently only supported on iOS.
+  @override
+  Future<void> close() {
+    if (defaultTargetPlatform != TargetPlatform.iOS) {
+      // Currently not supported outside of iOS
+      return Future<void>.value();
+    }
```
miquelbeltran, line 123: If we go with the feature, I'd probably add an `assert` here, so the method fails in debug mode.

miquelbeltran:
Marking it as breaking as it makes changes on the platform interface.

miquelbeltran:
Hi again, after discussing internally, we decided we cannot this new feature unless a solution for Android is also provided. Sorry that I gave false expectations.

Looking at the code, I think it would be possible for you to deploy your solution as an independent plugin that captures the `RootViewController` and closes it, as that should work even outside the context of this plugin.

If later anyone is able to find a way to implement a similar solution for Android, then we can talk this again.

Nevertheless, thanks for the contribution!

PR #3166, opened 2024-08-09
docs: Add link to gradle plugin docs in COMPILATION.yaml

miquelbeltran:
## Description

Maybe we avoid some of the Android compilation tickets we have been seeing recently.


vbuberen (approved):
I have low hopes after seeing how many people don't even google their error logs.

Issue #3152, opened 2024-08-07
[Request]: Support Swift Package Manager


koji-1009:
### Plugin

battery_plus, connectivity_plus, device_info_plus, network_info_plus, package_info_plus, sensors_plus, share_plus

### Use case

When Swift Package Manager(SPM) support is added to the flutter stable channel, developers will get the plus_plugins package via SPM on iOS/macOS.

### Proposal

With the release of Flutter 3.24.0, SPM support has been announced in the main channel.

https://medium.com/flutter/flutter-3-24-dart-3-5-204b7d20c45d
https://docs.flutter.dev/packages-and-plugins/swift-package-manager/for-plugin-authors

Since SPM support is compatible with CocoaPods support, I think it would be wonderful if the popular plus_plugins could support SPM as early as possible.

Target packages are following.

- battery_plus
- connectivity_plus
- device_info_plus
- network_info_plus
- package_info_plus
- sensors_plus
- share_plus

koji-1009:
Thank you for the clarification!

I too think it would be good to keep support for CocoaPods and add support for SPM as well, as in the package managed by flutter.dev.
https://github.com/flutter/packages/pulls?q=is%3Apr+is%3Aclosed++%22Add+Swift+Package+Manager+support%22+

charafau:
Current status for this

| Plugin      | Status |
|-------------------|----|
| battery_plus | [PR](https://github.com/fluttercommunity/plus_plugins/pull/3154)  |
| connectivity_plus | ❌  |
| device_info_plus  | [PR](https://github.com/fluttercommunity/plus_plugins/pull/3167) |
| network_info_plus | ❌  |
| package_info_plus | ❌  |
| sensors_plus      | ❌  |
| share_plus        | ❌  |

vbuberen:
Would like to join the discussion here. 

First of all, thanks a lot for taking care of creating PRs for every plugin.

I would like to not merge any of those changes till SPM support gets out of the state of early stages and there is no warning on top of the page saying that it is a subject of change. I am against merging any half-baked solutions. There is no value except additional possible overhead due to changes that Google might introduce. For those who want to experiment - they can do it in their forks.

vbuberen:
> I think it would be wonderful if the popular plus_plugins could support SPM as early as possible.

Yes, it is a good point, but we are pretty quick to react with releases when some changes are introduced, so I don't see a problem of not having SPM support for now.

koji-1009:
[...]

PR #3167, opened 2024-08-09
feat(device_info_plus): Add Swift Package Manager support

charafau:

Since SPM support is comming I plan to add it to all plugins. 

Followed the guide - https://docs.flutter.dev/packages-and-plugins/swift-package-manager/for-plugin-authors 
and checked it for both cocoapods and spm


packages/device_info_plus/device_info_plus/ios/device_info_plus/.build/workspace-state.json
miquelbeltran, line 1: Should this file be checked in?

packages/device_info_plus/device_info_plus/ios/device_info_plus/Sources/device_info_plus/include/device_info_plus/.gitkeep
miquelbeltran, line 1: I guess this file can be removed

packages/device_info_plus/device_info_plus/.gitignore
```diff
@@ -45,3 +45,4 @@ build/
 .project
 .classpath
 .settings
+!.gitkeep
```
miquelbeltran, line 48: Since .gitkeep would be removed, we don't need this change either

miquelbeltran (commented):
I am not super familiar with the process yet, but took a look at it, and left some comments.

Also, we should do the same for the macOS part of the project, right? In that case, we should indicate in the Title that this only covers iOS

packages/device_info_plus/device_info_plus/ios/device_info_plus/.build/workspace-state.json
koji-1009: I suggest updating the root `.gitignore` to remove two directory from all packages.
The following PR is helpful.

https://github.com/flutter/packages/pull/6705

charafau:
> I am not super familiar with the process yet, but took a look at it, and left some comments.
> 
> Also, we should do the same for the macOS part of the project, right? In that case, we should indicate in the Title that this only covers iOS

Thanks for fast review, this is also my first time doing it. I will apply suggestions from you and @koji-1009 and upload today. Hopefully migrate all plugins

charafau:
Thanks guys, should be good now. I also added support for macos

miquelbeltran:
Please take a look at the macos integration jobs, as they are now failing.

charafau:
> Please take a look at the macos CI jobs, as they are now failing.

Sorry didn't notice that it's a scrollable list 😅  fixed and is passing now

packages/device_info_plus/device_info_plus/ios/device_info_plus.podspec
```diff
@@ -14,11 +14,11 @@ Downloaded by pub (not CocoaPods).
   s.author           = { 'Flutter Community Team' => 'authors@fluttercommunity.dev' }
   s.source           = { :http => 'https://github.com/fluttercommunity/plus_plugins/tree/main/packages/device_info_plus' }
   s.documentation_url = 'https://pub.dev/packages/device_info_plus'
-  s.source_files = 'Classes/**/*'
-  s.public_header_files = 'Classes/**/*.h'
+  s.source_files = 'device_info_plus/Sources/device_info_plus/**/*.{h,m}'
+  s.public_header_files = 'device_info_plus/Sources/device_info_plus/include/**/*.h'
```
loic-sharma, line 18:
```suggestion
  s.public_header_files = 'device_info_plus/Sources/device_info_plus/include/**/*.h'
```

packages/device_info_plus/device_info_plus/ios/device_info_plus/Package.swift
```diff
@@ -0,0 +1,27 @@
...
+            dependencies: [],
+            resources: [
+                .process("PrivacyInfo.xcprivacy"),
+            ],
+            cSettings: [
+                .headerSearchPath("include/device_info_plus")
+            ]
+        )
+    ]
+)
```
loic-sharma, line 27:
```suggestion
)

```

vbuberen (approved):
Tested on iOS and MacOS devices with both SPM on/off and everything works alright. 

Thanks for your contribution. Merging.

PR #3172, opened 2024-08-12
feat(network_info_plus): Add Swift Package Manager support

charafau:

Another spm support


miquelbeltran (commented):
LGTM, still want to test on my device before approval and merge

vbuberen (approved):
Fixed the privacy manifest location and declaration for MacOS. 
Unfortunately, I currently don't have a personal account with Wi-Fi capability, so couldn't test this one, but considering that all other PRs on this topic are alright, I believe we are good to go.

PR #3173, opened 2024-08-12
feat(connectivity_plus): Add Swift Package Manager support

charafau:

Another spm support

removed obsolete objc code, moved plugin to swift only, which avoid mixing swift project


packages/connectivity_plus/connectivity_plus/darwin/Classes/ConnectivityPlusPlugin.m
miquelbeltran, line 1: Would removing this code cause an issue on older apps still using Obj-C?

packages/connectivity_plus/connectivity_plus/example/pubspec.yaml
miquelbeltran, line 1: Changes in this file should be reverted

miquelbeltran (changes requested):
The example pubspec should not be changed.

Instead, use `melos bootstrap` as explained in the contributors document to setup the example project.

packages/connectivity_plus/connectivity_plus/darwin/Classes/ConnectivityPlusPlugin.m
charafau: I'm sure on 99% it won't. Example project works without changing anything, support for swift in plugin was here already.

charafau:
> The example pubspec should not be changed.
> 
> Instead, use `melos bootstrap` as explained in the contributors document to setup the example project.


Thank you Miguel, will educate myself with melon :)

vbuberen (approved):
Tested and everything works good. 

Thanks for another contribution!

PR #3258, opened 2024-09-17
Update SystemUUID.swift

fmatuszewski:
Fix For older Maxos platforms.
#available(macOS 12, *) is not available on runtime for older platforms it must be replaced by Compiletime directive #if MACOS12_OR_LATER

## Description

I have an older laptop and I can't compile this code because  if #available(macOS 12, *) { this function is not available on older mac platforms. Ater I replaced it with compile time directive it start to works.

### DIFF
--- a/packages/device_info_plus/device_info_plus/macos/Classes/SystemUUID.swift
+++ b/packages/device_info_plus/device_info_plus/macos/Classes/SystemUUID.swift
@@ -6,11 +6,13 @@ public struct SystemUUID {
         let dev = IOServiceMatching("IOPlatformExpertDevice")
         
         var platformExpert: io_service_t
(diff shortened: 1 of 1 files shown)


miquelbeltran:
Hello. I am not sure this change is correct, you are changing a runtime OS version check by a compilation flag.

As I understand, if you compile the code on an older machine and deploy it to a modern machine, it will break because it will use `kIOMasterPortDefault` when it should be using `kIOMainPortDefault`. And the same the other way around.

vbuberen:
I would appreciate if you check the provide PR template and fill in info as well as read contributors guide and name the PR properly. Respect other people's time.

miquelbeltran:
Closing since no follow up

PR #3270, opened 2024-09-30
feat(share_plus): add option to add custom link title on Android

florian-obernberger:
## Description

This adds the possibility to add a custom link title to URIs on Android. As far as I am aware this feature is not available on iOS, so I did not implement it. I only added it to the `shareURI` method, as this is already a mobile only and already has platform specific parameters (`sharePositionOrigin`).

## Related Issues

There is no open issue about it, I was thinking about opening one up but decided to implement it instead.

## Checklist

## Breaking Change

### DIFF
--- a/packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/MethodCallHandler.kt
+++ b/packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/MethodCallHandler.kt
@@ -29,6 +29,7 @@ internal class MethodCallHandler(
                         call.argument<Any>("uri") as String,
                         subject = null,
                         withResult = isWithResult,
+                        call.argument<Any>("title") as String?,
                     )
                     success(isWithResult, result)
                 }
@@ -38,6 +39,7 @@ internal class MethodCallHandler(
                         call.argument<Any>("text") as String,
                         call.argument<Any>("subject") as String?,
                         isWithResult,
+                        null,
                     )
                     success(isWithResult, result)
                 }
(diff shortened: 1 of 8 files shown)


miquelbeltran:
Hello and thanks for the contribution!

I will need to think about this, I'd like to find a better way to provide platform-exclusive parameters and/or provide Android intent extras.

florian-obernberger:
My pleasure!

I had a few different options in mind, but decided to stick to the existing design patterns. Some alternatives I had thought about were:

- Something like an `AndroidURIShareOptions` object that gets passed along
- A similar approach to [flutter_local_notifications](https://pub.dev/packages/flutter_local_notifications) where you can resolve to the platform specific implementation which has more features / parameters available in the method. This however would most likely require a complete restructuring of the package, since you'd be work with instance methods and not static ones

miquelbeltran:
Another advantage of allowing instantiation of the `Share` class is that it would allow mocking, which is a complaint we had in the past from using static methods.

I was also thinking on having something like `AndroidShareOptions` or similar. Since Android allows adding `Intent` extras, it could also include a generic `Map<String, String>` for extras so you can pass in any intent extras you want.

And moving the `sharePositionOrigin` to an iOS (and macOS?) share options object to keep it consistent.

 These options objects could be passed to the three share methods as well.

And another different question is the `subject` vs. `title`, as internally sometimes `subject` is used for "mailto" subject line, sometimes is passed as the `title` e.g. in web, and on Android is used as the intent `EXTRA_SUBJECT` which maybe is not what receiving apps expect.

So you see, it gets a bit messy :)

miquelbeltran:
As mentioned in #3306, I'd prefer if we can discuss this in a feature request ticket, and find a solution that fits most people. Then we can create a new PR.

miquelbeltran:
Created the ticket: https://github.com/fluttercommunity/plus_plugins/issues/3307

PR #3335, opened 2024-10-23
chore(all): Update the release process docs

vbuberen:
## Description

With recent Melos changes I now have to not only push the new branch to remote, but also specify the tracking information with git command in order to run `melos version...`. Without setting tracking information I see such exception:

Screenshot 2024-10-23 at 12 50 59

Update documentation to mention this requirement.


miquelbeltran (approved):
Well, that's annoying!

Issue #1099, opened 2022-09-27
[Request]: share_plus refactor and cleanup


miquelbeltran:
### Plugin

share_plus

### Use case

Currently, share_plus is a bit messy:

- 4 different methods that do somehow the same (`share`, `shareFiles`, `shareWithResult`, and `shareFilesWithResult`).
- Sharing URLs, something that iOS and macOS support, is currently broken.

### Proposal

The refactor should:

- Unify all four methods into a single one.
- Deprecate the old methods.
- Remove code duplication
- Add a way to share a URL on iOS (not as String, but as NSURL)

miquelbeltran:
Share files implementation for web to be done in another ticket: https://github.com/fluttercommunity/plus_plugins/issues/212

ueman:
Since file io isn't supported on web via dart:io it could make sense to use https://pub.dev/packages/cross_file in order to pass files to the share plugin.

miquelbeltran:
@ueman note that the web implementation is discussed in #212

ThexXTURBOXx:
I would happily take this issue on as part of Hacktoberfest, if possible and still free :)

miquelbeltran:
There's no ticket "reservation". Whoever gets the PR approved gets it.

ThexXTURBOXx:
Oh okay, I remembered last year when it worked like this. Sorry about that then!

ThexXTURBOXx:
I have developed a strategy on how to refactor the plugins. However, I still have two questions about this, because I am not sure which solution is desirable:
 - What should be the name of the unified method? Normally it would be called `share`, but this method already exists (as it is not deleted, but deprecated). I thought of `shareObject` or similar.
 - Should the platform interface and the platform-dependent implementations also be unified? I would have said that all `share_plus` related packages should be refactored, but if desired, I can also modify only the `share_plus` package and not the other packages.

Details of this are very welcome in order to achieve the desired refactoring.

ueman:
> * What should be the name of the unified method? Normally it would be called `share`, but this method already exists (as it is not deleted, but deprecated). I thought of `shareObject` or similar.

You could create a `ShareV2` class which would allow you to call the method just `share`. That would also make deprecation a bit easier, I guess.

miquelbeltran:
New class `ShareV2` or new method `shareV2` are both valid options indeed. Adding `@Deprecated` to the old methods then.

> Should the platform interface and the platform-dependent implementations also be unified?

[...]

PR #3372, opened 2024-11-27
feat(share_plus): support rich preview on the share sheet

### LINKED ISSUE #1221: [share_plus] [Request]: Thumbnail support
### Plugin

share_plus

### Use case

See image for expectation:

BBMRzLdUu9RE2mQ
Left: Flutter behavior (no thumbnail screenshot) Right: native behavior with thumbnail

This seems to be available on iOS. Not sure on Android or other platforms.

### Proposal

See native functionality above including thumbnail preview.

### LINKED ISSUE #3307: [Request]: (share_plus) `EXTRA_TITLE` in share Intent on Android
### Plugin

share_plus

### Use case

As a plugin user I'd like to be able to set the `EXTRA_TITLE` when sharing conent (URL, Text or Files) with `share_plus` when using Android.

### Proposal

See PRs #3306 and #3270.

Several options:

1. Add a `title` optional parameter. This is only used on Android.
2. Simply pass the `subject` as the `EXTRA_TITLE` intent extra.
3. Create a way to pass Intent extras directly (e.g. a `Map<String, String>`)
4. Create an "Android params" optional payload for the share methods.

Each one has advantages and disadvantages, I'd like to discuss them in this ticket.

slaci:
## Description

Support [rich preview](https://developer.android.com/training/sharing/send#adding-rich-content-previews) on Android with the help of the new `title` and `thumbnail` parameters. The `title` is also used for the Web Share API's `title` parameter in favor of the `subject`, but falls back to the `subject` for backward compatibility (as the previous code used the `subject` for this).

The contribution guide says features for only one platform won't be accepted, but I still want to try.

## Related Issues

- Fix #3307 
- Fix #1221
- Related #1099

## Checklist

## Breaking Change

### DIFF
--- a/packages/share_plus/share_plus/README.md
+++ b/packages/share_plus/share_plus/README.md
@@ -60,6 +60,22 @@ sharing to email.
 Share.share('check out my website https://example.com', subject: 'Look what I made!');
 ```
 
+The optional `title` and `thumbnail` parameters enable
+[rich content preview](https://developer.android.com/training/sharing/send#adding-rich-content-previews)
+on Android when sharing text.
+
+On the web the `title` or the `subject` (when the `title` is omitted) is passed to the
+[Web Share API](https://web.dev/web-share/)'s title parameter.
+
+```dart
+Share.share('Content which will be shared', title: 'Preview title', thumbnail: XFile('path/to/thumbnail.png'));
+```
+
+> [!CAUTION]
+> For the `thumbnail` parameter the
+> [Sharing data created with XFile.fromData](#sharing-data-created-with-xfilefromdata)
+> limitation has to be considered.
+
 `share()` returns `status` object that allows to check the result of user action in the share sheet.
 
 ```dart
--- a/packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/MethodCallHandler.kt
+++ b/packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/MethodCallHandler.kt
@@ -38,6 +38,8 @@ internal class MethodCallHandler(
                         call.argument<Any>("text") as String,
                         call.argument<Any>("subject") as String?,
                         isWithResult,
+                        title = call.argument<String?>("title"),
+                        thumbnailPath =  call.argument<String?>("thumbnailPath"),
                     )
                     success(isWithResult, result)
                 }
--- a/packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/Share.kt
+++ b/packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/Share.kt
@@ -2,11 +2,13 @@ package dev.fluttercommunity.plus.share
 
 import android.app.Activity
 import android.app.PendingIntent
+import android.content.ClipData
 import android.content.Context
 import android.content.Intent
 import android.content.pm.PackageManager
 import android.net.Uri
 import android.os.Build
+import android.util.Log
 import androidx.core.content.FileProvider
 import java.io.File
 import java.io.IOException
@@ -21,6 +23,10 @@ internal class Share(
     private var activity: Activity?,
     private val manager: ShareSuccessManager
 ) {
+    companion object {
+        const val TAG = "FlutterSharePlus"
+    }
+
     private val providerAuthority: String by lazy {
         getContext().packageName + ".flutter.share_provider"
     }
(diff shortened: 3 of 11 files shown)


miquelbeltran:
Hi @slaci thanks for the contribution!

I am not convinced about adding another two parameters to the method, specially `title`, which added to `text` and `subject` just becomes more confusing what does what on which platform.

The share method definitely needs a refactor and a parameter cleanup in some way. Maybe this PR is an opportunity to do that?

There is some discussion about that in #3307 

> The contribution guide says features for only one platform won't be accepted, but I still want to try.

That's true, maybe there is a way to provide a preview thumbnail on iOS as well?

vbuberen:
I like the idea of adding this functionality as it is something that Android offers by default. But agree with Miguel that we might think about the API more here.

slaci:
I can extract the thumbnail to a `SharePlusAndroidSettings` class or something. The `sharePositionOrigin` could go into an `IOS` or `Apple` one, but that would be a breaking change (or the parameter could be deprecated first of course).

The rest of the parameters are problematic, as the new `title` param is not the odd one, but the `subject`. Android uses the title for the `EXTRA_TITLE` parameter and the web uses it as the `title` parameter of the Web Share API. For iOS I may need some more research indeed, but as I see that could also use the title. There is the `thumbnailImageForActivityType` method which currently returns the image when an image is shared. It may be worth to check how this works (what happens if I pass a thumbnail to it when sharing text). When an `NSURL` is shared, then the `activityViewControllerLinkMetadata` callback runs and creates an `LPLinkMetadata` which contains a `title` + `thumbnail`, but this is sadly only for urls (as far as I understand).

slaci:
I have spent some hours checking the iOS thumbnail possibilities, but sadly I couldn't figure out when the mentioned [thumbnailImageForActivityType](https://developer.apple.com/documentation/uikit/uiactivityitemsource/1620462-activityviewcontroller) runs. 

There are multiple reports about this, but sadly without any answers: [SO](https://stackoverflow.com/questions/63367696/swift-5-thumbnailimageforactivitytype-method-not-called) | [Apple Forum](https://developer.apple.com/forums/thread/49274)

From the docs it seems like it depends on the target app (though not 💯 ), like the subject (eg. the [subjectForActivityType](https://developer.apple.com/documentation/uikit/uiactivityitemsource/1620455-activityviewcontroller) runs for the Mail app), but I couldn't find any app that would trigger the thumbnail callback or maybe it is totally bugged on iOS 18.

I haven't played with the title too much yet, but currently the `subject` is used for the `LPLinkMetadata`'s `title` when it is available which is incosistent with my changes. However it is only used for `shareXFiles` (because files are provided as NSURL-es, so they have this metadata) where I don't pass the `title` as providing a title when sharing files didn't do anything on Android nor web (will need to check what happens on iOS if anything in this case).

So if I could make the thumbnail work on iOS, then it could stay as a common option, otherwise I guess I would need to extract it to a Platform specific option class.

miquelbeltran:
Thanks for looking into it @slaci !

Re: params, I'm thinking that just having a single `ShareParams` class with all options may be the easiest, unify it all into a single `share(ShareParams)` method, and that should go all the way down. That will allow us to use any of the provided optional params no matter if sharing a uri, text, or files. We may need to add some asserts to enforce providing certain combination of params, but that would be it.

```dart
class ShareParams {
  final String? text;

  /// Used as share sheet title where supported (e.g. EXTRA_TITLE on Android)
  final String? extraTitle;

  /// Only used as email subject where supported (e.g. EXTRA_SUBJECT on Android)
  final String? subject;

  /// Only used in Android as preview thumbnail
  final XFile? previewThumbnail;

  /// Only used in iPad and Mac as share sheet origin
  final Rect? sharePositionOrigin;

  /// Share a URI, alternative to sharing [text]
  final Uri? uri;

  /// Share multiple files, can be used in combination with [text]
  final List<XFile>? files;
  final List<String>? fileNameOverrides;

  ShareParams({
    this.text,
    this.subject,
    this.extraTitle,
    this.previewThumbnail,
    this.sharePositionOrigin,
    this.uri,
    this.files,
    this.fileNameOverrides,
  });
}
```

Since we would be at it, I'd like to unify all methods into one, and to make `Share` class instantiable, instead of providing a static method (some people have asked this to be able to mock it)

miquelbeltran:
Added a comment on https://github.com/fluttercommunity/plus_plugins/issues/3307#issuecomment-2513723902 your feedback would be appreciated!

miquelbeltran:
FYI, started working on the shareplus refactor, that should simplify adding this and other optional parameters: https://github.com/fluttercommunity/plus_plugins/pull/3404

orkun1675:
Should this be re-opened since there is a code pointer to this PR?

The [previewThumbnail](https://github.com/fluttercommunity/plus_plugins/blob/bc6fafc8936f5734e2d1581fb7d7694547e70bfe/packages/share_plus/share_plus_platform_interface/lib/platform_interface/share_plus_platform.dart#L76) is also currently unused IIUC.

PR #3376, opened 2024-12-03
feat(share_plus): Added excludedActivityTypes support for the iOS platform.

### LINKED ISSUE #3377: [Request]: (share_plus) Add platform exclusion functionality for iOS.
### Plugin

share_plus

### Use case

I noticed that share_plus indeed lacks platform exclusion functionality for iOS. Specifically, this feature is outlined in [Apple's documentation](https://developer.apple.com/documentation/uikit/uiactivityviewcontroller/1622009-excludedactivitytypes). Currently, the plugin does not support this feature, but I believe it's quite useful. I also saw that some developers suggested this feature with fix messages like issus: (https://github.com/fluttercommunity/plus_plugins/issues/3075), 
but unfortunately, it hasn't been added yet.

### Proposal

I hope this feature can be added. I have submitted a request.

StanleyCocos:
The iOS platform is missing the exclusion feature. This PR adds the platform exclusion functionality for iOS.

fix: https://github.com/fluttercommunity/plus_plugins/issues/3377

## Checklist

## Breaking Change


miquelbeltran:
Thanks for this contribution as well! We have to keep this one on hold because I plan to refactor the `share()` method parameters, as also discussed in https://github.com/fluttercommunity/plus_plugins/pull/3372

Your input on this topic would be appreciated as well.

StanleyCocos:
@miquelbeltran 
Thank you for your reply. 
I understand your plan now. If needed, I can still improve this PR after your plan.

miquelbeltran:
FYI, started working on the shareplus refactor, that should simplify adding this and other optional parameters: https://github.com/fluttercommunity/plus_plugins/pull/3404

StanleyCocos:
Hi, that's great! I've been studying the code in your branch. It's really well-organized and tidy. Next, I need to add the parameter excludedActivityType for this pull request (PR) to ShareParams. After that, I'll wait for you to finish merging xx into the main branch. Once that's done, I'll synchronize my changes accordingly.

miquelbeltran:
Thanks! I haven't looked too deep into it yet, but can you describe how it works in this PR, so I can understand the change better. 

And as well, add a bit of documentation about this feature on the `README.md` so users are aware of it.

StanleyCocos:
Let me briefly describe the changes in this PR:

- In share_plus_platform_interface, I declared a CupertinoActivityType enum and added it as a parameter to ShareParams. Before passing it to the iOS layer, it is transmitted as a string to make it easier for users to exclude unwanted share options on iOS and macOS.

-  On the iOS native side, the string passed from the Flutter layer is received and converted back into an enum, which is then applied to the share configuration.

-  In the example, I added a multi-selection page for easier testing. Due to structural issues, I also introduced a new MyApp class that returns a MaterialApp (this caused a number of structural and formatting changes).

packages/share_plus/share_plus_platform_interface/test/share_plus_platform_interface_test.dart
```diff
@@ -95,11 +97,11 @@ void main() {
...
-          files: [XFile(fd.path)],
-          subject: 'some subject to share',
-          text: 'some text to share',
-          sharePositionOrigin: const Rect.fromLTWH(1.0, 2.0, 3.0, 4.0),
-        ),
+            files: [XFile(fd.path)],
+            subject: 'some subject to share',
+            text: 'some text to share',
+            sharePositionOrigin: const Rect.fromLTWH(1.0, 2.0, 3.0, 4.0),
+            excludedActivityType: null),
```
miquelbeltran, line 104: Trailing comma missing after `null`

packages/share_plus/share_plus/README.md
```diff
@@ -110,6 +110,17 @@ if (result.status == ShareResultStatus.dismissed) {
 }
 ```
 
+On iOS or macOS, if you want to exclude certain options from appearing in your share sheet,
+you can set the excludedActivityTypes array.
+For the list of supported excludedActivityTypes, you can refer to Apple’s [documentation](https://developer.apple.com/documentation/uikit/uiactivityviewcontroller/excludedactivitytypes/).
```
miquelbeltran, line 115:
```suggestion
you can set the `excludedActivityTypes` array.
For the list of supported `excludedActivityTypes`, you can refer to Apple's [documentation](https://developer.apple.com/documentation/uikit/uiactivityviewcontroller/excludedactivitytypes/).
```

Small formatting suggestion

miquelbeltran: Shouldn't devs check the `CupertinoActivityType` enum instead of the Apple docs?

packages/share_plus/share_plus_platform_interface/lib/platform_interface/share_plus_platform.dart
```diff
@@ -136,6 +136,12 @@ class ShareParams {
   ///   Parameter ignored on other platforms.
   final bool mailToFallbackEnabled;
 
+  /// Exclude services on iOS and macOS that you think are not suitable for your content.
+  ///
+  /// * Supported platforms: iOS, macOS
+  ///   Parameter ignored on other platforms.
+  final List<CupertinoActivityType>? excludedActivityType;
```
miquelbeltran, line 143: Since it's an apple exclusive param, can we name it so it references that? e.g. `excludedCupertinoActivities` since the type is `CupertinoActivityType`.

packages/share_plus/share_plus_platform_interface/test/share_plus_platform_interface_test.dart
```diff
@@ -64,23 +64,25 @@ void main() {
   test('sharing origin sets the right params', () async {
     await sharePlatform.share(
       ShareParams(
-        uri: Uri.parse('https://pub.dev/packages/share_plus'),
-        sharePositionOrigin: const Rect.fromLTWH(1.0, 2.0, 3.0, 4.0),
-      ),
+          uri: Uri.parse('https://pub.dev/packages/share_plus'),
+          sharePositionOrigin: const Rect.fromLTWH(1.0, 2.0, 3.0, 4.0),
+          excludedActivityType: [CupertinoActivityType.airDrop]),
```
miquelbeltran, line 69: Missing trailing comma

miquelbeltran (commented):
Just some comments, I like the overall idea. I will also have to test this somehow.

packages/share_plus/share_plus_platform_interface/lib/platform_interface/share_plus_platform.dart
StanleyCocos: Your suggestion is excellent — the naming alone makes the purpose clear to developers.

packages/share_plus/share_plus_platform_interface/test/share_plus_platform_interface_test.dart
StanleyCocos: Thank you, it’s been updated.

packages/share_plus/share_plus_platform_interface/test/share_plus_platform_interface_test.dart
StanleyCocos: Thank you, it’s been updated.

packages/share_plus/share_plus/README.md
StanleyCocos: You’re right, it should be based on CupertinoActivityType first.

StanleyCocos:
@miquelbeltran 
Sorry to bother you. I noticed this request has been pending for quite a while. Do you have any plans to complete the code review soon?

packages/share_plus/share_plus_platform_interface/lib/platform_interface/share_plus_platform.dart
```diff
@@ -203,3 +210,34 @@ enum ShareResultStatus {
...
+  markupAsPDF,
+  sharePlay,
+  collaborationInviteWithLink,
+  collaborationCopyLink,
+  addToHomeScreen,
+}
+
+extension CupertinoActivityTypeValue on CupertinoActivityType {
+  String get value => toString().split('.').last;
+}
```
miquelbeltran, line 243: What is this extension used for? I don't see it used in the PR

packages/share_plus/share_plus_platform_interface/lib/platform_interface/share_plus_platform.dart
```diff
@@ -203,3 +210,34 @@ enum ShareResultStatus {
   /// but the user action can not be determined
   unavailable,
 }
+
+/// An abstract class that you subclass to implement app-specific services
+/// for iOS and macOS.
+///
+/// https://developer.apple.com/documentation/uikit/uiactivity/activitytype
```
miquelbeltran, line 217: This comment doesn't seem correct to me, as these are enum values not abstract classes. Can this be reworded differently?

packages/share_plus/share_plus_platform_interface/lib/platform_interface/share_plus_platform.dart
StanleyCocos: It’s used at line 73 in method_channel_share. But I think turning it into a separate method might be a better approach now.

packages/share_plus/share_plus_platform_interface/lib/platform_interface/share_plus_platform.dart
StanleyCocos: You’re right — I may have overlooked updating the documentation here. It’s now been revised; could you please take another look?

packages/share_plus/share_plus/ios/share_plus/Sources/share_plus/FPPSharePlusPlugin.m
```diff
@@ -371,14 +428,26 @@ + (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar> *)registrar {
...
+  if (excludedActivityType.count > 0) {
+    NSMutableArray *excludedActivityTypes = [[NSMutableArray alloc] init];
+    for (NSString *type in excludedActivityType) {
+      UIActivityType activityType = activityTypeForString(type);
+      if (activityType != nil) {
+        [excludedActivityTypes addObject:activityType];
+      }
+    }
+    activityViewController.excludedActivityTypes = excludedActivityTypes;
+  }
```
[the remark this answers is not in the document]
StanleyCocos: Thank you for your suggestion — I’ve made the adjustments. Much appreciated!

packages/share_plus/share_plus_platform_interface/lib/method_channel/method_channel_share.dart
```diff
@@ -67,6 +67,13 @@ class MethodChannelShare extends SharePlatform {
       map['mimeTypes'] = mimeTypes;
     }
 
+    if (params.excludedCupertinoActivities != null &&
+        params.excludedCupertinoActivities!.isNotEmpty) {
+      final activityTypes =
+          params.excludedCupertinoActivities!.map((e) => e.value).toList();
+      map['excludedActivityType'] = activityTypes;
+    }
+
```
miquelbeltran, line 76: Sorry, I'll have to request another change. We renamed the variable `excludedCupertinoActivities` as it previously was `excludedActivityType`. However, I think it would be good if the map key would also be the same, to avoid confusion with that. I missed that when I first reviewed.

miquelbeltran (changes requested):
Requesting one more change, and I think this would be completed. Thanks for keeping up with the PR!

miquelbeltran:
Tested on iOS 18.5 simulator using the example app ✅

packages/share_plus/share_plus_platform_interface/lib/method_channel/method_channel_share.dart
StanleyCocos: Thank you for your suggestion. 

Slightly adjusted, but I still have some other thoughts.

I have a slightly different idea about this part. I think using excludedCupertinoActivities as the key name in the map may not be entirely appropriate, because accessing a key with “Cupertino” in its name on the iOS side feels a bit odd.

Also, I think if Android were to support a similar feature, we should still use this same key name to pass it down (even though Android doesn’t support it directly, it can potentially be handled by excluding package names).
If we do implement it for Android, I think we could introduce a new List<String> parameter for excluded package names.

miquelbeltran:
Thanks for the changes, I think it's good to go now!

miquelbeltran (approved):

PR #3394, opened 2024-12-12

vbuberen:


packages/device_info_plus/device_info_plus/example/pubspec.yaml
```diff
@@ -1,11 +1,10 @@
 name: device_info_plus_example
 description: Demonstrates how to use the device_info_plus plugin.
-version: 1.0.0
```
vbuberen, line 3: Removed version added in #3358 because Melos tries to increment this one as well. This time tried to make it 2.0.0

miquelbeltran (approved):
Thanks!

PR #3395, opened 2024-12-12
chore(docs): Simplify release creation guide

vbuberen:
## Description

Found a much simpler way to version packages with Melos by starting versioning on the `main` branch. 
Tested with #3394 - seems to be working.


miquelbeltran (approved):
Much better, yes

PR #3408, opened 2024-12-19
fix(network_info_plus): Improve Wi-Fi IP address retrieval on iOS to avoid null

### LINKED ISSUE #1652: [Question]: Problem with NetworkInfo().getWifiIP(); return null on iPhone 14 Pro and iOS 16.3.1
### What is your question?

Hi,
I need help to solve the problem because I have already searched everything and tried everything I could and I can't get the app to work correctly on iphone 14 pro devices with iOS 16.3.1

The problem occurs in the function that returns the IP of the Wi-Fi network NetworkInfo().getWifiIP().
I tried adding more permissions via plist.info but it still doesn't work.
It always returns a null value.

The code snippet I use is very simple:

`Future<Estados> _verificarConexionWifi() async {
    ipAddress = await NetworkInfo().getWifiIP();
    if (ipAddress == null) {
       estadoDeConexion = Estados.sinWifi;
    }
    return estadoDeConexion;
}`

The app has been published for a long time and we have no reports of problems on other devices.
I asked Apple to allow me permission to use multicast messages just in case, but it did not solve the problem.

I don't know what else to do and that is why I am writing to see if something similar is happening to someone and they already solved it to find out how they did it.

Thank you very much in advance for any help you can give me.

Unfortunately I don't have the debug log because we don't have an iphone 14 Pro and with the devices we have we can't recreate the problem.

### Checklist before submitting a question

- [X] I am using the latest version of the plugin
- [X] I did a `flutter clean`

vbuberen:
## Description

Based on information provided in https://github.com/fluttercommunity/plus_plugins/issues/1652#issuecomment-2373580942 changed the iOS implementation to retrieve IP address to get values even when device uses not `en0` interface for Wi-Fi.

Additionally did minor improvements with log entry when interfaces info retrieval fails and changed the comparison to simplify it a bit.

## Related Issues

Fixes #1652

## Breaking Change

### DIFF
--- a/packages/network_info_plus/network_info_plus/example/ios/Runner/AppDelegate.swift
+++ b/packages/network_info_plus/network_info_plus/example/ios/Runner/AppDelegate.swift
@@ -1,7 +1,7 @@
 import UIKit
 import Flutter
 
-@UIApplicationMain
+@main
 @objc class AppDelegate: FlutterAppDelegate {
   override func application(
     _ application: UIApplication,
--- a/packages/network_info_plus/network_info_plus/ios/network_info_plus/Sources/network_info_plus/FPPNetworkInfoPlusPlugin.m
+++ b/packages/network_info_plus/network_info_plus/ios/network_info_plus/Sources/network_info_plus/FPPNetworkInfoPlusPlugin.m
@@ -69,6 +69,7 @@
   __block NSString *addr = nil;
   [self enumerateWifiAddresses:AF_INET
                     usingBlock:^(struct ifaddrs *ifaddr) {
+                      if (addr) return;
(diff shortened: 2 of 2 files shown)


miquelbeltran (approved):

marcos930807:
After this merge the Method NetworkInfo().getWifiIP(); is giving me the Wrong ip address. Attached are Same method breakpoint on version 6.1.2 and 6.1.3. The real ip addres is 10.0.0.89 (version 6.1.2) the bad one is ("169.254.109.51")
Screenshot 2025-04-03 at 11 02 46 PM
Screenshot 2025-04-03 at 10 58 41 PM

PR #3409, opened 2024-12-22
docs(connectivity_plus): update requirements

mernen:
## Description

This is merely keeping the readme up to date with what was described in the changelog for version 6.0.1.

## Checklist

## Breaking Change

### DIFF
--- a/packages/connectivity_plus/connectivity_plus/README.md
+++ b/packages/connectivity_plus/connectivity_plus/README.md
@@ -22,8 +22,8 @@ This plugin allows Flutter apps to discover network connectivity types that can 
 
 ## Requirements
 
-- Flutter >=3.7.0
-- Dart >=3.2.0 <4.0.0
+- Flutter >=3.19.0
+- Dart >=3.3.0 <4.0.0
 - iOS >=12.0
 - MacOS >=10.14
 - Android `compileSDK` 34


miquelbeltran:
Thanks!

fyi @vbuberen This is probably already sdk >=3.4.0 since that's required by `web`. Does it make sense to keep listing these requirements in the README? And, should we bump the min sdk requirement in the pubspec of the plugin to align with web and the README?

miquelbeltran (approved):

vbuberen:
> This is probably already sdk >=3.4.0 since that's required by web

But we are still declaring version range for `web` from `0.5.0`, which didn't need Dart 3.4.0 https://github.com/fluttercommunity/plus_plugins/blob/main/packages/connectivity_plus/connectivity_plus/pubspec.yaml#L44 Or I am missing something here? In any case, I would bump min web to 1.0.0 anyway already and update the requirements afterwards. Suggest to do it in another PR after merging this one.

> Does it make sense to keep listing these requirements in the README

I believe it makes sense for convenience of both users and us, maintainers, especially when we need to point the reporter of the issue related to wrong environment, like we regularly have.

miquelbeltran:
> But we are still declaring version range for `web` from `0.5.0`,

Oh, I didn't notice that.

> In any case, I would bump min web to 1.0.0 anyway already and update the requirements afterwards. Suggest to do it in another PR after merging this one.

Yep! agree.

> I believe it makes sense for convenience of both users and us, maintainers, especially when we need to point the reporter of the issue related to wrong environment, like we regularly have.

All good then!

Issue #3186, opened 2024-08-19
[Question]: Huawei and Honor models, can the text in the middle of the pop-up window be modified?


SheepYang1993:
### What is your question?

![6bc1814f3ceb0e2c81cb19685769301](https://github.com/user-attachments/assets/de62a653-5285-41a8-8958-0ad0a52d0069)
如上图所示

### Checklist before submitting a question

- [X] I searched issues in this repository and couldn't find such bug/problem
- [X] I Google'd a solution and I couldn't find it
- [X] I searched on StackOverflow for a solution and I couldn't find it
- [X] I read the README.md file of the plugin
- [X] I am using the latest version of the plugin
- [X] All dependencies are up to date with `flutter pub upgrade`
- [X] I did a `flutter clean`
- [X] I tried running the example project

vbuberen:
Please use translator and write on English, so others can understand without need to use translation tools.

SheepYang1993:
Huawei and Honor models, can the text in the middle of the pop-up window be modified?
On the left is the Honor phone.
On the right is a Huawei phone.

PenguinBIUBIUBIU:
same question, the ui for huawei is really ugly

miquelbeltran:
The plugin doesn't have the capacity to change the phone OS UI, as it only calls to standard Android OS APIs.

This should be reported to Huawei instead.

OnClickListener2048:
same here

Issue #1588, opened 2023-03-03
分享pdf文件


qjb2404:
### What is your question?

华为鸿蒙系统分享不了图片，文件，有对应的解决办法没

### Checklist before submitting a question

- [X] I Google'd a solution and I couldn't find it
- [X] I searched on StackOverflow for a solution and I couldn't find it
- [X] I read the README.md file of the plugin
- [X] I am using the latest version of the plugin
- [X] All dependencies are up to date with `flutter pub upgrade`
- [X] I did a `flutter clean`
- [X] I tried running the example project

vbuberen:
Would ask you to use English language for your question

qjb2404:
Huawei HarmonyOS2.0 cannot share pictures and files. Is there any solution? "The direct share function cannot be used" is displayed.

Issue #1030, opened 2022-08-19
[Question]: I'm using share_plus to share on Android and the UI is inconsistent with other apps.


Leheih:
### What is your question?

The following is my UI using share_plus：
<img width="395" alt="image" src="https://user-images.githubusercontent.com/17791894/185590046-63044a2d-9026-400c-83c5-e32a7a6b658d.png">

Below is the sharing UI for another app on the same phone
<img width="414" alt="image" src="https://user-images.githubusercontent.com/17791894/185590505-a4fd4a7c-ee2e-4d26-86c9-f6c7e30ba00a.png">

miquelbeltran:
share_plus uses `Intent.createChooser` from Android internally, which is what is being used in the official Android docs: https://developer.android.com/training/sharing/send

tbh I am not sure what other apps may be doing. Any ideas?

LuisMiguelSS:
Perhaps the other apps include more intents, which add up the amount of results shown in the intent, thus having different results. Intent choosers should always be the same under the same circunstamces.

ueman:
I guess this is because `share_plus` doesn't make use of [`Rich content previews`](https://developer.android.com/training/sharing/send#adding-rich-content-previews) on Android.

SirusCodes:
Some other apps create their own chooser and use intents to directly target those apps.

Found [this](https://medium.com/@amirdes021_73588/android-custom-chooser-for-intent-share-send-358ee8cbc152) medium article to do so.

SirusCodes:
> I guess this is because `share_plus` doesn't make use of [`Rich content previews`](https://developer.android.com/training/sharing/send#adding-rich-content-previews) on Android.

It's not because of that but it would be a good idea to have support for Rich content previews 🤔

I would love to contribute that for hacktoberfest

Coronon:
@SirusCodes I would refrain from such a change at the moment as there was an announcement that a large portion of the plus plugins may be refactored. When we do add this, we should also add it to other platforms.

See #1221

miquelbeltran:
> I would refrain from such a change at the moment as there was an announcement that a large portion of the plus plugins may be refactored. When we do add this, we should also add it to other platforms.

There's nothing going on at the moment.

ipcjs:
you can use [huawei_share](https://pub.dev/packages/huawei_share) 😅

ipcjs:
👀

PR #3417, opened 2025-01-02
fix(share_plus) sharing content on huawei or honor vendor which changed the default intent of Intent.ACTION_SEND

OnClickListener2048:
fix(share_plus) sharing content on huawei or honor vendor which changed the default intent of Intent.ACTION_SEND

## Description

for those OEM vendor brand that changed default choose intent of  Intent.ACTION_SEND, I fixed Huawei and Honor vendor share intent

## Related Issues

- *Related #3186*
- *Related #1588*
- *Related #1030*
## Checklist

- [ ] I read the [Contributor Guide](https://github.com/fluttercommunity/plus_plugins/blob/main/CONTRIBUTING.md) and followed the process outlined there for submitting PRs.
- [ ] I titled the PR using [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0).
- [ ] I did not modify the `CHANGELOG.md` nor the plugin version in `pubspec.yaml` files.

## Breaking Change

### DIFF
--- a/packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/Share.kt
+++ b/packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/Share.kt
@@ -81,7 +81,11 @@ internal class Share(
             } else {
                 Intent.createChooser(shareIntent, null /* dialog title optional */)
             }
-        startActivity(chooserIntent, withResult)
+        startActivity(chooserIntent.apply {
+            getAvailableOemChooser()?.let {
+                action = it.action
+            }
+        }, withResult)
     }
 
     @Throws(IOException::class)
@@ -155,7 +159,11 @@ internal class Share(
                 )
             }
         }
-        startActivity(chooserIntent, withResult)
+        startActivity(chooserIntent.apply {
+            getAvailableOemChooser()?.let {
+                action = it.action
+            }
+        }, withResult)
     }
 
     private fun startActivity(intent: Intent, withResult: Boolean) {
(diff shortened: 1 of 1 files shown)


miquelbeltran:
Hello, I took a quick look, overall I understand why would you need this in the share_plus, but I have some concerns:

- Are `com.huawei.android.internal.app` and `com.huawei.intent.action.hwCHOOSER` documented somewhere? Is there an API spec somewhere? My biggest issue is that this seems like a hack, plus no one in the maintainer team has a device from those manufacturers, so we don't have a way to maintain this.
- `getAvailableOemChooser` runs every time a share action is done, and instead it should be optional, and IMO the option disabled by default.
- The change wouldn't be compatible with the refactor in progress, but that's not a big problem: https://github.com/fluttercommunity/plus_plugins/pull/3404

I'd prefer if we could make setting the intent action as a share parameter, or even the "package-name - intent" pairs something that can be passed as a parameter (following the changes in #3404) so at least we can give the option for those devs that want to target the custom share sheets from the different manufacturers.

OnClickListener2048:
> Hello, I took a quick look, overall I understand why would you need this in the share_plus, but I have some concerns:
> 
> * Are `com.huawei.android.internal.app` and `com.huawei.intent.action.hwCHOOSER` documented somewhere? Is there an API spec somewhere? My biggest issue is that this seems like a hack, plus no one in the maintainer team has a device from those manufacturers, so we don't have a way to maintain this.
> * `getAvailableOemChooser` runs every time a share action is done, and instead it should be optional, and IMO the option disabled by default.
> * The change wouldn't be compatible with the refactor in progress, but that's not a big problem: [feat(share_plus): SharePlus refactor #3404](https://github.com/fluttercommunity/plus_plugins/pull/3404)
> 
> I'd prefer if we could make setting the intent action as a share parameter, or even the "package-name - intent" pairs something that can be passed as a parameter (following the changes in #3404) so at least we can give the option for those devs that want to target the custom share sheets from the different manufacturers.

Thank you for your fine response for my last pr and i realized this change would be unfair for other kind system and the right way is passing paramaters for those OES OS, but i think the problem is really exist and does effect some developers especially in China,so do you think i just edit this pr or i should close this pr and reopen a new one after i repair this ?

miquelbeltran:
I will close this PR for the time being. If you can, write a proposal on a new ticket, and it should be implemented over the refactor in #3404

PR #3422, opened 2025-01-06
ci: Automatically lock issues and PRs older than a year

miquelbeltran:
## Description

This GH action closes PRs and Issues that have been closed and without activity for a year.

This is to avoid spam on older issues and force users to create new tickets if they have problems.

The default is 365 days since last activity. We can lower this value to something shorter e.g. 90 days.

## Related Issues

- None

## Checklist

## Breaking Change

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -30,8 +30,5 @@ jobs:
           exempt-issue-labels: enhancement
           exempt-pr-labels: keep-open
           operations-per-run: 1000
-      - name: Lock closed issues and PRs (defaults to 365 days)
-        uses: dessant/lock-threads@v5.0.1
-        with:
-          process-only: 'issues, prs' # Ignore Discussions
-          
+      - name: Lock closed issues and PRs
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -26,3 +30,8 @@ jobs:
           exempt-issue-labels: enhancement
           exempt-pr-labels: keep-open
           operations-per-run: 1000
+      - name: Lock closed issues and PRs (defaults to 365 days)
```
vbuberen, line 33: Nit: I suggest to change the name a bit, so it looks something like this:
```
# Defaults to 365
- name: Lock old issues and PRs
```

vbuberen (approved):
Do we want to add messages as well to locking action?

miquelbeltran:
> Do we want to add messages as well to locking action?

The other option is locking "discussions" but we have those disabled already.

miquelbeltran:
I simplified the CI script, removed the explicit `process-only` and as well the mention on the default inactive days value

PR #3431, opened 2025-01-13

vbuberen:
## Description

First release of 2025 with a few minor improvements


packages/share_plus/share_plus/example/ios/Runner/AppDelegate.swift
```diff
@@ -1,7 +1,7 @@
 import UIKit
 import Flutter
 
-@UIApplicationMain
+@main
```
miquelbeltran, line 4: Maybe included by mistake

miquelbeltran (approved):

PR #3434, opened 2025-01-15
feat(package_info_plus): add install time

Andrew-Bekhiet:
## Description

Adds package install time to iOS and Android

- On Android: uses `PackageManager.firstInstallTime`
- On iOS: gets the creation date of the app's documents directory
   - The documents directory should be created automatically when the app is installed, based on this [SO answer](https://stackoverflow.com/questions/4090512/how-to-determine-the-date-an-app-is-installed-or-used-for-the-first-time)

## Checklist

## Breaking Change

### DIFF
--- a/packages/package_info_plus/package_info_plus/android/src/main/kotlin/dev/fluttercommunity/plus/packageinfo/PackageInfoPlugin.kt
+++ b/packages/package_info_plus/package_info_plus/android/src/main/kotlin/dev/fluttercommunity/plus/packageinfo/PackageInfoPlugin.kt
@@ -39,6 +39,7 @@ class PackageInfoPlugin : MethodCallHandler, FlutterPlugin {
                 val buildSignature = getBuildSignature(packageManager)
 
                 val installerPackage = getInstallerPackageName()
+                val installTimeMillis = getInstallTimeMillis()
 
                 val infoMap = HashMap<String, String>()
                 infoMap.apply {
@@ -48,6 +49,7 @@ class PackageInfoPlugin : MethodCallHandler, FlutterPlugin {
                     put("buildNumber", getLongVersionCode(info).toString())
                     if (buildSignature != null) put("buildSignature", buildSignature)
                     if (installerPackage != null) put("installerStore", installerPackage)
+                    if (installTimeMillis != null) put("installTime", installTimeMillis.toString())
                 }.also { resultingMap ->
                     result.success(resultingMap)
                 }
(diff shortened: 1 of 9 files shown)


miquelbeltran:
Thanks! I took a quick look and looks promising, thanks for taking the time implementing tests as well.

Some ideas to extend this:

- Supporting web and Linux: This could be done looking at the creation or modification date of the `version.json` file, although this may depend on how the file was copied/if the OS preserved the creation/mod date, etc. but could give a rough idea.
- Supporting MacOS: Should be the same as with iOS?
- Supporting Windows: Similar to the web/linux implementation, but with the app .exe (this is how it reads the version atm)

Andrew-Bekhiet:
- Linux: Dart doesn't provide an API for checking file creation date, so I guess I'll use `stat` command or something similar
- Web: I'm not sure how would that work, as version.json would be the same for all users, right?
- MacOS: Yeah we can do the same as iOS. I'm not very experienced with Apple ecosystem, but how about the creation date of the app bundle? Would that work? I'm going to try it anyways
- Windows: I think there is a method that checks the windows registry, if that fails we can check the creation date of the app exe file
 
At first I didn't want to implement this for desktop OSes because the user could easily change files attributes and mess with this, so this shouldn't be used for anything critical like determining a free trial period or anything similar

In these cases developers are encouraged to implement this feature in their apps as it fits their needs

miquelbeltran:
thanks for the recent pushes! let me know when it's ready for review

Andrew-Bekhiet:
I think it is ready, here is a summary of what I decided to do when implementing this for desktop:

- Windows: check the creation date of the app executable, return last modified date if creation date is unavailable as some filesystems don't track creation date
- Linux: get the creation time (birth) of the app executable or fallback to modification date as some old linux kernels don't keep track of creation date
- MacOS: same implementation as iOS, checks the creation date of the app's `NSDocuments` directory

I have no idea how can this be implemented on web using only `version.json` because the `version.json` file will be the same for all users of the app, and I can't think of a way to get any metadata about it

However, It can be implemented by storing the first launch time in local storage or something similar, but I think this should be left out for the developers to choose how they want to implement it

What do you think about this? Do you have any suggestions?

miquelbeltran:
That's fine! Thanks for implementing the other platforms as well. Perhaps install time on web doesn't make much sense, as opposed to the other platforms, the user never installs anything, so it's fair to say this feature is not supported on web.

miquelbeltran:
The emulator job is very unstable so it's not mandatory to pass the review. I just need time to look at this (it may take a while)

miquelbeltran (approved):
Thanks specially for investing the time researching support for all platforms and implementing the unit and integration tests!

miquelbeltran:
Tested on Android and Linux, I don't have an Apple computer with me atm, but code seems fine and the integration tests on CI work

durant:
On macOS, apps need authorization to access the Documents directory, This modification will affect app on macOS

miquelbeltran:
@Andrew-Bekhiet can you confirm this? I didn't see this problem with the example app when I tested the feature.

Andrew-Bekhiet:
I don't have access to a Mac machine and I don't have much experience with Apple ecosystem
But after reading through the docs I think it should be fine. The app is only accessing attributes of the documents directory in it's sandbox

But I could be wrong, so I think someone should test this in release mode on a Mac

durant:
image
This prompt appears after upgrading to the latest version, because of getting the attribute of documents 
image

PR #3452, opened 2025-01-31
fix(android_intent_plus): adds error catching and forwarding

ViscousPot:
## Description
Currently, when sending an intent, any native errors occur silently without any notification to the flutter side of code.
This adds error forwarding back to flutter through the MethodCaller.
This is achieved by wrapping the whole caller in a try/catch

This can be tested by trying to send an intent to an Activity that doesn't exist.
- Previous behaviour, it would show a native error in the log but this is not catchable in flutter code
- New behaviour forwards the error and allows it to be caught

## Related Issues
No related issues exist for this

## Checklist

## Breaking Change


miquelbeltran (approved):
Thank you! LGTM

Issue #3473, opened 2025-02-13
[Bug]: device_info_plus memory leak in macOS


codingiran:
### Platform

macOS 15.3.1

### Plugin

device_info_plus

### Version

11.3.0

### Flutter SDK

3.27.4

### Steps to reproduce

Call `DeviceInfoPlugin().macOsInfo` in macOS cause memory leak, code stack see the picture below
![Image](https://github.com/user-attachments/assets/85539d80-03aa-47ec-9980-ce52d84256d5)

The issue lies with this line of code: https://github.com/fluttercommunity/plus_plugins/blob/50d12990f88b66c8a1c76426295cd28ad11e3194/packages/device_info_plus/device_info_plus/macos/device_info_plus/Sources/device_info_plus/SystemUUID.swift#L17

```swift
public static func getSystemUUID() -> String? {
        let dev = IOServiceMatching("IOPlatformExpertDevice")
        
        var platformExpert: io_service_t
        if #available(macOS 12, *) {
            platformExpert = IOServiceGetMatchingService(kIOMainPortDefault, dev)
        } else {
            platformExpert = IOServiceGetMatchingService(kIOMasterPortDefault, dev)
        }
        
        let serialNumberAsCFString = IORegistryEntryCreateCFProperty(platformExpert, kIOPlatformUUIDKey as CFString, kCFAllocatorDefault, 0)
        IOObjectRelease(platformExpert)
        let ser: CFTypeRef? = serialNumberAsCFString?.takeUnretainedValue()
        if let result = ser as? String {
            return result
        }
        return nil
    }
```

[takeUnretainedValue()](https://developer.apple.com/documentation/swift/unmanaged/takeunretainedvalue()) not consuming an unbalanced retain of `CFTypeRef`. It will cause `CFTypeRef` without [autorelease](https://developer.apple.com/documentation/objectivec/1418956-nsobject/1571951-autorelease).

Fix is simple, just replace [takeUnretainedValue()](https://developer.apple.com/documentation/swift/unmanaged/takeunretainedvalue()) with [takeRetainedValue()](https://developer.apple.com/documentation/swift/unmanaged/takeretainedvalue())

### Code Sample

```dart
import 'package:device_info_plus/device_info_plus.dart';

final deviceInfo = await DeviceInfoPlugin().macOsInfo;
```

### Logs

```shell
Not required
```

### Flutter Doctor

```shell
flutter doctor -v
[...]

PR #3474, opened 2025-02-13
fix(device_info_plus): fix memory leak when calling DeviceInfoPlugin().macOsInfo

codingiran:
## Description

Fix memory leak when call DeviceInfoPlugin().macOsInfo in macOS.

## Related Issues

[[Bug]: device_info_plus memory leak in macOS](https://github.com/fluttercommunity/plus_plugins/issues/3473)

## Checklist

## Breaking Change

### DIFF
--- a/packages/device_info_plus/device_info_plus/macos/device_info_plus/Sources/device_info_plus/SystemUUID.swift
+++ b/packages/device_info_plus/device_info_plus/macos/device_info_plus/Sources/device_info_plus/SystemUUID.swift
@@ -1,23 +1,16 @@
 import Foundation
 
-public struct SystemUUID {
-    
+public enum SystemUUID {
     public static func getSystemUUID() -> String? {
-        let dev = IOServiceMatching("IOPlatformExpertDevice")
-        
-        var platformExpert: io_service_t
-        if #available(macOS 12, *) {
-            platformExpert = IOServiceGetMatchingService(kIOMainPortDefault, dev)
+        let mainPort: mach_port_t
+        if #available(macOS 12.0, *) {
+            mainPort = kIOMainPortDefault
         } else {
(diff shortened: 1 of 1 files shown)


packages/device_info_plus/device_info_plus/macos/device_info_plus/Sources/device_info_plus/SystemUUID.swift
```diff
@@ -1,23 +1,16 @@
 import Foundation
 
-public struct SystemUUID {
-    
+public enum SystemUUID {
```
miquelbeltran, line 3: Why is there a change from `struct` to `enum`?

miquelbeltran (commented):
Hello! Thanks for the contribution. The refactor and the change seem fine to me, but before approving I'd need to test it and also fix CI because seems like something broke with the installation scripts, that may delay this PR for a while.

I've also got a question regarding the use of `enum`.

packages/device_info_plus/device_info_plus/macos/device_info_plus/Sources/device_info_plus/SystemUUID.swift
codingiran: This API is designed as `SystemUUID.getSystemUUID()`, essentially just placing `getSystemUUID()` under the `SystemUUID` namespace.

There’s no need to declare a separate `class` or `struct`—using an `enum` is the most lightweight way to achieve this in Swift. 

This is a common practice and also the default rule ([enumNamespaces](https://github.com/nicklockwood/SwiftFormat/blob/ca41c9751ded73223f3c6de54679247f16c27013/README.md?plain=1#L933C8-L933C22)) in [SwiftFormat](https://github.com/nicklockwood/SwiftFormat).

miquelbeltran (approved):
Thank you! tested on M1 Pro

Issue #3496, opened 2025-02-25
[Compile]: ComputerNameDnsFullyQualified isn't defined for the class DeviceInfoPlusWindowsPlugin


ThomasAunvik:
### Operating System

Linux

### Plugin

device_info_plus

### Version

11.3.1

### Flutter SDK

3.29.0

### Compilation command

flutter build appbundle
flutter build ipa

Temporary fix, downgrade to 11.3.0.

Web build works

### Dart Dependencies

```yaml

```

### Native dependencies

```text

```

### Compilation Error

```shell
../../../.pub-cache/hosted/pub.dev/device_info_plus-11.3.1/lib/src/device_info_plus_windows.dart:138:23: Error: The getter 'ComputerNameDnsFullyQualified' isn't defined for the class 'DeviceInfoPlusWindowsPlugin'.
 - 'DeviceInfoPlusWindowsPlugin' is from 'package:device_info_plus/src/device_info_plus_windows.dart' ('../../../.pub-cache/hosted/pub.dev/device_info_plus-11.3.1/lib/src/device_info_plus_windows.dart').
Try correcting the name to the name of an existing getter, or defining a getter or field named 'ComputerNameDnsFullyQualified'.
    GetComputerNameEx(ComputerNameDnsFullyQualified, nullptr, nSize);
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../../../.pub-cache/hosted/pub.dev/device_info_plus-11.3.1/lib/src/device_info_plus_windows.dart:144:29: Error: The getter 'ComputerNameDnsFullyQualified' isn't defined for the class 'DeviceInfoPlusWindowsPlugin'.
 - 'DeviceInfoPlusWindowsPlugin' is from 'package:device_info_plus/src/device_info_plus_windows.dart' ('../../../.pub-cache/hosted/pub.dev/device_info_plus-11.3.1/lib/src/device_info_plus_windows.dart').
Try correcting the name to the name of an existing getter, or defining a getter or field named 'ComputerNameDnsFullyQualified'.
          GetComputerNameEx(ComputerNameDnsFullyQualified, lpBuffer, nSize);
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Target kernel_snapshot_program failed: Exception
```

### Flutter Doctor

```shell
[...]

PR #3500, opened 2025-02-26
ci(all): check downgrade ci job

### LINKED ISSUE #3178: [Request]: Add dependencies versions downgrade into CI/CD pipelines
### Plugin

All

### Use case

After a recent release of plugins I saw on [the plugin's score page](https://pub.dev/packages/network_info_plus/score) on pub.dev that for `network_info_plus` we might have problems with older versions of `win32` dependency, which we allow in pubspec.yaml. 

Here is a screenshot to not lose the exact report:
Screenshot 2024-08-13 at 13 21 47

### Proposal

To avoid such problems we should add pipeline jobs to validate that everything works fine even with lowest allowed versions of dependencies. 
Here is one of suggested ways to test: https://dart.dev/tools/pub/dependencies#test-with-downgraded-dependencies
There is also a Github Action, which uses [`pana`](https://pub.dev/packages/pana) (the analyzer used on pub.dev) and which can help us run the checks on CI if we don't go the manual way: https://github.com/axel-op/dart-package-analyzer

### LINKED ISSUE #3498: [Request]: CI check `flutter pub downgrade`
### Plugin

all

### Use case

Would like to set up a CI job that runs `dart pub downgrade` and then `dart analyze` to check that the plugins still compile when using dependencies that are not up-to-date.

See https://dart.dev/tools/pub/cmd/pub-downgrade 

### Proposal

See above

miquelbeltran:
## Description

This PR adds CI jobs to check packages when running `dart pub downgrade` on them.

## Related Issues

- Closes #3178
- Related #3496

### DIFF
--- a/.github/workflows/all_plugins.yaml
+++ b/.github/workflows/all_plugins.yaml
@@ -107,7 +107,7 @@ jobs:
         run: melos run test:web --no-select
 
   analyze:
-    name: "Analyze all plugins"
+    name: "Analyze all with fatal infos"
     runs-on: ubuntu-latest
     timeout-minutes: 40
     steps:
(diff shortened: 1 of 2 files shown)


miquelbeltran:
Before merging this, I will fix/ignore all the infos, most of them come from examples anyway so they can be ignored.

miquelbeltran:
canceling this, will do a different approach

PR #3533, opened 2025-03-28
fix(android_intent_plus): startForegroundService on SDK higher than 26 

ViscousPot:
## Description
Currently, when starting a foreground service or a non-foreground service while the application is in the background, the application fails with 

`not allowed to start service intent app is in background`

This can be tested by trying to start a service where the application isn't in the foreground

- Previous behaviour, it would give the error above
- New behaviour sends the intent as expected and starts the service

## Related Issues
No related issues exist for this

## Checklist

## Breaking Change


miquelbeltran:
Thanks for looking into this! Could you add an example to the example that runs a foreground service? I'm trying to understand when this would be necessary, and I would like to test before approving. Also, the PR is set as Draft, not sure if that is intentional.

PR #3664, opened 2025-09-06
fix(android): Prevent potential crash in checkPowerServiceSaveMode

Darkx-dev:
### Description

This PR improves the null safety of the `checkPowerServiceSaveMode` method on the Android platform to prevent a potential crash.

### The Problem

Currently, the `checkPowerServiceSaveMode` method uses a non-null assertion (`!!`) on `applicationContext`. While the context is very unlikely to be null in normal operation, a library should be as resilient as possible against rare edge cases (e.g., race conditions during plugin detachment). A `NullPointerException` here would crash the entire host application.

### The Solution

I have replaced the `!!` operator with a safe-call (`?.let`) block. If `applicationContext` happens to be null, the method will now safely return a default value of `false` instead of crashing.

This is a small, targeted change that increases the stability of the plugin without altering its core logic.


vbuberen:
Could you list at least one case when it will be null? 

Reading the description I clearly see some AI generated text. Looks like the change is the same vibe coded thing. 
Since no real motivation and example on how to crash the app provided I am just closing it and would request to avoid opening some AI generated PR in future.

Issue #3649, opened 2025-08-21
[Request]: Android 16Kb page size support


Gagan5278:
### Plugin

share_plus 11.1.0

### Use case

Does package support Android 16Kb page size. We have a dead line Nov 1 for 16 Kb page size.

### Proposal

Android has a deadline for 16Kb page size. do we have already in Share_plus package

https://developer.android.com/guide/practices/page-sizes

vbuberen:
Not yet, but there will be a release for all packages in the upcoming week that includes updates for all the required dependencies.

Issue #3624, opened 2025-07-17
[Request]: Bump compile SDK of all plugins to 36 (Android 16)


vbuberen:
### Plugin

All plugins

### Use case

The whole set of Plus Plugins still declares compileSDK 34 (Android 14). It is behind current latest publicly available version SDK 36 (Android 16).

### Proposal

Update all packages to declare compileSDK 36

vbuberen:
Instead of declaring compileSDK as 36 all packages are switched to use `flutter.compileSdkVersion` variable for more flexibility.

PR #3665, opened 2025-09-07
feat(android_alarm_manager_plus)!: Change Android compile SDK, update Android build config

vbuberen:
## Description

The initial idea was to just do a bump to compile SDK 36, but decided to not do it and use `flutter.compileSdkVersion`. It should make us worry less about skipping Android version, like it happened with SDK 35.
At the same time I didn't want us to be strict with minSDK, so left it 21 as Flutter supported before 3.35.x release.

As to rest of changes in dependencies - it is an alignment with dependencies versions used by official Flutter packages (e.g. https://github.com/flutter/packages/blob/main/packages/shared_preferences/shared_preferences/example/android/settings.gradle ). I believe that such change should help with better compatibility across the pub ecosystem.

Additionally, thanks for Android Gradle Plugin bump we are preparing plugins to be compatible with apps moving to Android 16KB memory pages: https://developer.android.com/guide/practices/page-sizes

Marking this PR as breaking, because I am sure there will be lots of users, who will struggle with checking the requirements and adjusting Android part of their projects accordingly. 

Next up I will create similar PRs for the rest of plugins in the repo.

## Related Issues

- Part of #3624 
- Part of #3649 
- Supersedes #3617 

## Breaking Change

### DIFF
--- a/packages/android_alarm_manager_plus/README.md
+++ b/packages/android_alarm_manager_plus/README.md
@@ -20,8 +20,9 @@ Dart code in the background when alarms fire.
 - Flutter >=3.12.0
 - Dart >=3.1.0 <4.0.0
 - Java 17
+- Kotlin 2.2.0
 - Android Gradle Plugin >=8.12.1
-- Gradle wrapper >=8.11.1
+- Gradle wrapper >=8.13
 
 ## Getting Started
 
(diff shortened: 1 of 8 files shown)


ekuleshov:
@vbuberen stating this is at very least very hostile to all users of the plus plugins. 

> ...I am sure there will be lots of users, who will struggle with checking the requirements and adjusting Android part of their projects accordingly.

The best practice for a major version bump like that is to provide some migration guidelines in a plugin's readme

PR #3669, opened 2025-09-07
feat(network_info_plus)!: Change Android compile SDK, update Android build config

vbuberen:
## Description

Same as #3665 but for `network_info_plus`

## Related Issues

- Part of #3624
- Part of #3649 

## Breaking Change


vbuberen:
The failing test is due to running latest Flutter which has minSDK 24 for Android now.

PR #3721, opened 2025-12-01
docs(package_info_plus): add installerStore values documentation

### LINKED ISSUE #1468: Add store names to "installerStore" documentation
### Plugin

package_info_plus

### Use case

My goal is to open my apps page based on the store where the app was downloaded.

So if the user has downloaded the app on Google Play, I would like to open the Google Play developer page, same thing for iOS, Amazon, and other stores.

To do this, I need to know from which store the app was installed.

You added installerStore, but in my opinion it's not documented enough, because I don't know what names to compare the string with. I don't know if installerStore will return me for example "com.android.vending" or "Google Play Store" or something else.

The package I've used up to now (and which internally allowed you to understand exactly which store you were in) was this: https://pub.dev/packages/store_checker

### Proposal

I see two possibilities to improve the function:

- internally create a comparison system like the plugin I linked to you, in order to allow the developer to compare the store in a clear way

or

- update the documentation page https://pub.dev/documentation/package_info_plus/latest/package_info_plus/PackageInfo/installerStore.html and add the names of the stores that would return calling installerStore

nathannewyen:
Closes #1468

## Checklist

## Breaking Change

### DIFF
--- a/packages/package_info_plus/package_info_plus/README.md
+++ b/packages/package_info_plus/package_info_plus/README.md
@@ -48,15 +48,16 @@ String version = packageInfo.version;
 String buildNumber = packageInfo.buildNumber;
 ```
 
-## Installer Store
+### Installer Store
 
 The `installerStore` property indicates which app store installed the application. This is useful for directing users to the appropriate store page for ratings or updates.
 
 ```dart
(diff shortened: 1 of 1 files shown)


nathannewyen:
Can someone review and approve?

PR #3769, opened 2026-03-03
feat(share_plus): Avoid exceptions on iPads with no sharePositionOrigin

LuisGabrielZamora:
## Description
- Convert sharePositionOrigin from window to source view coordinates
- Use center fallback when no valid origin is provided on iPad
- iPhone behavior unchanged
- Closes: #3768 

## Changes
- **iPad – coordinate handling:** Convert `sharePositionOrigin` from window to source view coordinates using `convertRect:fromView:nil` before setting `sourceRect`
- **iPad – no origin fallback:** Use a center-of-view fallback rect instead of returning a `FlutterError` when no valid origin is provided
- **iPhone:** Unchanged behavior

## Backwards compatibility
- iPhone: No change
- iPad without `sharePositionOrigin`: Previously could crash; now works (center fallback)
- iPad with `sharePositionOrigin`: Popover now anchors correctly; no API changes

## Description

Fixes share sheet crash on iPad (iOS 17–26) when `sharePositionOrigin` is missing or in the wrong coordinate space.

## Related Issues

- [Related issue](https://github.com/fluttercommunity/plus_plugins/issues/3645)
- [Code already merged](https://github.com/fluttercommunity/plus_plugins/pull/3699)

## Checklist

- [X] I read the [Contributor Guide](https://github.com/fluttercommunity/plus_plugins/blob/main/CONTRIBUTING.md) and followed the process outlined there for submitting PRs.
- [X] I titled the PR using [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0).
- [X] I did not modify the `CHANGELOG.md` nor the plugin version in `pubspec.yaml` files.
- [X] All existing and new tests are passing.
- [X] The analyzer (`flutter analyze`) does not report any problems on my PR.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate that with a `!` in the title as explained in [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0)).
- [X] No, this is *not* a breaking change.

### Changes
- **iPad – coordinate handling:** Convert `sharePositionOrigin` from window to source view coordinates using `convertRect:fromView:nil` before setting `sourceRect`
- **iPad – no origin fallback:** Use a center-of-view fallback rect instead of returning a `FlutterError` when no valid origin is provided
- **iPhone:** Unchanged behavior

### Backwards compatibility
- iPhone: No change
[...]

Issue #3768, opened 2026-03-03
[Bug]: Avoid crash on iOS 26 on iPhones with no sharePositionOrigin


LuisGabrielZamora:
### Platform

Ipad - IOS From 17 to 26

### Plugin

share_plus

### Version

12.0.1

### Flutter SDK

3.38.3

### Description

Fixes share sheet on iPad (iOS 17–26) when sharePositionOrigin was missing or in the wrong coordinate space.

### Context

This problem is related with this the following previous bugs, they tried to fix the problem but this is still occurring:

- [PR](https://github.com/fluttercommunity/plus_plugins/pull/3699)
- [Related issue](https://github.com/fluttercommunity/plus_plugins/issues/3645)

### Why the previous code didn’t work (iPad)

**1. Wrong coordinate space**

Flutter sends sharePositionOrigin in global (window) coordinates (from localToGlobal()). The old code used that rect directly as sourceRect. On iPad, UIPopoverPresentationController.sourceRect must be in the source view’s coordinate space. Using window coordinates as-is broke the containment check and could misplace the popover or trigger validation failures.

**2. Over-strict validation**

The code required CGRectContainsRect(controller.view.frame, origin) and a non-empty origin. Because origin is in window coordinates and controller.view.frame is in another space, the check often failed even with a valid rect. When the app didn’t pass sharePositionOrigin (e.g. CGRectZero), it returned a FlutterError and blocked sharing instead of using a fallback.

**3. Result**

On iPad (iOS 17–26), share could fail with a platform error or the popover could anchor in the wrong place. iPhone was unaffected because it doesn’t rely on the same popover path.

### Changes

iPad – coordinate handling: Convert sharePositionOrigin from window to the source view with [controller.view convertRect:origin fromView:nil] before setting sourceRect, so the popover anchors correctly.
iPad – no origin: If no valid origin is provided, use a fallback rect at the center of the controller’s view instead of returning a FlutterError, so sharing always works on iPad. iPhone: Unchanged; sourceRect is still set only when a non-empty origin is provided.

### Backwards compatibility

- iPhone: Unchanged.
- iPad without sharePositionOrigin: Previously could fail; now works (center fallback).
- iPad with sharePositionOrigin: Popover now anchors correctly; no API changes.

### Code Sample

```dart
// ignore_for_file: unused_element, deprecated_member_use
// =============================================================================
// STANDALONE QA FILE — NO ABCMOUSE DEPENDENCIES
// =============================================================================
[...]

Issue #3645, opened 2025-08-12
[Bug]: iOS 26 Unhandled Exception: PlatformException(error, sharePositionOrigin: argument must be set, {{0, 0}, {0, 0}} must be non-zero and within coordinate space of source view: {{0, 0}, {414, 896}}, null, null)


yahyatabba:
### Platform

iPhone 11 - iOS 26 beta

### Plugin

share_plus

### Version

11.1.0

### Flutter SDK

3.33.0-1.0.pre-1359

### Steps to reproduce

final params = ShareParams(uri: Uri.parse("https://www.google.com"));
SharePlus.instance.share(params);

it returns:
[ERROR:flutter/runtime/dart_vm_initializer.cc(40)] Unhandled Exception: PlatformException(error, sharePositionOrigin: argument must be set, {{0, 0}, {0, 0}} must be non-zero and within coordinate space of source view: {{0, 0}, {414, 896}}, null, null)
#0      StandardMethodCodec.decodeEnvelope (package:flutter/src/services/message_codecs.dart:653:7)
#1      MethodChannel._invokeMethod (package:flutter/src/services/platform_channel.dart:367:18)
<asynchronous suspension>
#2      MethodChannelShare.share (package:share_plus_platform_interface/method_channel/method_channel_share.dart:25:20)
<asynchronous suspension>

### Code Sample

```dart
final params = ShareParams(uri: Uri.parse("https://www.google.com/"));
SharePlus.instance.share(params);
```

### Logs

```shell
[ERROR:flutter/runtime/dart_vm_initializer.cc(40)] Unhandled Exception: PlatformException(error, sharePositionOrigin: argument must be set, {{0, 0}, {0, 0}} must be non-zero and within coordinate space of source view: {{0, 0}, {414, 896}}, null, null)
#0      StandardMethodCodec.decodeEnvelope (package:flutter/src/services/message_codecs.dart:653:7)
#1      MethodChannel._invokeMethod (package:flutter/src/services/platform_channel.dart:367:18)
<asynchronous suspension>
#2      MethodChannelShare.share (package:share_plus_platform_interface/method_channel/method_channel_share.dart:25:20)
<asynchronous suspension>
```

### Flutter Doctor

```shell
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel master, 3.33.0-1.0.pre-1359, on macOS 26.0 25A5327h darwin-arm64, locale en-SY)
[...]

PR #3813, opened 2026-04-28
fix(share_plus): avoid iOS 26 share sheet crash for non-image files

MohamedAbdallah-14:
Fixes #3784.

## What's wrong

Pre-iOS 26 the plugin used a fake file URL trick to display the file size and extension as a subtitle in the iOS share sheet:

```objc
// description = "DOCX • 500 KB"
metadata.originalURL = [NSURL fileURLWithPath:description];
```

This worked because pre-iOS 26 versions of `LinkPresentation` accepted any `NSURL` and only used it for display lookup.

iOS 26 tightened validation of `LPLinkMetadata.originalURL`. When `SHSheetActivityItemsManager` asynchronously fetches link metadata for the bogus URL — reproducible when sharing `.docx` files — `LPLinkPreview` returns a `nil` or incomplete metadata object, and `_updateMetadataItemForActivityItem:withFetchedMetadata:` crashes inserting it into an `NSMutableArray`:

```
*** -[__NSArrayM insertObject:atIndex:]: object cannot be nil
SHSheetActivityItemsManager -[SHSheetActivityItemsManager _updateMetadataItemForActivityItem:withFetchedMetadata:]
SHSheetActivityItemsManager -[SHSheetActivityItemsManager _fetchMetadataForActivityItem:withHandler:]
```

## Fix

Gate the fake-URL trick on iOS < 26 with `@available(iOS 26.0, *)`. On iOS 26 and later, leave `originalURL` unset (defaults to `nil`). The share sheet still works — the title and image preview both still render — but the size/extension subtitle that the fake URL was producing is no longer shown on iOS 26+.

That tradeoff matches what the reporter on #3784 already suggested:

> Setting `metadata.originalURL = nil` instead of the fake path may help.

PDFs and plain text are unaffected, and pre-iOS 26 behaviour is unchanged.

## Test plan

- [ ] Manually verify on an iOS 26.x device or simulator: share a `.docx` file, confirm no crash.
- [ ] Manually verify on an iOS 17 device or simulator: share a `.docx` file, confirm subtitle still shows (\"DOCX • 500 KB\").
- [ ] Manually verify on iOS 17 and iOS 26: image sharing still produces a thumbnail preview (`metadata.imageProvider` path is unaffected).

I do not currently have an iOS 26 simulator to run the repro myself; opening as a draft so a maintainer or CI can validate the iOS 26 behaviour before this is marked ready.

## Risks

- iOS 26+ users lose the file size subtitle in the share sheet. The title and image preview are unaffected. This matches the reporter's expectation in the issue.
- If a future iOS version reintroduces support for arbitrary `originalURL` values, this gate could be removed; for now `iOS 26.0` is the safe lower bound based on the reproduction in #3784.

## Related

- #3768 — separate iOS 26 crash with `sharePositionOrigin` (PR #3769).
- #3645 — `sharePositionOrigin must be set` crash.

### DIFF
--- a/packages/share_plus/share_plus/ios/share_plus/Sources/share_plus/FPPSharePlusPlugin.m
+++ b/packages/share_plus/share_plus/ios/share_plus/Sources/share_plus/FPPSharePlusPlugin.m
@@ -247,7 +247,19 @@ activityTypesForStrings(NSArray<NSString *> *activityTypeStrings) {
     }
 
     // https://stackoverflow.com/questions/60563773/ios-13-share-sheet-changing-subtitle-item-description
-    metadata.originalURL = [NSURL fileURLWithPath:description];
+    // The fake file URL above is a long-standing trick used to surface the
+    // file size/extension as a subtitle in the share sheet. iOS 26 introduced
+    // stricter validation of LPLinkMetadata.originalURL: when the share sheet
+    // tries to fetch link metadata (SHSheetActivityItemsManager) for the
+    // bogus URL — for example when sharing .docx files — it crashes with
+    // "*** -[__NSArrayM insertObject:atIndex:]: object cannot be nil".
+    // Skip the fake URL on iOS 26+ and accept the loss of the subtitle in
+    // exchange for not crashing. See issue #3784.
+    if (@available(iOS 26.0, *)) {
+      // Leave metadata.originalURL nil.
+    } else {
+      metadata.originalURL = [NSURL fileURLWithPath:description];
+    }
     if (_mimeType && [_mimeType hasPrefix:@"image/"]) {
       UIImage *image = [UIImage imageWithContentsOfFile:_path];
       metadata.imageProvider = [[NSItemProvider alloc]


MohamedAbdallah-14:
The Android API 36 job timed out waiting for the emulator to boot (Timeout waiting for emulator to boot at the getprop sys.boot_completed polling step) — it never reached the test phase. Fail-fast then cancelled API 32, API 29, and the iOS job before they could run.

This PR is iOS-only (FPPSharePlusPlugin.m); no Android code is touched. Could you re-run the failed jobs when you get a chance? The iOS integration test in particular needs to actually execute to validate the fix.

MohamedAbdallah-14:
Friendly bump. The only red CI is the API 36 Android emulator boot timeout (`getprop sys.boot_completed` never returned); fail-fast then cancelled API 29/32 and the iOS job. This PR only touches `FPPSharePlusPlugin.m` (iOS), so the Android failure isn't from the diff. A re-run of the cancelled jobs would clear it. Happy to address any review feedback.

vbuberen:
Thanks for your contribution, but I couldn't reproduce the issue and docx files share properly on iOS 26 devices in the latest share_plus version. That bug report references function which is deprecated since version 11.0.0.

PR #3819, opened 2026-05-01
feat(device_info_plus)!: Migrate Android plugin to built-in Kotlin

mohamed-foly:
## Description
Move the Android plugin to Flutter's built-in Kotlin setup and remove the old `kotlin-android` wiring. This keeps the plugin aligned with the supported Android build path before the old KGP flow becomes a build problem.

## Related Issues
None.

## Checklist
- [x] I titled the PR using Conventional Commits.
- [x] I ran the available validation for this change: `flutter pub get` and the example build on the Flutter 3.44 line.
- [x] I checked the app side with `flutter analyze`; it only reports the pre-existing infos.

## Breaking Change
Yes, this raises the minimum Flutter/Dart floor for the package release.

### DIFF
--- a/packages/device_info_plus/device_info_plus/android/build.gradle
+++ b/packages/device_info_plus/device_info_plus/android/build.gradle
@@ -22,7 +22,6 @@ rootProject.allprojects {
 }
 
 apply plugin: 'com.android.library'
-apply plugin: 'kotlin-android'
 
 android {
     namespace 'dev.fluttercommunity.plus.device_info'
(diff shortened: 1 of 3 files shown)


vbuberen:
I would also ask to read the contributors guide, because it would be better to start with an issue open in the repo.

> environment:
  sdk: ">=3.12.0-0 <4.0.0"
  flutter: ">=3.44.0-0.0.pre"

There will be no such sharp bump of requirements. I don't plan to do such a bump for now unless there is a real need, like latest Flutter stable (not `pre` or alpha or whatever) version requires to do so because of some change. 

For now this PR and the rest make no sense and even if they in a draft. I will close them as there is no plan to accept them now.

PR #3832, opened 2026-05-21
fix(device_info_plus): guard iOS vision selector

andyzhshg:
## Description

This PR fixes an iOS crash in `device_info_plus` when collecting iOS device info on some iOS 26.1 runtimes.

Crash signature:

```text
NSInvalidArgumentException: -[_NSSwiftProcessInfo isiOSAppOnVision]: unrecognized selector sent to instance
  device_info_plus -[FPPDeviceInfoPlusPlugin handleMethodCall:result:] (FPPDeviceInfoPlusPlugin.m:35)
```

The existing implementation guarded the call with `@available(iOS 26.1, *)`, but availability only confirms that the SDK/runtime version is new enough. It does not guarantee that the concrete `NSProcessInfo` instance responds to the Objective-C selector. In the crash above the object is `_NSSwiftProcessInfo`, and sending `isiOSAppOnVision` directly can raise an unrecognized selector exception.

This change keeps the default value as `false` and only invokes `isiOSAppOnVision` after verifying that the `NSProcessInfo` instance responds to the selector. It also adds a focused regression test for the native source guard so this crash path does not get reintroduced accidentally.

Verification run for the affected package:

```text
flutter test
flutter analyze
dart analyze . --fatal-infos
git diff --check
```

## Related Issues

No existing upstream issue was found for this specific `isiOSAppOnVision` crash signature.

## Checklist

## Breaking Change

### DIFF
--- a/packages/device_info_plus/device_info_plus/ios/device_info_plus/Package.swift
+++ b/packages/device_info_plus/device_info_plus/ios/device_info_plus/Package.swift
@@ -11,11 +11,15 @@ let package = Package(
     products: [
         .library(name: "device-info-plus", targets: ["device_info_plus"])
     ],
-    dependencies: [],
+    dependencies: [
+        .package(name: "FlutterFramework", path: "../FlutterFramework")
+    ],
     targets: [
         .target(
             name: "device_info_plus",
-            dependencies: [],
+            dependencies: [
+                .product(name: "FlutterFramework", package: "FlutterFramework")
+            ],
             resources: [
                 .process("PrivacyInfo.xcprivacy"),
             ],
--- a/packages/device_info_plus/device_info_plus/ios/device_info_plus/Sources/device_info_plus/FPPDeviceInfoPlusPlugin.m
+++ b/packages/device_info_plus/device_info_plus/ios/device_info_plus/Sources/device_info_plus/FPPDeviceInfoPlusPlugin.m
@@ -32,7 +32,9 @@
     }
     NSNumber *isiOSAppOnVision = [NSNumber numberWithBool:NO];
     if (@available(iOS 26.1, *)) {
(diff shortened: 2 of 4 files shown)


vbuberen:
Are there any actual example of such crash? 
Also I see that the PR has changes non-related to the title.

vbuberen:
Closing due to not feedback and no reproducer.

ShirishAryal:
Hello @vbuberen  I'm hitting this same crash on an iPhone 13 running iOS 26. It's not consistently reproducible on other devices with identical specs, so it seems tied to a specific runtime condition rather than the device model itself.

Crash log:
```
* First throw call stack:
(0x190197988 0x18d0e17c4 0x1902308f4 0x19011735c 0x19011f200 0x1005050d4 0x1064deb7c 0x106031920 0x1c8d58adc 0x1c8d727ec 0x1c8d8fb24 0x1c8d67ec8 0x1c8d67e04 0x19013c2c8 0x1900efb3c 0x1900eea6c 0x23081b498 0x195a8917c 0x195a2dc2c 0x195b5e140 0x101bc736c 0x101bc72dc 0x101bc743c 0x18d136e28)
* Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[_NSSwiftProcessInfo isiOSAppOnVision]: unrecognized selector sent to instance 0x103d70480'
libc++abi: terminating due to uncaught exception of type NSException
Process 15394 resuming
Process 15394 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
frame #0: 0x00000002393a10cc libsystem_kernel.dylib`__pthread_kill + 8
```
This matches the crash signature described in this PR. Given that it's reproducible (just not on every device), could we reopen this and take another look? Happy to share the exact iOS build number and app scenario that triggers it if that's useful for narrowing down the root cause.

PR #3435, opened 2025-01-18
feat(share_plus): Add the intent options for the Android platform

ozyl:
## Description

- `android_intent_plus` does not support sharing pictures (Maybe in the future)
- `share_plus`does not support sharing to target app on android (This PR)

This PR enables share_plus to support sharing of specified applications on Android. 

### Platform-Specific Features

Use `PlatformOptions` to configure platform features.

#### Android

Supports fine-grained configuration of intent in Android. You can use `AndroidIntentOptions` to specify the app to share.

```dart
  Share.share('check out my website https://example.com', platformOptions:PlatformOptions(
    androidIntentOptions: AndroidIntentOptions(
    	packageName: 'com.example.app',
    	componentName: "com.example.app.ShareActivity",
    	flags: [
    	  AndroidIntentFlag.FLAG_ACTIVITY_CLEAR_TOP,
    	  AndroidIntentFlag.FLAG_ACTIVITY_NEW_TASK,
    	]
  ));
```

## Related Issues

- *Request #3269*
- *Related #2539*

## Checklist

- [x] I read the [Contributor Guide](https://github.com/fluttercommunity/plus_plugins/blob/main/CONTRIBUTING.md) and followed the process outlined there for submitting PRs.
- [x] I titled the PR using [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0).
- [x] I did not modify the `CHANGELOG.md` nor the plugin version in `pubspec.yaml` files.
- [x] All existing and new tests are passing.
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate that with a `!` in the title as explained in [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0)).
- [x] No, this is *not* a breaking change.

Issue #2682, opened 2024-03-13
[Bug]: can't back to my app after share via QQ Email


komhkomh:
### Platform

Android 13

### Plugin

share_plus

### Version

7.2.2

### Flutter SDK

3.10.5

### Steps to reproduce

Provided that the QQ Email is installed and the phone is Android system(I am using Galaxy S22, and it also doesn't work on Galaxy Z Fold4 ).

1. Press the 'Add image' button to add local image;
2. Press the 'Share' button and you can see QQ Mail, then selected the text of  'Send via Email';
3. Use system back button to go back to the app. It can't achieved. 

### Code Sample

```dart
import 'dart:io';

import 'package:file_selector/file_selector.dart';
import 'package:flutter/foundation.dart';
import 'package:flutter/material.dart';
import 'package:image_picker/image_picker.dart';
import 'package:share_plus/share_plus.dart';

void main() {
  runApp(const MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({super.key});

  // This widget is the root of your application.
  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      title: 'Flutter Demo',
      theme: ThemeData(
        colorScheme: ColorScheme.fromSeed(seedColor: Colors.deepPurple),
        useMaterial3: true,
      ),
      home: const MyHomePage(title: 'Flutter Demo Home Page'),
    );
  }
}

class MyHomePage extends StatefulWidget {
  const MyHomePage({super.key, required this.title});

  final String title;

  @override
  State<MyHomePage> createState() => _MyHomePageState();
}

class _MyHomePageState extends State<MyHomePage> {
  String text = '';
  String subject = '';
  String uri = '';
  List<String> imageNames = [];
  List<String> imagePaths = [];

  @override
  Widget build(BuildContext context) {
    return Scaffold(
        body: SingleChildScrollView(
      padding: const EdgeInsets.all(24),
      child: Column(
        crossAxisAlignment: CrossAxisAlignment.start,
        mainAxisSize: MainAxisSize.min,
        children: <Widget>[
          const SizedBox(height: 16),
          ImagePreviews(imagePaths, onDelete: _onDeleteImage),
          ElevatedButton.icon(
            label: const Text('Add image'),
            onPressed: () async {
              // Using `package:image_picker` to get image from gallery.
              if (!kIsWeb &&
                  (Platform.isMacOS ||
                      Platform.isLinux ||
                      Platform.isWindows)) {
[...]

Issue #2267, opened 2023-10-15
[Request]: Specify Intent.FLAG_ACTIVITY_NEW_TASK for sharing


micahmo:
### Plugin

share_plus

### Use case

I am experiencing a behavior where sharing via the Android share sheet causes the target app to be displayed it the same app switcher entry as the originating app. From my research, this is because the sharing intent is not created with the `Intent.FLAG_ACTIVITY_NEW_TASK` flag. If we could optionally (or always) pass this flag when sharing, I believe that would make the flow much more intuitive for users.

### Proposal

Have an option to (or just always) set `Intent.FLAG_ACTIVITY_NEW_TASK` when creating the share intent on Android.

### Expected behavior

This is sharing from Keep to Drive, and it works as expected.

https://github.com/fluttercommunity/plus_plugins/assets/7417301/3fc7a343-7d9e-4de5-a52b-863ec66a7a07

### Current behavior

This is sharing from Thunder (which uses `share_plus`) to Drive, and it does not work as expected.

https://github.com/fluttercommunity/plus_plugins/assets/7417301/0ec0877f-ba8b-47ac-a2b9-b351f1ccb1ec

vbuberen:
Thanks for suggesting. Indeed looks useful, even though I more often see apps doing sharing in the same task (for example, Twitter). But I like the idea or providing developers with choice.

CaviarChen:
We also found a case that this can be useful. Also I believe this might be related to: https://github.com/fluttercommunity/plus_plugins/issues/2682

PR #3838, opened 2026-05-26
feat(share_plus): launch share target in its own task on Android

leonardo-oliveir-a:
## Summary

Adds `Intent.FLAG_ACTIVITY_NEW_TASK` to the chooser intent on Android so the destination app (WhatsApp, Telegram, Drive, etc.) launches in its own task instead of being stacked on top of the calling app's task.

## Motivation

Resolves the long-standing UX inconsistency described in:

- **#2267** ([Request]: Specify Intent.FLAG_ACTIVITY_NEW_TASK for sharing) — open since Oct 2023
- **#2682** (can't back to my app after share via QQ Email) — closed, but root cause was the same; maintainer @miquelbeltran [acknowledged](https://github.com/fluttercommunity/plus_plugins/issues/2682#issuecomment-2012213175) the fix mechanism in the comments

Without the flag, the recents switcher shows a single entry for the calling app with the target app visible inside it — inconsistent with how native Android apps (Google Keep, Instagram, etc.) handle share, where target and caller appear as separate recents entries.

Demonstration from #2267 (linked videos): Keep → Drive works correctly; Thunder (using share_plus) → Drive does not.

## Previous attempts

- **#3435** (Jan 2025, `AndroidIntentOptions` with configurable flags) — closed in favor of the refactor in #3404, but the refactor merged without absorbing this feature
- **#3751** (Feb 2026, equivalent one-line fix) — closed 19 seconds after creation by the conventional-commits CI bot rejecting the title; never reopened

This PR uses the same minimal approach as #3751 with a conventional-commits-compliant title.

## Change

One file, ~10 lines (including comment):

```kotlin
// packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/Share.kt
chooserIntent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
```

Applied right after `Intent.createChooser(...)` and before the file-permission grant + `startActivity`.

## Trade-off considered

@miquelbeltran's [comment on #2267](https://github.com/fluttercommunity/plus_plugins/issues/2267#issuecomment-1768063127) noted that some apps (e.g. Twitter) do share within the same task and asked for "choice". This PR opts for always-on rather than introducing a new API surface, matching what #3751 attempted. Rationale:

- The same-task behavior is generally considered a UX regression vs. native Android share — most apps that need same-task behavior do so unintentionally.
- An always-on flag has zero API surface change → no migration burden, no deprecation noise.
- If the team prefers opt-in, happy to convert this to a `ShareParams.newTask` (default true) — let me know and I'll push.

## Test plan

- [x] Manual test on Android 11 (OnePlus 6T): share text + image to Telegram and WhatsApp → target apps now appear as separate recents entries (was: single consolidated entry).
- [x] Manual test: back button in target app still returns to caller app correctly.
- [x] Manual test: deep links (`https://...`) inside shared text are preserved and handled by target.

## Checklist

## Breaking Change

No API surface change. Behavioral change only: target apps will now appear as separate recents entries (the same way native Android share behaves). Apps that explicitly relied on the previous same-task stacking would need to wrap share calls in a task-affinity workaround — but I'm not aware of any documented case.

### DIFF
--- a/packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/Share.kt
+++ b/packages/share_plus/share_plus/android/src/main/kotlin/dev/fluttercommunity/plus/share/Share.kt
@@ -131,6 +131,16 @@ internal class Share(
                 Intent.createChooser(shareIntent, title)
             }
 
+        // Launch the destination app in its own task so it appears as a
+        // separate entry in the recents switcher instead of being stacked
+        // on top of the calling app's task. Without this flag the target
+        // (WhatsApp, Telegram, Drive, etc.) inherits the caller's task —
+        // user sees a single recents entry for the caller with the target
+        // visible inside, which is confusing and inconsistent with how
+        // Android share works for most native apps (e.g. Google Keep,
+        // Instagram). See plus_plugins issue #2267.
+        chooserIntent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
+
         // Grant permissions to all apps that can handle the files shared
         if (fileUris != null) {
             val resInfoList = getContext().packageManager.queryIntentActivities(


FaFre:
Was just about to implement this as well, why did you close this pr @leonardo-oliveir-a ?

leonardo-oliveir-a:
> Was just about to implement this as well, why did you close this pr @leonardo-oliveir-a ?

Sorry for that, it was closed by mistake.

vbuberen:
Thanks for the PR, but closing this one because the always-on approach can't work: since the plugin refactor, every foreground share goes through startActivityForResult (that's how ShareResult is delivered), and per the Activity.startActivityForResult docs, launching with FLAG_ACTIVITY_NEW_TASK makes the caller receive an immediate RESULT_CANCELED. So this change would make share() resolve as dismissed while the sheet is still open, for every user, silently.

PR #3842, opened 2026-05-27
feat(connectivity_plus): add hasConnectivity() utility extension

### LINKED ISSUE #3841: [Request]: Add hasConnectivity getter extension to List<ConnectivityResult>
### Plugin

connectivity_plus

### Use case

A convenient getter that makes the API design more self-explanatory and slightly easier to read:

```dart
final result = await checkConnectivity();
print(result.hasConnectivity);
```

In many cases, only the general connectivity state is required (connected vs disconnected), independent of the underlying transport (e.g., Wifi, Ethernet).

A very minor feature that is commonly used in published packages (e.g., [serverpod_flutter](https://github.com/serverpod/serverpod/blob/1aa3e67130f44d5b8f7ec887001af73e262c523f/packages/serverpod_flutter/lib/src/flutter_connectivity_monitor.dart#L22-L23)).

### Proposal

```dart
// connectivity_plus_platform_interface/lib/src/enums.dart

extension ConnectivityResultListX on List<ConnectivityResult> {
  bool get hasConnectivity =>
      !(length == 1 && first == ConnectivityResult.none);
}
```

This implementation respects the current doc comment of `connectivity_plus`:

```dart
/// The returned list is never empty. In case of no connectivity, the list contains
/// a single element of [ConnectivityResult.none]. Note also that this is the only
/// case where [ConnectivityResult.none] is present.
```

EchoEllet:
## Description

A very minor feature that is commonly used (e.g., [serverpod_flutter](https://github.com/serverpod/serverpod/blob/1aa3e67130f44d5b8f7ec887001af73e262c523f/packages/serverpod_flutter/lib/src/flutter_connectivity_monitor.dart#L22-L23)):

```dart
final result = await checkConnectivity();
print(result.hasConnectivity);
```

## Related Issues

- Fix #3841

## Checklist

## Breaking Change

### DIFF
--- a/packages/connectivity_plus/connectivity_plus/lib/connectivity_plus.dart
+++ b/packages/connectivity_plus/connectivity_plus/lib/connectivity_plus.dart
@@ -41,19 +41,9 @@ class Connectivity {
   /// status changes, this is a known issue that only affects simulators.
   /// For details see https://github.com/fluttercommunity/plus_plugins/issues/479.
   ///
-  /// The emitted list is never empty.
-  /// {@template no_connectivity_case}
-  /// In case of no connectivity, the list contains
+  /// The emitted list is never empty. In case of no connectivity, the list contains
   /// a single element of [ConnectivityResult.none]. Note also that this is the only
   /// case where [ConnectivityResult.none] is present.
-  ///
-  /// Use `hasConnectivity` to determine whether any active network connection exists:
-  ///
-  /// ```dart
-  /// final result = await checkConnectivity();
-  /// print(result.hasConnectivity);
-  /// ```
-  /// {@endtemplate}
   ///
   /// This method applies [Stream.distinct] over the received events to ensure
   /// only emitting when connectivity changes.
@@ -67,8 +57,9 @@ class Connectivity {
   /// make a network request, it only gives you the radio status. Instead, listen
   /// for connectivity changes via [onConnectivityChanged] stream.
   ///
-  /// The returned list is never empty.
-  /// {@macro no_connectivity_case}
+  /// The returned list is never empty. In case of no connectivity, the list contains
+  /// a single element of [ConnectivityResult.none]. Note also that this is the only
+  /// case where [ConnectivityResult.none] is present.
   Future<List<ConnectivityResult>> checkConnectivity() {
     return _platform.checkConnectivity();
   }
--- a/packages/connectivity_plus/connectivity_plus/lib/connectivity_plus.dart
+++ b/packages/connectivity_plus/connectivity_plus/lib/connectivity_plus.dart
@@ -9,7 +9,7 @@ import 'package:collection/collection.dart';
 
 // Export enums from the platform_interface so plugin users can use them directly.
 export 'package:connectivity_plus_platform_interface/connectivity_plus_platform_interface.dart'
-    show ConnectivityResult;
+    show ConnectivityResult, ConnectivityResultListX;
 
 export 'src/connectivity_plus_linux.dart'
     if (dart.library.js_interop) 'src/connectivity_plus_web.dart';
--- a/packages/connectivity_plus/connectivity_plus/lib/connectivity_plus.dart
+++ b/packages/connectivity_plus/connectivity_plus/lib/connectivity_plus.dart
@@ -9,10 +9,11 @@ import 'package:collection/collection.dart';
 
 // Export enums from the platform_interface so plugin users can use them directly.
 export 'package:connectivity_plus_platform_interface/connectivity_plus_platform_interface.dart'
-    show ConnectivityResult, ConnectivityResultListX;
+    show ConnectivityResult;
 
 export 'src/connectivity_plus_linux.dart'
     if (dart.library.js_interop) 'src/connectivity_plus_web.dart';
+export 'src/enum_extension.dart';
 
 /// Discover network connectivity configurations: Distinguish between WI-FI and cellular, check WI-FI status and more.
 class Connectivity {
--- a/packages/connectivity_plus/connectivity_plus/lib/connectivity_plus.dart
+++ b/packages/connectivity_plus/connectivity_plus/lib/connectivity_plus.dart
@@ -41,9 +41,19 @@ class Connectivity {
   /// status changes, this is a known issue that only affects simulators.
   /// For details see https://github.com/fluttercommunity/plus_plugins/issues/479.
   ///
-  /// The emitted list is never empty. In case of no connectivity, the list contains
+  /// The emitted list is never empty.
+  /// {@template no_connectivity_case}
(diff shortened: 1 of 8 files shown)


EchoEllet:
This PR does not modify `sensors_plus`, so this [CI failure](https://github.com/fluttercommunity/plus_plugins/actions/runs/26543635416/job/78190616476?pr=3842) is most likely irrelevant.

packages/connectivity_plus/connectivity_plus/lib/connectivity_plus.dart
```diff
@@ -41,12 +41,22 @@ class Connectivity {
...
+  /// {@template no_connectivity_case}
+  /// In case of no connectivity, the list contains
   /// a single element of [ConnectivityResult.none]. Note also that this is the only
   /// case where [ConnectivityResult.none] is present.
   ///
+  /// Use `hasConnectivity` to determine whether any active network connection exists:
+  ///
+  /// ```dart
+  /// final result = await checkConnectivity();
+  /// print(result.hasConnectivity);
```
vbuberen: It won't work the way it is described as currently calls like `checkConnectivity()` are not done without `Connectivity()` class. And `hasConnectivity` is only available if there is an import of 'package:connectivity_plus_platform_interface/connectivity_plus_platform_interface.dart' in the file. 
So this doc comment is misleading and needs an improvement. I would suggest to move this example comment about `hasConnectivity` to be above the extension itself.

packages/connectivity_plus/connectivity_plus/lib/connectivity_plus.dart
vbuberen, line 12: Let's add the new extension here to avoid the need for additional import of platform interface by package users.

packages/connectivity_plus/connectivity_plus/lib/connectivity_plus.dart
```diff
@@ -41,12 +41,22 @@ class Connectivity {
   /// status changes, this is a known issue that only affects simulators.
   /// For details see https://github.com/fluttercommunity/plus_plugins/issues/479.
   ///
-  /// The emitted list is never empty. In case of no connectivity, the list contains
+  /// The emitted list is never empty.
+  /// {@template no_connectivity_case}
```
vbuberen: Revert this change with templates as it makes sense only for generated API docs, while for people actually reading the source code only complicates things.

vbuberen (commented):
I like the idea, but there is some work to do before we can merge this. 

My main suggestion is to export extension via `show` as I suggested in one of my review comments.

Additionally, let's add mention and code example with the new extension to the README file and the example app, so it is easy to find for package users.

packages/connectivity_plus/connectivity_plus/lib/connectivity_plus.dart
EchoEllet: This will cause a temporary Dart analysis warning since `connectivity_plus` depends on `connectivity_plus_platform_interface` from pub.dev, and `ConnectivityResultListX` does not exist in the latest published version.

Should I add a temporary dependency override, or leave that to the project maintainers?

We also probably want to validate whether `ConnectivityResultListX` is a good name, because changing it later is a breaking change.

packages/connectivity_plus/connectivity_plus/lib/connectivity_plus.dart
EchoEllet: > I would suggest to move this example comment about `hasConnectivity` to be above the extension itself.

`ConnectivityResultListX` is inside `connectivity_plus_platform_interface`, so it should probably not assume the API shape of code that depends on `connectivity_plus`.

I would avoid the reference to `Connectivity()` entirely inside the platform interface package.

> So this doc comment is misleading and needs an improvement. 

With the proposed change, consumers cannot find this extension by only reading `connectivity_plus`.

I think we can add a doc comment to `hasConnectivity` instead:

```dart
extension ConnectivityResultListX on List<ConnectivityResult> {
  /// Returns whether any active network connection exists.
  bool get hasConnectivity =>
      !(length == 1 && first == ConnectivityResult.none);
}
```

And then document it in `README.md` of `connectivity_plus` instead?

packages/connectivity_plus/connectivity_plus/lib/connectivity_plus.dart
EchoEllet: Sure.

>  only for generated API docs, 

And IDEs as well.

> while for people actually reading the source code only complicates things.

Agreed. Repeating the same doc comment is a reasonable tradeoff here.

packages/connectivity_plus/connectivity_plus/lib/connectivity_plus.dart
vbuberen: > ConnectivityResultListX is inside connectivity_plus_platform_interface, so it should probably not assume the API shape of code that depends on connectivity_plus.

And still it would be better to have this comment there than mislead users with the comment that was added initially. I have high doubts that anybody would actually use platform_interface separately from plugin. 

I thought a little bit more about this extension - I believe it might be better to move it to the `connectivity_plus` package itself our of platform interface. It should be a better place for such thing.

packages/connectivity_plus/connectivity_plus/lib/connectivity_plus.dart
vbuberen: > Should I add a temporary dependency override, or leave that to the project maintainers?

If you use Melos and run `melos bootstrap` automatic dependency override will happen for local development purposes.

> We also probably want to validate whether ConnectivityResultListX is a good name, because changing it later is a breaking change.

My only request here is to not have a chance of clashing with some other package and with this name I don't think we will have such clashes. How good it is or how nice it is - don't really care.

packages/connectivity_plus/connectivity_plus/lib/connectivity_plus.dart
EchoEllet: > And still it would be better to have this comment there than mislead users with the comment that was added initially.

The previous example was intended to demonstrate the extension method on `List<ConnectivityResult>`, not how to construct `Connectivity` or use the plugin. 
That was already implied in the README and doc comments of `connectivity_plus`. I think it may mislead some AI models if it does not have enough context. However, the doc comment is attached directly to the `Connectivity.checkConnectivity()`.

The subtle error was that I missed exporting the extension in `connectivity_plus`.

>  I have high doubts that anybody would actually use platform_interface separately from plugin.

Agreed. This is typically used by consumers of the functionality, not by platform implementations (e.g., [connectivity_plus_linux_portal](https://pub.dev/packages/connectivity_plus_linux_portal)).
This should also allow us to document it there without assuming the API shape of a non-transitive dependency.

vbuberen (approved):
LGTM. Thanks for this contribution and updating everything quickly.

PR #3847, opened 2026-05-28
ci(device_info_plus): migrate android workflows to ubuntu-24.04 and kvm

mahendrimd:
## Description
This PR migrates the Android workflows for `device_info_plus` from `macos-15-intel` to `ubuntu-24.04` environment.

### The Problem
Currently, the repo utilize macOS runner which make android build takes more time and android integration tests running on API level 36 consistently time out and fail across the repository's packages. This PR serves as one part of a plan to migrate all nine packages to `ubuntu-24.04` to resolve this issue.

### Consideration and Reference
* This aligns with the [android-emulator-runner README.md](https://github.com/ReactiveCircus/android-emulator-runner), which references the official [GitHub Blog regarding hardware-accelerated Android virtualization on Linux](https://github.blog/changelog/2023-02-23-hardware-accelerated-android-virtualization-on-actions-windows-and-linux-larger-hosted-runners/):
  > "...Testing on a 4-core machine with hardware acceleration is around 2-3 times faster than not using hardware acceleration and around 2 times faster than using MacOS."
* The community has reported substantial performance improvements on Linux runners compared to macOS instances (see the [Kiwix Android migration PR](https://github.com/kiwix/kiwix-android/pull/3786) and [GitHub Runner Images Issue #1336](https://github.com/actions/runner-images/issues/1336#issuecomment-1310421187).

### Benchmark & Validation Data on `android_example_build`
The table below presents the total runtime of the workflow job for build. Control baseline were retrieved from the repo historical runtime logs, compared against two validation runs on the Linux environment on my other branch:

| No | Package Name | Control (`macos-15-intel`) | Test 1 (`ubuntu-24.04`) | Test 2 (`ubuntu-24.04`) |
| :--- | :--- | :---: | :---: | :---: |
| 1 | `android_alarm_manager_plus` | 10m 2s | 5m 31s | 5m 23s |
| 2 | `android_intent_plus` | 10m 57s | 4m 30s | 4m 19s |
| 3 | `battery_plus` | 13m 1s | 4m 46s | 4m 33s |
| 4 | `connectivity_plus` | 17m 21s | 4m 24s | 4m 24s |
| 5 | `device_info_plus` | 12m 1s | 4m 53s | 4m 40s |
| 6 | `network_info_plus` | 12m 2s | 4m 48s | 4m 55s |
| 7 | `package_info_plus` | 8m 14s | 4m 41s | 4m 41s |
| 8 | `sensors_plus` | 6m 6s | 4m 25s | 4m 41s |
[...]

PR #3849, opened 2026-05-28
ci(connectivity_plus): migrate android workflows to ubuntu-24.04 and kvm

mahendrimd:
## Description
This is part of migration across the repository's packages. For the full  description, please refer to the pull request #3847

### DIFF
--- a/.github/workflows/connectivity_plus.yaml
+++ b/.github/workflows/connectivity_plus.yaml
@@ -44,7 +44,7 @@ jobs:
           working-directory: ./packages/connectivity_plus
 
   android_example_build:
-    runs-on: macos-15-intel
+    runs-on: ubuntu-24.04
     timeout-minutes: 30
     steps:
       - name: "Checkout repository"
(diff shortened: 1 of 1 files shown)


mahendrimd:
The Android workflow migration is a success and running as expected. The unrelated macOS/iOS and `all_plugins` failures are explained in the master PR referenced above.

Note on API Level 24 Failure: the integration test failure on Android API 24 is a pre-existing issue and is unrelated to this runner migration.

Upon investigation, the failure comes from a legacy architectural behavior in older Android emulators. For API levels below 25, the emulator does not provision a simulated Wi-Fi access point (AndroidWifi) out of the box. This is confirmed by [archived Android Developer Documentation](https://web.archive.org/web/20180315020612/https://developer.android.com/studio/run/emulator.html) obtained from this [StackOverflow question](https://stackoverflow.com/questions/7876302/enabling-wifi-on-android-emulator):

>Wi-Fi
> When using an AVD with API level 25 or higher, the emulator provides a simulated Wi-Fi access point ("AndroidWifi"), and Android automatically connects to it.

Because the tests expect network state that does not exist on API 24 environments, this test fails. This can be patched in a separate PR.

vbuberen (approved):

PR #3889, opened 2026-06-21
feat(device_info_plus): Add support of built-in Kotlin

### LINKED ISSUE #3835: Migrate Plugin to Built-in Kotlin
### Plugin

device_info_plus

### Use case

Starting with Android Gradle Plugin (AGP) 9.0, support for applying the Kotlin Gradle Plugin (KGP) has been removed. Because this plugin applies KGP, it causes a compilation error that prevents my app from building. Here is an example of the error https://github.com/flutter/flutter/issues/181383.

Flutter has temporarily added support to allow KGP while apps and plugins migrate to AGP 9.0+, but this support will be removed in a future version of Flutter.

Please migrate this plugin to use built-in Kotlin to ensure your plugin users can successfully build their apps in future versions of Flutter.

Here is the Flutter [migration guide for plugin authors](https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-plugin-authors).

Please be respectful and mindful of the plugin repository's rules and code of conduct when reporting issues and interacting with plugin authors.

For reference, see the [Flutter Code of Conduct](https://github.com/flutter/flutter/blob/master/CODE_OF_CONDUCT.md).

### Proposal

Please migrate to Flutter's request.

vbuberen:
## Description

Backwards compatible support of built-in Kotlin as per https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-plugin-authors#supporting-flutter-versions-earlier-than-3-44

Additionally updated example app to not show warnings about versions of Gradle, Kotlin or NDK.

## Related Issues

Closes #3835

## Breaking Change

### DIFF
--- a/packages/device_info_plus/device_info_plus/example/android/app/build.gradle.kts
+++ b/packages/device_info_plus/device_info_plus/example/android/app/build.gradle.kts
@@ -1,5 +1,6 @@
 plugins {
     id("com.android.application")
+    id("kotlin-android")
     id("dev.flutter.flutter-gradle-plugin")
 }
 
--- a/packages/device_info_plus/device_info_plus/example/android/app/build.gradle.kts
+++ b/packages/device_info_plus/device_info_plus/example/android/app/build.gradle.kts
@@ -1,6 +1,5 @@
 plugins {
     id("com.android.application")
-    id("kotlin-android")
     id("dev.flutter.flutter-gradle-plugin")
 }
 
@@ -25,6 +24,7 @@ android {
         targetSdk = flutter.targetSdkVersion
         versionCode = flutter.versionCode
         versionName = flutter.versionName
+        ndkVersion = flutter.ndkVersion
     }
 
     lint {
--- a/packages/device_info_plus/device_info_plus/example/android/gradle/wrapper/gradle-wrapper.properties
+++ b/packages/device_info_plus/device_info_plus/example/android/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
 #Tue Oct 05 15:15:38 CEST 2021
 distributionBase=GRADLE_USER_HOME
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.13-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.14-bin.zip
 distributionPath=wrapper/dists
 zipStorePath=wrapper/dists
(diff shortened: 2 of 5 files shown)


packages/device_info_plus/device_info_plus/example/android/app/build.gradle.kts
```diff
@@ -1,6 +1,5 @@
 plugins {
     id("com.android.application")
-    id("kotlin-android")
```
mahendrimd, line 3: So I just tested this PR locally using Flutter 3.38.1, and the build for the `example` failed.
The removal of the explicit `kotlin-android` plugin relies on the auto-apply behavior introduced in Flutter upstream via flutter/flutter#184227.

packages/device_info_plus/device_info_plus/example/android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 #Tue Oct 05 15:15:38 CEST 2021
 distributionBase=GRADLE_USER_HOME
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.13-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.14-bin.zip
```
mahendrimd, line 3: the Android Studio recommend using 8.14.5 and when I click that, it automatically appended these when updating to Gradle 8.14.5
```
...
distributionUrl=https\://services.gradle.org/distributions/gradle-8.14.5-bin.zip
networkTimeout=10000
validateDistributionUrl=true
...
```
It is safe to keep these; anyone using older Gradle versions will simply ignore unrecognized properties rather than crash.

mahendrimd (commented):
Just something I caught while testing the PR.

packages/device_info_plus/device_info_plus/example/android/gradle/wrapper/gradle-wrapper.properties
vbuberen: It depends on version of Android Studio. I don't plan to update anything here.

packages/device_info_plus/device_info_plus/example/android/app/build.gradle.kts
vbuberen: Thanks for catching, but just sharing the link was enough. Such lectures are not needed.

Issue #3944, opened 2026-07-31
Android plugins apply KGP based on AGP major version only, breaking on AGP 9 with android.builtInKotlin=false


4akloon:
## Summary

Seven Android plugins decide whether to apply the Kotlin Gradle Plugin (KGP) based on the AGP **major version alone**:

```kotlin
val agpMajor = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION.substringBefore('.').toInt()

if (agpMajor < 9) {
    apply(plugin = "org.jetbrains.kotlin.android")
}
```

This assumes AGP 9 always means built-in Kotlin is active. That is not true. AGP 9 enables built-in Kotlin only when `android.builtInKotlin` is **not explicitly `false`** — and `android.builtInKotlin=false` is exactly what `flutter create` writes into every new app's `android/gradle.properties` (via the Flutter migrator), and what every example app in this repository currently sets.

So on **AGP 9 + `android.builtInKotlin=false`** the guard is `false`, the plugin does not apply KGP, and AGP does not provide Kotlin either. Nothing compiles the plugin's Kotlin sources, and configuration fails a few lines below at the unconditional extension lookup:

```
* Where:
Build file '.../device_info_plus/android/build.gradle.kts' line: 35

* What went wrong:
Extension of type 'KotlinAndroidProjectExtension' does not exist. Currently registered extension types:
[ExtraPropertiesExtension, FlutterExtension, VersionCatalogsExtension, BasePluginExtension, SourceSetContainer,
ReportingExtension, JavaToolchainService, JavaPluginExtension, LibraryExtension, LibraryAndroidComponentsExtension,
LintLifecycleExtension, NamedDomainObjectContainer<BaseVariantOutput>]
```

## Why this is not visible yet

Flutter's Gradle plugin applies `kotlin-android` on behalf of any Android subproject whose build script does not appear to declare KGP. It decides that by regex over the build file **text**, and the Kotlin-DSL regex (`FlutterPluginUtils.kgpRegexKotlin`) only matches `plugins { }` blocks:

```kotlin
internal val kgpRegexKotlin =
[...]

PR #3946, opened 2026-07-31
fix(battery_plus): apply KGP unless built-in Kotlin is enabled

4akloon:
## Description

`battery_plus`'s Android build script decided whether to apply the Kotlin Gradle Plugin (KGP) from the AGP major version alone:

```kotlin
if (agpMajor < 9) {
    apply(plugin = "org.jetbrains.kotlin.android")
}
```

That assumes AGP 9 always means built-in Kotlin is active. AGP 9 enables built-in Kotlin only when `android.builtInKotlin` is **not explicitly `false`** — and `android.builtInKotlin=false` is exactly what `flutter create` writes into every new app's `android/gradle.properties`, and what this repository's example apps set.

So on **AGP 9 + `android.builtInKotlin=false`** the plugin skipped KGP and AGP provided no Kotlin either. Nothing compiled the plugin's Kotlin sources, and configuration failed a few lines below at the unconditional extension lookup:

```
Extension of type 'KotlinAndroidProjectExtension' does not exist.
```

This is not visible today only because Flutter's `FlutterPluginUtils.kgpRegexKotlin` matches `plugins { }` blocks but not the imperative `apply(plugin = "…")` form used here, so Flutter does not see the declaration and applies KGP itself — accidentally papering over the broken guard. Flutter already warns that this fallback is going away.

This PR guards on the same condition AGP itself uses:

```kotlin
// AGP 9 provides Kotlin support natively unless the consuming app opts out with
// android.builtInKotlin=false, which is what `flutter create` writes by default.
val builtInKotlinEnabled =
    agpMajor >= 9 &&
        (providers.gradleProperty("android.builtInKotlin").orNull?.toBoolean() ?: true)

if (!builtInKotlinEnabled) {
    apply(plugin = "org.jetbrains.kotlin.android")
}
```

The `KotlinAndroidProjectExtension` block below is unchanged — AGP 9's built-in Kotlin registers that extension, so it resolves under both branches. On AGP 8 behaviour is provably identical to today: `agpMajor >= 9` is false, so KGP is applied exactly as before.

Full mechanism, AGP bytecode evidence and reproduction steps are in #3944.

### Verification

| Scenario | Result |
| --- | --- |
| AGP 9.1.0 / Kotlin 2.4.0 / Gradle 9.3.1, `builtInKotlin=false`, with Flutter's fallback regex patched out, **before** fix | Fails with `KotlinAndroidProjectExtension does not exist` |
| Same, **after** fix | `flutter build apk --debug` succeeds |
| AGP 9, `builtInKotlin=true` | Succeeds; KGP correctly not applied |
| Stock AGP 8.12.1 (current pin) | Succeeds |

`melos run analyze` and `melos run test:unit_all` both pass.

## Related Issues

- Related #3944

## Checklist

## Breaking Change


ajosephau:
Confirming this downstream. We hit exactly this on a Flutter app with AGP 9: flutter create writes android.builtInKotlin=false, so the agpMajor < 9 guard skips applying KGP while AGP doesn't supply built-in Kotlin either - leaving neither, and the build dies with UnknownDomainObjectException: Extension of type 'KotlinAndroidProjectExtension' does not exist. Checking the property rather than inferring from the AGP version is the right fix.

We're currently pinning a fork carrying just this commit and #3950 so we can move to AGP 9.3.2. With both applied, verified green: assembleDebug, assembleRelease (R8 full mode), and unit tests across the app and a Wear module. Reverting either reproduces the error.

FWIW the failing checks (as of the time I wrote this comment) here look unrelated to the change: they're iOS/macOS/Windows jobs failing on Unable to load contents of file list: '/Target Support Files/Pods-Runner/...xcfilelist' (a CocoaPods PODS_ROOT path issue). While every Android check passes (including integration tests on API 24/29/32/36) other open PRs show the same 5 failures, so it looks repo-wide rather than PR-specific.

PR #3949, opened 2026-07-31
fix(package_info_plus): apply KGP unless built-in Kotlin is enabled

4akloon:
## Description

`package_info_plus`'s Android build script decided whether to apply the Kotlin Gradle Plugin (KGP) from the AGP major version alone:

```kotlin
if (agpMajor < 9) {
    apply(plugin = "org.jetbrains.kotlin.android")
}
```

That assumes AGP 9 always means built-in Kotlin is active. AGP 9 enables built-in Kotlin only when `android.builtInKotlin` is **not explicitly `false`** — and `android.builtInKotlin=false` is exactly what `flutter create` writes into every new app's `android/gradle.properties`, and what this repository's example apps set.

So on **AGP 9 + `android.builtInKotlin=false`** the plugin skipped KGP and AGP provided no Kotlin either. Nothing compiled the plugin's Kotlin sources, and configuration failed a few lines below at the unconditional extension lookup:

```
Extension of type 'KotlinAndroidProjectExtension' does not exist.
```

This is not visible today only because Flutter's `FlutterPluginUtils.kgpRegexKotlin` matches `plugins { }` blocks but not the imperative `apply(plugin = "…")` form used here, so Flutter does not see the declaration and applies KGP itself — accidentally papering over the broken guard. Flutter already warns that this fallback is going away.

This PR guards on the same condition AGP itself uses:

```kotlin
// AGP 9 provides Kotlin support natively unless the consuming app opts out with
// android.builtInKotlin=false, which is what `flutter create` writes by default.
val builtInKotlinEnabled =
    agpMajor >= 9 &&
        (providers.gradleProperty("android.builtInKotlin").orNull?.toBoolean() ?: true)

if (!builtInKotlinEnabled) {
    apply(plugin = "org.jetbrains.kotlin.android")
}
```

The `KotlinAndroidProjectExtension` block below is unchanged — AGP 9's built-in Kotlin registers that extension, so it resolves under both branches. On AGP 8 behaviour is provably identical to today: `agpMajor >= 9` is false, so KGP is applied exactly as before.

Full mechanism, AGP bytecode evidence and reproduction steps are in #3944.

### Verification

| Scenario | Result |
| --- | --- |
| AGP 9.1.0 / Kotlin 2.4.0 / Gradle 9.3.1, `builtInKotlin=false`, with Flutter's fallback regex patched out, **before** fix | Fails with `KotlinAndroidProjectExtension does not exist` |
| Same, **after** fix | `flutter build apk --debug` succeeds |
| AGP 9, `builtInKotlin=true` | Succeeds; KGP correctly not applied |
| Stock AGP 8.12.1 (current pin) | Succeeds |

`melos run analyze` and `melos run test:unit_all` both pass.

## Related Issues

- Related #3944

## Checklist

## Breaking Change


CleanClubManager:
This patch is also what unblocks AGP 9 for Flutter 3.47 apps. Adding a data point in
case it helps prioritise the review, since the workaround usually suggested for #3944
(`android.builtInKotlin=true`) is not available on that Flutter version.

**Environment:** Flutter 3.47.0 stable, AGP 9.0.1, Gradle 9.1.0, KGP 2.3.20, JDK 17,
`compileSdk`/`targetSdk` 36, `android.newDsl=false`.

### There is no escape hatch on Flutter 3.47

Both sides of the `android.builtInKotlin` switch are closed:

- **`builtInKotlin=true`** — AGP's built-in Kotlin pins **KGP 2.2.10**, and Flutter 3.47
  refuses to build below 2.2.20:

  ```
  Error: Your project's Kotlin version (2.2.10) is lower than Flutter's minimum
  supported version of 2.2.20. Please upgrade your Kotlin version.
  ```

  The pin is not specific to one AGP release — the `kotlin-gradle-plugin` dependency in
  the AGP POMs is `2.2.10` in 9.0.1, 9.1.0, 9.2.0, 9.3.0 and 9.4.1 alike, and I could not
  find a documented property to override it. The only bypass Flutter offers is
  `--android-skip-build-dependency-validation`, which silences the deprecation warnings
  we are trying to act on in the first place.

  (`android.newDsl=true` is independently unavailable: Flutter 3.47's Gradle plugin still
  casts to `AbstractAppExtension` and fails with
  `ApplicationExtensionImpl$AgpDecorated_Decorated cannot be cast to AbstractAppExtension`.)

- **`builtInKotlin=false`** — this bug, with the same signature @dea6cat reported above:

  ```
  GeneratedPluginRegistrant.java:69: error: cannot find symbol
        flutterEngine.getPlugins().add(new dev.fluttercommunity.plus.packageinfo.PackageInfoPlugin());
    symbol:   class PackageInfoPlugin
  ```

### Flutter's KGP fallback does not rescue it

`FlutterPluginUtils.detectApplyingKotlinGradlePlugin` applies `kotlin-android` to
subprojects that lack KGP, but it decides by looking for the plugin declaration in the
build file. Here the `apply(plugin = "org.jetbrains.kotlin.android")` *is* present — it
just sits inside the dead `if (agpMajor < 9)` branch — so the fallback concludes KGP is
already applied and skips the module.

### Downgrading is not a way out either

`package_info_plus` 9.0.1 applies KGP unconditionally and builds fine, but it cannot be
selected alongside a `file_picker` recent enough for AGP 9:

| `file_picker` | outcome under AGP 9 |
|---|---|
| 8.3.7 | hardcodes `compileSdk 34`; rejected, `flutter_plugin_android_lifecycle` requires 36 |
| 11.0.3 | same `agpMajor >= 9` assumption as this bug |
| ≥ 12 | builds, but requires `package_info_plus: ^10.2.1` |

So `file_picker` ≥ 12 forces `package_info_plus` 10.x, and 10.2.1 is the latest release.
Every combination I tried fails; I verified each one against a real `flutter build
appbundle --release`, not just resolution.

Net effect: for an app on Flutter 3.47 that depends on both packages, this one-file
change is the difference between "AGP 9 is a bit awkward" and "AGP 9 is unreachable".
Happy to test a pre-release if that is useful.

